PR #25, opened 2025-07-11
[model] feat:add wan

plorrrrrrr:
+ add wan support

### DIFF
--- a/veomni/utils/arguments.py
+++ b/veomni/utils/arguments.py
@@ -71,7 +71,7 @@ class ModelArguments:
         default=False,
         metadata={"help": "Whether to encode target with decoder. Only supports stable diffusion as decoder."},
     )
-    attn_implementation: Optional[Literal["eager", "sdpa", "flash_attention_2"]] = field(
+    attn_implementation: Optional[Literal["eager", "sdpa", "flash_attention_2", "flash_attention_3"]] = field(
         default="flash_attention_2",
         metadata={"help": "Attention implementation to use."},
     )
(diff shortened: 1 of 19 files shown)


veomni/utils/helper.py
```diff
@@ -117,8 +141,12 @@ def state_dict(self) -> Dict[str, Any]:
     def load_state_dict(self, state_dict: Dict[str, Any]):
         self.consume_tokens = state_dict["consume_tokens"]
 
-    def add(self, micro_batch: Dict[str, "torch.Tensor"]) -> None:
-        seqlens = _compute_seqlens(micro_batch, self.rmpad, self.rmpad_with_pos_ids)
+    def add(self, micro_batch: Dict[str, "torch.Tensor"], model_type: Optional[str] = None) -> None:
+        if model_type == "wan":
```
Fazziekey: For wan only?

veomni/utils/arguments.py
```diff
@@ -300,6 +300,10 @@ class TrainingArguments:
         default=False,
         metadata={"help": "Enable activation offload to CPU."},
     )
+    ops_to_save: List[str] = field(
```
Fazziekey: move it to train_wan.py custom args

veomni/utils/helper.py
plorrrrrrr: > For wan only?

yes，the calculation process of different dit models is not the same

veomni/utils/arguments.py
plorrrrrrr: > move it to train_wan.py custom args

done

Fazziekey (approved):

PR #28, opened 2025-07-30
[model] feat: add flux

yuyu5333:
# Add model Flux.1 dev support

- support flux sft, lora training
- add lora_utils

## Key Changes
- Model Configuration Parameters
Added new LoRA-related parameters under the model namespace to control training behavior:
```
model:
  lora_target_modules: Specifies the layers to apply LoRA training (e.g., "q,k,v").
  lora_target_modules_support: Defines all layers in the model that are compatible with LoRA training (e.g., "q,k,v,o,ffn.0,ffn.2" for models Wan2.1).
  lora_alpha: Scaling parameter for LoRA updates.
  lora_rank: Rank of the low-rank matrices used in LoRA.
  init_lora_weights: Method for initializing LoRA weights (default: "kaiming").
  pretrained_lora_path (optional): Path to load pre-trained LoRA weights if available.
```

- Training Architecture Specification
```
train:
  train_architecture: Parameter to explicitly set the training mode to "lora", ensuring the training pipeline uses LoRA-specific logic.
```

# fix requirements

The ALL_ATTENTION_FUNCTIONS module in transformers.modeling_utils now supports registration via the AttentionInterface register  decorator (introduced in v4.51.0). 

```
# Global AttentionInterface shared by all models which do not need to overwrite any of the existing ones
ALL_ATTENTION_FUNCTIONS: AttentionInterface = AttentionInterface()

class AttentionInterface(MutableMapping):
    ...
    @classmethod
    def register(cls, key: str, value: Callable):
        cls._global_mapping.update({key: value})
```

This allows custom attention mechanisms to be dynamically registered, resolving previous errors where custom functions couldn't be integrated without modifying core library code.


veomni/utils/arguments.py
```diff
@@ -87,6 +87,30 @@ class ModelArguments:
         default=False,
         metadata={"help": "Force loading model from huggingface."},
     )
+    lora_rank: int = field(
```
Fazziekey: move to train_flux.py

veomni/models/transformers/flux/config_flux.py
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from transformers import PretrainedConfig
+
+
+class FluxConfig(PretrainedConfig):
+    model_type = "flux"  # 必须与注册时的 model_type 一致
```
plorrrrrrr: change comments to English

yuyu5333: fixed

veomni/utils/arguments.py
yuyu5333: done

PR #49, opened 2025-08-18
apply weight decay only to weight

yangtian6781:
只对不是bias对权重施加weight decay


yangtian6781:
只对不是bias的权重施加weight decay

PR #57, opened 2025-08-22
fix: argument validation

jxlw2014:
### Summary
When training with single GPU, `fsdp_enabled` is false, 
```python
@property
def fsdp_enabled(self) -> bool:
    return self.fsdp_size > 1
```
then framework won't build FSDP model anymore even if `dp_mode` is `fsdp1` or `fsdp2`. Some logics like 
```python
if args.train.data_parallel_mode == "fsdp1":
    grad_norm = model.clip_grad_norm_(args.train.max_grad_norm).item()
else:
```
will crash


Adenialzz:
Hi, so single gpu must set ddp (not fsdp), right ?

PR #67, opened 2025-09-08
[model] fix: qwen3-moe-a3b load

ziqi-wlb:
### fix qwen3-moe-a3b load

When running the qwen3-moe-a3b model, I found that the initial loss(17+) was too high. The final location was a bug in the model loading, then the experts were randomly initialized.

- before fix, first-step loss=17
image

- solution:
1. The loading problem is mainly due to the mismatch between the model definition and the checkpoint format: VeOmni's Qwen3MoeExperts uses a single stacked parameter `torch.nn.Parameter(self.num_experts, ...)`, while the hf-checkpoint uses independent expert parameters `experts.0.gate_proj.weight`, `experts.1.gate_proj.weight`, etc.
2. Loading speed optimization: There are multiple experts, for example, qwen3-moe-a3b has 128 experts. Searching and loading them individually each time will be extremely slow. The cache mechanism is used to avoid repeated module searches, and batch loading is used to merge multiple experts into a single loading operation.

- after fix, first-step loss=2.8
image

- Reproduction script:
```sh

MODEL_PATH=/cpfs/user/weishi/models/Qwen3-30B-A3B
SAVE_PATH=Qwen3-30B-A3B
CONFIG=configs/pretrain/qwen3-moe.yaml

DATA_PATH=/prodcpfs/user/weishi/data/fineweb/sample/10BT/
DATA_SET_TYPE=mapping

bash train.sh tasks/train_torch.py $CONFIG \
    --model.model_path ${MODEL_PATH} \
    --data.train_path $DATA_PATH \
    --data.num_workers 32 \
    --data.datasets_type $DATA_SET_TYPE \
    --train.global_batch_size 128 \
    --train.lr 5e-7 \
    --train.ulysses_parallel_size 1 \
    --train.save_steps 10 \
    --train.max_steps 20 \
    --train.ckpt_manager dcp \
    --train.output_dir ${SAVE_PATH}
```

### DIFF
--- a/veomni/models/module_utils.py
+++ b/veomni/models/module_utils.py
@@ -162,6 +162,182 @@ def _dispatch_parameter(
         module._parameters[name].data.copy_(tensor)
 
 
+def _dispatch_parameter_cached(
+    module: "nn.Module",
+    name: str,
+    tensor: "torch.Tensor",
+    module_cache: dict,
+    dtensor_factory: Optional[Callable[["torch.Tensor", Any, Any], "torch.Tensor"]] = None,
+) -> None:
+    """
+    Assigns parameter to an empty model with module reference caching.
+
+    This is optimized for MoE models where the same module path is accessed multiple times.
+    """
+    # Check cache first
+    if name in module_cache:
+        target_module, param_name = module_cache[name]
+    else:
+        # Cache miss: find module and cache the result
+        target_module, param_name = _find_submodule(module, name)
+        module_cache[name] = (target_module, param_name)
+
+    # Get the parameter tensor
+    orig_tensor = target_module._parameters[param_name].data
+    tensor = tensor.to(orig_tensor)
+    if hasattr(orig_tensor, "device_mesh"):  # dtensor
+        if orig_tensor.device.type == "cpu":
+            raise ValueError("Cannot load dtensor on CPU.")
+
+        device_mesh = getattr(orig_tensor, "device_mesh")
+        placements = getattr(orig_tensor, "placements")
+        target_module._parameters[param_name].data.copy_(dtensor_factory(tensor, device_mesh, placements))
+    else:  # not dtensor
+        target_module._parameters[param_name].data.copy_(tensor)
+
+
+def _batch_dispatch_moe_parameters(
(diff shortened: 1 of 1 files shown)


TueVNguyen:
Did you run scripts/moe_merge.py first?

ziqi-wlb:
> Did you run scripts/moe_merge.py first?

OK, `scripts/moe_merge.py` is not executed (it is not mentioned in the document, I did not see this file~). From the perspective of ease of use, it should be insensitive to users. Reducing manual operations is relatively more elegant?

Fazziekey:
see https://github.com/ByteDance-Seed/VeOmni/blob/main/scripts/moe_ckpt_merge/moe_merge.py

ziqi-wlb:
> see https://github.com/ByteDance-Seed/VeOmni/blob/main/scripts/moe_ckpt_merge/moe_merge.py

We use fsdp to train models because it is simpler to use than megatron-core, so the burden of manual operation should be reduced as much as possible, and there are no errors or loading failures that would cause people to step into the trap~ (For example, the early megatron-core had to rely on different TP/PP ckpt splits before training, which was always complained about, but was later optimized)

Luosuu:
Hi @ziqi-wlb , thank you very much for your efforts and the PR! This is great work from the user experience.

However, we believe a more reasonable way of storing experts weight should be the "merged" instead of the current style of Qwen3-MoE HuggingFace checkpoints. Personally, I believe this is a design mistake on their side. Instead of trying to catching their mistakes, we prefer to have a dedicated script to "fix" their checkpoints first. 

From another perspective,  we should not assume all other models have the same way of storing expert weights with the current Qwen3-MoE hf checkpoints. We do not write different loaders / dispatchers for different models inside the library itself, which can be pretty messy from the view of maintainability.

By holding a simple assumption of the "merged" expert weight form, we can 
1. keep the core codebase easy to read
2. make model loading part consistent to different MoE models
3. more maintainable by providing dedicated scripts to fix model weights if they are not in the way we expected

Hopefully this make sense to you

ziqi-wlb:
Ok, I'll close this PR.

PR #86, opened 2025-09-24
[dist] fix: FSDP2 with flexible mixed policy control

Luosuu:
This PR fixes flexible MP policy control.
* When a model provides modules that should ignore MP policy, we would apply a separate `fully_shard` on them without `mp_policy` key.
* This consequently changes our implementation of prefetching configuration, where we need to prefetch every FSDPModule in the next/prev layer during forward/backward.

### DIFF
--- a/veomni/distributed/torch_parallelize.py
+++ b/veomni/distributed/torch_parallelize.py
@@ -328,7 +328,7 @@ def parallelize_model_fsdp2(
     #      | -- Gating (to be ignored with mix precision)
     #      | -- MLP (experts)
     for layer_fqn, layer_mod, experts_mod in layer_pairs:
-        # register all the FSDPModule inside this deocder layer for the convenience of prefetching configuration
+        # register all the FSDPModule inside this decoder layer for the convenience of manual prefetching configuration
         layer_mod._fsdp_modules = []
         if parallel_state.ep_enabled and experts_mod is not None:
             # shard expert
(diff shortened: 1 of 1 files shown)


veomni/distributed/torch_parallelize.py
```diff
@@ -312,59 +322,72 @@ def _experts_shard_placement_fn(param):
...
+                # average EP grads across EP ranks
                 experts_mod.set_gradient_divide_factor(parallel_state.ep_size)
-        # shard decoder layer
-        _apply_fsdp_to_block(layer_mod)
+            layer_mod._fsdp_modules.append(experts_mod)
+        # shard module that needs to ignore mixed precision control
+        if ignore_classes:
+            for sub_mod in layer_mod.modules():
+                if isinstance(sub_mod, ignore_classes) and sub_mod is not layer_mod:
+                    # this will also create a AllGather communication group
```
piyifan123, line 344: as long as it does not prevent fsdp2 from AG other layers (e.g., expert) i think a separate AG for those modules are fine?

veomni/distributed/torch_parallelize.py
```diff
@@ -299,7 +298,18 @@ def parallelize_model_fsdp2(
     else:
         modules_to_ignore_in_mixed_precision = None
 
-    # 2.2: ep_fsdp2 kwargs
+    if modules_to_ignore_in_mixed_precision:
+        assert isinstance(modules_to_ignore_in_mixed_precision, tuple), (
+            "modules_to_ignore_in_mixed_precision needs to be a tuple!"
+        )
+        ignore_classes = modules_to_ignore_in_mixed_precision
```
piyifan123, line 305: nit: mp_ignored_classes to be more exact?

veomni/distributed/torch_parallelize.py
```diff
@@ -312,59 +322,72 @@ def _experts_shard_placement_fn(param):
...
-        ):
-            # pass all kwargs except mp_policy
-            fully_shard(block, **{k: v for k, v in fsdp_kwargs.items() if k != "mp_policy"})
-        else:
-            fully_shard(block, **fsdp_kwargs)
-
+    # Here we have a basic assumption for the decoder layer hierarchy
+    # Decoder layer
+    #      | -- Attention
+    #      | -- Gating (to be ignored with mix precision)
```
piyifan123, line 328: nit: i think a better way to phrase this would be
Decoder Layer
| -- layers that inherit the fully_shard as Decode Layer (e.g., Attention)
| -- experts layer (apply fully_shard separately in order  to shard across EP of the same rank instead of sharding globally)
| -- layers (declared in model.modules_to_ignore_in_mixed_precision) that need to apply fully_shard separately due to different mp policy as the decoder layer (e.g., some models requires MoE TopK gate layer to have parameters in higher FP32 precision in forward).

WDYT?

Luosuu: You mean the comment contents? I will improve this in next PR. Thank you!

veomni/distributed/torch_parallelize.py
Luosuu: sure.

veomni/distributed/torch_parallelize.py
Luosuu: maybe, I have not tested MFU at large-scale. Will do so later.

Issue #89, opened 2025-09-24
Assertion on output_dir depends on internal veomni_patch。


KaijingOfficial:
Hi, I noticed that this [assertion ](https://github.com/ByteDance-Seed/VeOmni/blob/53fc5b56a9b7be5dac3e573dd0b86813f36a1116/veomni/utils/arguments.py#L552C1-L554C10)depends on helper.is_remote_path, which in turn relies on your internal patch. This raises an error when training with multi-node setups. Maybe we should make this check more flexible?

Luosuu:
ahh yes, please comment this out and it should work? you can make a PR if you want or I can fix it later

KaijingOfficial:
Got it.  I’ll PR a fix shortly. Thanks!

PR #97, opened 2025-09-25
fix:remove a failing assertion

KaijingOfficial:
### What does this PR do?

As noted in #89, I've removed the unnecessary path check assertion.

### DIFF
--- a/veomni/utils/arguments.py
+++ b/veomni/utils/arguments.py
@@ -553,9 +553,13 @@ class TrainingArguments:
         if self.rmpad and self.rmpad_with_pos_ids:
             raise ValueError("`rmpad` and `rmpad_with_pos_ids` cannot be both True.")
 
-        assert int(os.getenv("GROUP_WORLD_SIZE", 1)) == 1 or helper.is_remote_path(self.output_dir), (
-            f"Number of distribute nodes is {int(os.getenv('GROUP_WORLD_SIZE'))}, `train.output_dir` requires a remote path, should use hdfs, fuse or nas path, but got {self.output_dir}."
-        )
+        num_nodes = int(os.getenv("WORLD_SIZE", 1)) // int(os.getenv("LOCAL_WORLD_SIZE", 1))
+        if num_nodes > 1:
+            logger.warning_rank0(
+                f"Detected {num_nodes} nodes. "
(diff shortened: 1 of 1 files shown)


Luosuu:
Hi @KaijingOfficial , thank you very much for your PR!

The risk here is that, if the path is not a remote path that is mounted on every node, the saved checkpoints are incomplete in every node.

For example, if there are two nodes, and you specify a local path `./save_ckpt` (and this path is not a remote path), then node0 and node1 saves their share of the model on their local path. 

If the path is a mounted remote path like `/mnt/save_ckpt` then the checkpointer on each node saves their parts to this same path. 

Simply commenting this out would work silently though, but I think we need to notify the user about this risk. So could you please add some warning here?

cc @Fazziekey

KaijingOfficial:
Hi @Luosuu, thanks for pointing this out — I appreciate the feedback!

In my case (Slurm cluster), all compute nodes share the same filesystem, so this change works fine.
However it may not work in other setups — I’ll add a warning to notify users of this limitation.

veomni/utils/arguments.py
```diff
@@ -549,10 +549,6 @@ def __post_init__(self):
         if self.rmpad and self.rmpad_with_pos_ids:
             raise ValueError("`rmpad` and `rmpad_with_pos_ids` cannot be both True.")
 
-        assert int(os.getenv("GROUP_WORLD_SIZE", 1)) == 1 or helper.is_remote_path(self.output_dir), (
```
Luosuu, line 556: add some warnings to notify the users about the risk. then we can merge!

Luosuu (changes requested):
need some warning info

Fazziekey (approved):

veomni/utils/arguments.py
KaijingOfficial: Done! Added the warning per your suggestion. Feel free to review.

Luosuu:
We should do rebase instead of merge otherwise the git history will be messy.

cc @Fazziekey

Luosuu:
Can you try undo your commits on your local and perform rebase instead? Thank you

KaijingOfficial:
Thanks for the review! I've rebased the branch onto the latest main and force-pushed the final version. @Luosuu

PR #106, opened 2025-09-26
[dist] fix: DCP auto load

Luosuu:
find latest DCP checkpoint path by scanning paths with legit `.metadata` file, which represents DCP saving was successful.

usage: `train.sh tasks/train_torch.py /path/to/job.yaml --train.ckpt_manager dcp --train.load_checkpoint_path auto`

reference: https://github.com/pytorch/torchtitan/blob/5b5d46856b400c8550989415bee91473aab4f921/torchtitan/components/checkpoint.py#L619

tests will be added after [PR#98](https://github.com/ByteDance-Seed/VeOmni/pull/98)

### DIFF
--- a/veomni/utils/checkpoint_utils.py
+++ b/veomni/utils/checkpoint_utils.py
@@ -23,6 +23,10 @@ try:
 except ImportError:
     from .hdfs_io import copy, exists, isdir, listdir
 
+from .logging import get_logger
+
+
+logger = get_logger(__name__)
 
 _GLOBAL_STEP_PREFIX = "global_step_"
 
@@ -86,6 +90,8 @@ def dcp_get_last_iteration(output_dir):
 
     if not valid_steps:
         return None
+
+    logger.info_rank0(f"found valid previously saved checkpointed steps: {checkpoints_dir}/global_step_{valid_steps}")
 
     return max(valid_steps)
 
--- a/veomni/checkpoint/checkpointer.py
+++ b/veomni/checkpoint/checkpointer.py
@@ -22,6 +22,7 @@ import torch.distributed as dist
 from torch.distributed._tensor import DeviceMesh, DTensor, Shard
 
 from ..distributed.parallel_state import get_parallel_state
+from ..utils.checkpoint_utils import _GLOBAL_STEP_PREFIX
 from ..utils.import_utils import is_torch_version_greater_than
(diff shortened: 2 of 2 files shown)


veomni/utils/checkpoint_utils.py
```diff
@@ -19,9 +19,35 @@
...
 
 try:
-    from hdfs_io import copy, exists
+    from hdfs_io import copy, exists, isdir, listdir
 except ImportError:
-    from .hdfs_io import copy, exists
+    from .hdfs_io import copy, exists, isdir, listdir
+
+
+_GLOBAL_STEP_PREFIX = "global_step_"
```
piyifan123, line 31: modify the save code to use this as well

veomni/utils/checkpoint_utils.py
```diff
@@ -19,9 +19,35 @@
...
+    try:
+        step = int(step_str)
+    except ValueError:
+        return None
+
+    checkpoint_path = os.path.join(checkpoints_dir, entry)
+    if not isdir(checkpoint_path):
+        return None
+
+    model_metadata_path = os.path.join(checkpoint_path, "model/.metadata")
```
piyifan123, line 49: which folder are dataloader state and other states stored in?

veomni/utils/checkpoint_utils.py
```diff
@@ -46,7 +72,33 @@ def get_last_iteration(output_dir, is_rank0: bool):
...
-def get_checkpoint_path(output_dir, is_rank0: bool):
-    iteration = get_last_iteration(output_dir, is_rank0)
-    if iteration:
-        return os.path.join(output_dir, "checkpoints", f"global_step_{iteration}")
+def dcp_get_last_iteration(output_dir):
+    checkpoints_dir = os.path.join(output_dir, "checkpoints")
+    if not exists(checkpoints_dir):
+        return None
+
+    entries = listdir(checkpoints_dir)
```
piyifan123, line 84: nit: add comment on if it works.

piyifan123 (approved):

veomni/utils/checkpoint_utils.py
Luosuu: extra states are not saved in dcp format but in `.pt` so we cannot check whether they have `.metadata`

Issue #102, opened 2025-09-26
Assertion Failure When Comparing Optimizer State Dictionaries After Synchronous Checkpoint Saving in Distributed Training


dyyoungg:
### Environment Configuration:
- torch==2.7.1+cu126
- transformers==4.56.2
- triton==3.3.1
- liger-kernel==0.5.10

### Test Script
`tests/utils/test_trainer_saveload.py`

**Execution Command:**
```bash
torchrun --nnodes=1 --nproc-per-node=8 --master-port=4321 tests/utils/test_trainer_saveload.py \
    --model.model_path Qwen3-30B-A3B-Instruct-2507-veomni-merge \
    --model.moe_implementation fused \
    --model.attn_implementation flash_attention_2 \
    --train.expert_parallel_size 4 \
    --train.global_batch_size 8 \
    --train.micro_batch_size 1 \
    --data.max_seq_len 128 \
    --data.train_path "dummy" \
    --train.output_dir ./test_trainer_saveload \
    --train.max_steps 10 \
    --train.rmpad false \
    --train.rmpad_with_pos_ids true \
    --train.data_parallel_mode "fsdp2" \
    --train.init_device "meta" \
    --train.ckpt_manager "dcp"
```
**Modification**
-  Changed `Checkpointer.save` to use synchronous mode (save_async=False) due to hanging issues with asynchronous saving.
-  Changed `build_stream_dataloader` to `build_dataloader` due to the lack of `veomni_patch`

### Problem Description:​​
1. When using `Checkpointer.save(..., save_async=False)`,  after completing training and performing synchronous checkpoint saving, the test fails during comparison of optimizer state dictionaries. Multiple ranks report tensor mismatches with both absolute and relative differences exceeding allowed tolerances. The failure occurs specifically during check_state_dict()validation of optimizer states.

**Detalied Log error**
```
[rank2]: Traceback (most recent call last):
[rank2]:   File "/path/to/test_trainer_saveload.py", line 380, in <module>
[rank2]:     main()
[rank2]:   File "/path/to/test_trainer_saveload.py", line 369, in main
[rank2]:     check_state_dict(golden_optim_sd, optimizer.state_dict(), need_flatten=True)
[rank2]:   File "/path/to/test_trainer_saveload.py", line 99, in check_state_dict
[rank2]:     torch.testing.assert_close(rhs_val, lhs_val)
[...]

PR #107, opened 2025-09-27
[dist] feat: Unified dcp saving for model and optimizer

Luosuu:
This PR unifies dcp saving session for model and optimizer,
so that we can avoid concurrent async saving sessions which would cause hang sometimes

this fixes the dcp async hang mentioned in #102

### DIFF
--- a/veomni/checkpoint/checkpointer.py
+++ b/veomni/checkpoint/checkpointer.py
@@ -293,7 +293,7 @@ class DistributedCheckpointer(CheckpointerBase):
     Distributed checkpointer for torch.distributed.checkpoint
     """
 
-    save_model_future: Optional[Any] = None
+    dcp_save_future: Optional[Any] = None
     # Dedicated process group for async saves (created on first use)
     _async_process_group: Optional[Any] = None
 
@@ -331,12 +331,12 @@ class DistributedCheckpointer(CheckpointerBase):
             if cls._async_process_group is None:
                 cls._async_process_group = dist.new_group(backend="gloo")
 
-            if cls.save_model_future is not None:
+            if cls.dcp_save_future is not None:
                 logger.info_rank0("waiting for previous DCP saving session to end...")
-                cls.save_model_future.result()
-                cls.save_model_future = None
+                cls.dcp_save_future.result()
+                cls.dcp_save_future = None
 
-            cls.save_model_future = dcp.async_save(
+            cls.dcp_save_future = dcp.async_save(
                 state_dict=save_state,
                 storage_writer=FileSystemWriter(
                     checkpoint_dir,
--- a/veomni/checkpoint/checkpointer.py
+++ b/veomni/checkpoint/checkpointer.py
@@ -46,8 +46,6 @@ else:
 logger = get_logger(__name__)
 
 _EXTRA_STATE_FORMAT = "extra_state_rank_{}.pt"
-_MODEL_DIR = "model"
-_OPTIMIZER_DIR = "optimizer"
 _EXTRA_STATE_DIR = "extra_state"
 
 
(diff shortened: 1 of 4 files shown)


veomni/checkpoint/checkpointer.py
```diff
@@ -325,67 +322,40 @@ def save(
...
             # Lazily create a dedicated Gloo process group for async DCP saves
             if cls._async_process_group is None:
                 cls._async_process_group = dist.new_group(backend="gloo")
 
-            if cls.save_model_future is not None:
-                logger.info_rank0("waiting for previous DCP model saving session to end...")
-                cls.save_model_future.result()
-                cls.save_model_future = None
+            if cls.dcp_save_future is not None:
+                logger.info_rank0("waiting for previous DCP saving session to end...")
```
piyifan123: why info_rank0? should we just info on whichever rank who is waiting for future? IIUC, every rank has their own future.

BTW, does torchtitan add a dist barrier here to ensure every rank has resovled their future before proceding?

piyifan123 (approved):
LGTM. Please wait for @Fazziekey reviews as well.

Like this change, much cleaner to save as one state now.

veomni/checkpoint/checkpointer.py
```diff
@@ -325,67 +322,40 @@ def save(
...
-                    state_dict={"state": OptimizerState(model=state["model"], optimizer=state["optimizer"])},
-                    storage_writer=FileSystemWriter(
-                        optimizer_dir,
-                        thread_count=16,
-                        single_file_per_rank=True,
-                        sync_files=False,
-                    ),
-                )
 
         if "extra_state" in state:
```
piyifan123: should we save extra_state together as well? i'm wondering what happens if the main saving step succeed but this fails? 

if it's hard to be done in the same state for extra_state. how about the followng

1. torch.save extra_state at first.
2. initialize torch.save for the main ckpt.
3. How do we check the latest successul ckpt? we should have both extra_state and main ckpt with a valid .metadata

veomni/checkpoint/checkpointer.py
Luosuu: Indeed, I will change the logger to inform every rank that has not finished their future.

Nope, torchtitan just wait on each rank to result the future. I think adding a dist.barrier() is safer here.

veomni/checkpoint/checkpointer.py
Luosuu: Sure. I will save extra_state first. 

extra_state are not saved in dcp format so they do not have a `.metadata` file for validation..At least by saving them first, we can guarantee that for successfully saved ckpts they must have their extra_state saved before them.

Luosuu: done

veomni/checkpoint/checkpointer.py
Luosuu: done

veomni/checkpoint/checkpointer.py
```diff
@@ -46,8 +46,6 @@
 logger = get_logger(__name__)
 
 _EXTRA_STATE_FORMAT = "extra_state_rank_{}.pt"
-_MODEL_DIR = "model"
-_OPTIMIZER_DIR = "optimizer"
```
Fazziekey, line 50: 这个删除了 _MODEL_DIR _OPTIMIZER_DIR 存下来的目录是怎么样的

Luosuu: `output_dir/checkpoints/global_step_xxx/xxxx.distcp` for a whole state dict for both model and optimizer.

To load model, we use `state_dict['model']`; To load optimizer, we use `state_dict['optimizer']`

extra states are at `output_dir/checkpoints/global_step_xxx/extra_state/extra_state_rank_{rank}.pt`

Luosuu: @Fazziekey does this look good to you?

PR #112, opened 2025-10-01
[dist] fix: remove unnecessary assert

Luosuu:
Remove unnecessary assertion in prefetching since even in MoE models not necessarily every decoder layer contains expert module


piyifan123 (approved):
Please add a unit test to test with fake models to ensure this logic works for all corner cases.

Luosuu:
will do so in the following PRs

PR #113, opened 2025-10-01
[misc] feat: option to profile rank0 only or all the ranks

Luosuu:
Adding `profile_rank0_only` option (default as true) to end user.
Then a `profile_this_rank` argument is determined in the post init of training arguments.
Trainer script would leverage `profile_this_rank` to decide whether to create a profiler on this rank.
Consequently, we add a `global_rank` option to `create_profiler` for correct naming profile artifacts.

Also provides a script to merge profiles:
```
python scripts/profile/merge_chrome_trace.py --input-dir ./trace
```

* Comes with a fix to only use `ProfileWithMem` profiler when `profile_memory` is true instead of when CUDA is available.

### DIFF
--- a/veomni/utils/arguments.py
+++ b/veomni/utils/arguments.py
@@ -499,7 +499,12 @@ class TrainingArguments:
         default=True,
         metadata={"help": "Whether or not to record the stack traces."},
     )
-    profile_rank0_only: bool = field(default=True, metadata={"help": "whether to profile rank0 only"})
+    profile_rank0_only: bool = field(
+        default=True,
+        metadata={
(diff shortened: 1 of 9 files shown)


veomni/utils/arguments.py
```diff
@@ -499,6 +499,7 @@ class TrainingArguments:
         default=True,
         metadata={"help": "Whether or not to record the stack traces."},
     )
+    profile_rank0_only: bool = field(default=True, metadata={"help": "whether to profile rank0 only"})
```
piyifan123: add tooltip on what happens false and warning on it could take more spaces and time.

Luosuu: donw

piyifan123 (approved):

PR #115, opened 2025-10-03
[config] feat: add no_cuda_launch_blocking

Luosuu:
prevent CUDA_LAUNCH_BLOCKING from accidentally enabled

### DIFF
--- a/veomni/utils/arguments.py
+++ b/veomni/utils/arguments.py
@@ -383,6 +383,10 @@ class TrainingArguments:
         default=False,
         metadata={"help": "Enable full determinism."},
     )
+    no_cuda_launch_blocking: bool = field(
+        default=True,
+        metadata={"help": "Prevent CUDA_LAUNCH_BLOCKING from being accidentally enabled."},
+    )
     empty_cache_steps: int = field(
         default=500,
         metadata={"help": "Number of steps between two empty cache operations."},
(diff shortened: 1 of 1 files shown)


veomni/utils/arguments.py
```diff
@@ -631,6 +635,16 @@ def __post_init__(self):
...
             self.profile_this_rank = False
 
+        # Prevent CUDA_LAUNCH_BLOCKING from being accidentally enabled
+        if self.no_cuda_launch_blocking:
+            assert not self.enable_full_determinism, (
+                "no_cuda_launch_blocking and enable_full_determinism are enabled at the same time. But enable_full_determinism would set CUDA_LUANCH_BLOCKING to 1!"
+            )
+            cuda_launch_blocking_val = os.environ.get("CUDA_LAUNCH_BLOCKING", "").strip()
+            assert cuda_launch_blocking_val != "1", (
+                "CUDA_LAUNCH_BLOCKING=1 is set when no_cuda_launch_blocking is also enabled!"
```
piyifan123: please suggest user not to set CUDA_LAUNCH_BLOCKING=1 when launching a training job. unless they are debugging something.

veomni/utils/arguments.py
```diff
@@ -383,6 +383,10 @@ class TrainingArguments:
         default=False,
         metadata={"help": "Enable full determinism."},
     )
+    no_cuda_launch_blocking: bool = field(
```
piyifan123: allow_cuda_launch_blocking_debug_only default: False.

Turn on CLB affect perfomance significantly. if you don't know what you are doing (e.g., debug), do not set this to be True. and if you get an error on setting this, try to change your launch command not to set env CLB=1...

PR #121, opened 2025-10-07
[ckpt] fix: remove unnecessary path joining for dcp

Luosuu:
DCP checkpoints no longer have separate "model" and "optimizer" folders for checkpoints but a unified storage.

### DIFF
--- a/veomni/checkpoint/format_utils.py
+++ b/veomni/checkpoint/format_utils.py
@@ -118,6 +118,7 @@ def dcp_to_torch_state_dict(save_checkpoint_path: Union[str, os.PathLike]) -> ST
         no_dist=True,
     )
     if "state" in state_dict:
+        # this happens when the model state dicts are flatten during saving
         state_dict = state_dict["state"]
 
     return state_dict["model"]


veomni/checkpoint/format_utils.py
```diff
@@ -122,6 +118,7 @@ def dcp_to_torch_state_dict(save_checkpoint_path: Union[str, os.PathLike]) -> ST
         no_dist=True,
     )
     if "state" in state_dict:
-        state_dict = state_dict["state"]
+        # in case the model state dicts were flattened at saving
+        state_dict = state_dict["state"]["model"]
```
piyifan123: i don't understand this: would the next line make it state_dict["state"]["model"]["model"]?

Luosuu: ah actually you are right. I should add a return here or just leave it as it was

veomni/checkpoint/format_utils.py
```diff
@@ -108,10 +108,6 @@ def dcp_to_torch_state_dict(save_checkpoint_path: Union[str, os.PathLike]) -> ST
         To avoid OOM, it's recommended to only run this function on a single rank.
     """
 
-    model_dir = os.path.join(save_checkpoint_path, _MODEL_DIR)
```
piyifan123, line 111: please add a comment in L120 on when that happens...

piyifan123 (approved):

PR #125, opened 2025-10-10
Fix Qwen3-Moe MFU

zhihaofang1017:
### What does this PR do?

This PR corrects the FLOPs (Floating-Point Operations Per Second) calculation logic in the _estimate_qwen3_moe_flops method, ensuring the attention module's computational cost estimation aligns with Qwen3-MoE's actual attention mechanism and fixes a redundant head_dim configuration read issue.

### Design & Code Changes

1. Adjust Attention QKV FLOPs CoefficientModified the attn_qkv_flops calculation by adding a * 0.5 coefficient. This adjustment accounts for Qwen3-MoE's optimized attention design, which reduces actual computational load compared to standard Multi-Head Attention—fixing the previous overestimation of attention-layer FLOPs.
2. Standardize head_dim Configuration ReadingReplaced the hardcoded head_dim = hidden_size // num_attention_heads with head_dim = getattr(self.config, "head_dim", self.config.hidden_size // self.config.num_attention_heads). This allows the method to prioritize explicitly configured head_dim from the model config in Transformers while retaining the fallback calculation.

After modification, the FLOPs estimation for Qwen3-MoE shows better during inference/training.
The head_dim logic now correctly reads explicit config values (if present) and falls back to the default calculation, with no dimension errors observed in tests with standard and custom Qwen3-MoE configs.

### DIFF
--- a/veomni/utils/count_flops.py
+++ b/veomni/utils/count_flops.py
@@ -198,7 +198,7 @@ class VeomniFlopsCounter:
         seqlen_square_sum = 0
         for seqlen in batch_seqlens:
             seqlen_square_sum += seqlen * seqlen
-        attn_qkv_flops = 12 * seqlen_square_sum * head_dim * num_attention_heads * num_hidden_layers * 0.5
+        attn_qkv_flops = 12 * seqlen_square_sum * head_dim * num_attention_heads * num_hidden_layers
 
         # all_layer & all_token fwd & bwk flops
         flops_all_token = dense_N_flops + attn_qkv_flops


veomni/utils/count_flops.py
```diff
@@ -198,7 +198,7 @@ def _estimate_qwen3_moe_flops(self, tokens_sum, batch_seqlens, delta_time):
         seqlen_square_sum = 0
         for seqlen in batch_seqlens:
             seqlen_square_sum += seqlen * seqlen
-        attn_qkv_flops = 12 * seqlen_square_sum * head_dim * num_attention_heads * num_hidden_layers
```
piyifan123, line 201: we discussed this internally before and decided to use 12 as the coefficient even if it's actually casual attention which requires only half of flops in theory. 

this number is being used in other framework like veRL as well https://github.com/volcengine/verl/blob/c9910c266da1da70139959024654424d2394b584/verl/utils/flops_counter.py#L266

so it's better to keep it in sync with other frameworks and convention here.

zhihaofang1017: I agree with you, and I have deleted "0.5" for attn_qkv_flops. However, the calculation logic of "head_dim" is right, it is being used in VeRL as well
https://github.com/volcengine/verl/blob/c9910c266da1da70139959024654424d2394b584/verl/utils/flops_counter.py#L157

tardis-key:
@Fazziekey @FoolPlayer @Coach257 Can u review  or merge it？

PR #139, opened 2025-10-15
[model] fix: deepseek-v3

Luosuu:
* fix deepseek-v3
* support weight merging for deepseek-v3
* add new script to trim hidden layer number in safetensors for convenience of testing

### DIFF
--- a/scripts/trim_safetensor_layers.py
+++ b/scripts/trim_safetensor_layers.py
@@ -1,4 +1,7 @@
 """
+This script trims hidden layer number for a given safetensor dir
+so that we can test weight loading against large models like deepseek conveniently
+
 Example usage:
 python scripts/trim_safetensor_layers.py \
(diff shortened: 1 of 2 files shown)


scripts/trim_safetensor_layers.py
```diff
@@ -0,0 +1,253 @@
+"""
```
piyifan123, line 1: add a comment on what it does.

veomni/models/transformers/deepseek_v3/modeling_deepseek.py
```diff
@@ -55,7 +57,7 @@
 
 
 if is_flash_attn_2_available():
```
piyifan123: delete

Luosuu: done

Fazziekey (approved):

PR #140, opened 2025-10-15
[ckpt] fix: rank for get last iteraton for non-dcp path

Luosuu:
For non-dcp path, `is_rank0` is used for downloading checkpoints to every node. 
Here we should use local_rank==0 instead of global_rank==0

### DIFF
--- a/veomni/utils/arguments.py
+++ b/veomni/utils/arguments.py
@@ -626,7 +626,7 @@ class TrainingArguments:
             from .checkpoint_utils import get_checkpoint_path
 
             self.load_checkpoint_path = get_checkpoint_path(
-                output_dir=self.output_dir, is_rank0=self.local_rank == 0, ckpt_manager=self.ckpt_manager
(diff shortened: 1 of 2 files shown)


veomni/utils/arguments.py
```diff
@@ -626,7 +626,7 @@ def __post_init__(self):
             from .checkpoint_utils import get_checkpoint_path
 
             self.load_checkpoint_path = get_checkpoint_path(
-                output_dir=self.output_dir, is_rank0=self.global_rank == 0, ckpt_manager=self.ckpt_manager
+                output_dir=self.output_dir, is_rank0=self.local_rank == 0, ckpt_manager=self.ckpt_manager
```
piyifan123: can you also change the parameter name to is_local_rank0?

piyifan123 (approved):

veomni/utils/arguments.py
Luosuu: sure

PR #141, opened 2025-10-16
[model] fix: remove Qwen3-MoE redundant flashattention prep and fix input_ids access bug

Juntian777:
Previous PR [#136](https://github.com/ByteDance-Seed/VeOmni/pull/136)
 already computed and added these FlashAttention arguments as the part of batch data, so there is no need to compute them again in the modeling code. This change also fixes a bug in the old FlashAttention2 preparation logic in modeling code where input_ids[0] was accessed even when input_ids was None. The issue occurred in vision-language models (InternVL3.5 MoE) where the language model (LM) part receives precomputed inputs_embeds instead of input_ids, causing a error.


piyifan123 (approved):
LGTM 

Please check if @Fazziekey has any other comments.

PR #145, opened 2025-10-18
[dist] feat: enable EP-aware optimizer for FSDP2-based MoE-VLM training.

Juntian777:
Removes the restriction that skipped EP-aware splitting when param_groups are provided, allowing VLMs with separate ViT learning rates to also use build_ep_fsdp2_optimizer. 

The updated logic cleanly partitions parameters into EP and non-EP groups, merges ViT subgroups with custom LRs into non-EP optimizers, and constructs a MultiOptimizer for proper checkpointing under FSDP2 + EP. 

Ensures both LLM and VLM pipelines share unified optimizer behavior while retaining per-group LR flexibility.

### DIFF
--- a/veomni/optim/optimizer.py
+++ b/veomni/optim/optimizer.py
@@ -275,7 +275,7 @@ def build_optimizer(
     weight_decay: float = 1e-2,
     fused: bool = False,
     optimizer_type: str = "adamw",
-    param_groups: Optional[Dict[str, Dict[str, Any]]] = None,
+    param_groups: Optional[Sequence[Dict[str, Any]]] = None,
     no_decay_modules: Optional[List[str]] = None,
     no_decay_params: Optional[List[str]] = None,
 ) -> "torch.optim.Optimizer":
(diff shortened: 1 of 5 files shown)


veomni/optim/optimizer.py
```diff
@@ -332,10 +329,26 @@ def build_ep_fsdp2_optimizer(
 ):
     """
     Build a MultiOptimizer instance when model is parallelized with EP+FSDP2
+
+    param_groups is only provided for VLM (e.g., [{"params": vit_params, "lr": vit_lr}, {"params": other_params, "lr": default_lr}]):
+    - vit_params go directly to non_ep_groups with vit_lr
+    - other_params are split between ep_params and non_ep_params based on DTensor mesh
```
Luosuu: maybe we can make naming of param_groups more robust. Instead of `other_params`, maybe it should be named in something like `lm_params`

Luosuu: and maybe param_groups should be a dict instead of a list, something like

```python
{
"vit_params": {"params": vit_params, "lr": vit_lr},
"lm_params": {"params": lm_params, "lr": lm_lr}
}

```

veomni/optim/optimizer.py
```diff
@@ -332,10 +329,26 @@ def build_ep_fsdp2_optimizer(
...
     """
+    # Determine which params to iterate over
+    if param_groups is not None:
+        vit_group = param_groups[0]  # {"params": vit_params, "lr": vit_lr}
+        other_group = param_groups[1]  # {"params": other_params, "lr": default_lr}
```
Luosuu: after we make `param_groups` a dict, we can do assert on its key names to make this process more robust. Now we directly index into the list which is not a good practice imo.

Luosuu:
LGTM, please verify more models besides MoE-VLM before merge like qwen3-dense/moe

PR #153, opened 2025-10-23
perf: set reshard_after_forward to False for modules without MixedPrecision

Luosuu:
For modules sharded without mixed precision on purpose like gating modules, we set `reshard_after_forward` to false for them to avoid all-gather them in the backward

### DIFF
--- a/veomni/distributed/torch_parallelize.py
+++ b/veomni/distributed/torch_parallelize.py
@@ -307,6 +307,7 @@ def parallelize_model_fsdp2(
         fsdp_kwargs_without_mp = dict(fsdp_kwargs)
         fsdp_kwargs_without_mp.pop("mp_policy", None)
         # for high-precision modules, we do not reshard them after forward to avoid all-gather them in backward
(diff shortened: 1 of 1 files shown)


veomni/distributed/torch_parallelize.py
```diff
@@ -306,6 +306,8 @@ def parallelize_model_fsdp2(
         mp_ignored_classes = modules_to_ignore_in_mixed_precision
         fsdp_kwargs_without_mp = dict(fsdp_kwargs)
         fsdp_kwargs_without_mp.pop("mp_policy", None)
+        # for high-precision modules, we do not reshard them after forward to avoid all-gather them in backward
```
piyifan123, line 309: add a comment saying this will increase the VRAM usage (i.e., params from hp modules will stay during the entire fwd/bwd thus all layers will be there) so that please ensure that high precision modules does not have too many parameters.

piyifan123 (approved):

veomni/distributed/torch_parallelize.py
Luosuu: sure

Issue #157, opened 2025-10-28
About logging lineno of the xx_rank0 method


ValMystletainn:
### System Info

veomni 1.1.2
ubuntu 22.04
python 3.11

### Information

- [x] The official example scripts
- [ ] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

veomni inherit and modified the built in logging, add `info_rank0`, `warn_rank0`, `debug_rank0`, ...

however they are built like this:

```python
def info_rank0(self: "logging.Logger", *args, **kwargs) -> None:
    if int(os.getenv("LOCAL_RANK", "0")) == 0:
        self.info(*args, **kwargs)


logging.Logger.info_rank0 = info_rank0
```

so when set the logging format with lineno(like in `veomni/utils/logging.py` function `_configure_library_root_logger`), the lineno is from callstack of `self.info(*args, **kwargs)` rather than `logger.info_rank0` so get this:

<img width="2414" height="428" alt="Image" src="https://github.com/user-attachments/assets/f2b12739-47fd-415b-b5f5-995d0054973b" />

all of the lineno is `114` the lineno of `self.info`

### Expected behavior

I think call  like `self.info(*args, stack, **kwargs)` and get the lineno from the stack of `logger.info_rank0` could be better for providing more infomative logs, like:

<img width="2448" height="440" alt="Image" src="https://github.com/user-attachments/assets/2f2a7e7a-94e6-4ba0-b297-edb893f65d75" />

So I raise this issue to ask is it a bug or a feature in expectation

Luosuu:
Hi @ValMystletainn 

Thank you very much for the insightful report. Would you like to make a PR to fix this?

ValMystletainn:
> Hi [@ValMystletainn](https://github.com/ValMystletainn)
> 
> Thank you very much for the insightful report. Would you like to make a PR to fix this?

The fixing PR is done, it's wellcome to ask me to do any further changes.

PR #160, opened 2025-10-29
[logging] fix: the logging lineno of log_rank0

ValMystletainn:
### What does this PR do?

turn the stacklevel +1 to get the correct called position of `info_rank0`, `warning_rank0`, `debug_rank0` method.
fix issue #157 

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: ...
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `misc`, `ci`, `config`, `docs`, `core`, `data`, `dist`, `omni`, `logging`, `model`, `optim`, `ckpt`, `release`, `task`, `
  - If this PR involves multiple modules, separate them with `,` like `[megatron, torch, liger-kernals,fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, torch, liger-kernals] feat: dynamic batching`

### Test

can be tested by ci

### API and Usage Example

Any training/inference scripts that import the model with `logger.info_rank0` will see the logging info change

the minimum call can be just the import command like

```bash
python -m veomni.models
```

### Design & Code Changes

force the stacklevel up to 1/2

> [!IMPORTANT]

- [x] Apply [pre-commit checks](https://github.com/ByteDance-Seed/VeOmni/blob/main/CONTRIBUTING.md?plain=1#L22): `make commit && make style && make quality`
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/ByteDance-Seed/VeOmni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: cover by original tests

### DIFF
--- a/veomni/utils/logging.py
+++ b/veomni/utils/logging.py
@@ -111,6 +111,7 @@ def set_verbosity_info() -> None:
 
 def info_rank0(self: "logging.Logger", *args, **kwargs) -> None:
     if int(os.getenv("LOCAL_RANK", "0")) == 0:
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1
         self.info(*args, **kwargs)
 
 
@@ -119,6 +120,7 @@ logging.Logger.info_rank0 = info_rank0
 
 def debug_rank0(self: "logging.Logger", *args, **kwargs) -> None:
     if int(os.getenv("LOCAL_RANK", "0")) == 0:
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1
         self.debug(*args, **kwargs)
 
 
@@ -127,6 +129,7 @@ logging.Logger.debug_rank0 = debug_rank0
 
 def warning_rank0(self: "logging.Logger", *args, **kwargs) -> None:
     if int(os.getenv("LOCAL_RANK", "0")) == 0:
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1
         self.warning(*args, **kwargs)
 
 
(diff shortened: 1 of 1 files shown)


Luosuu:
@piyifan123 @Fazziekey please take a look

veomni/utils/logging.py
```diff
@@ -111,6 +111,7 @@ def set_verbosity_info() -> None:
 
 def info_rank0(self: "logging.Logger", *args, **kwargs) -> None:
     if int(os.getenv("LOCAL_RANK", "0")) == 0:
+        kwargs["stacklevel"] = kwargs.get("stacklevel", 1) + 1
```
piyifan123, line 114: IIUC, this will correct the stacklevel since this info_rank0 adds an additional stack level? anyway, please add comments on why this is needed and what problem will happen without this.

ValMystletainn: Another modification could be `kwargs["stacklevel"] = kwargs.get("stacklevel", 2)`, right?

there is no different between the two modification when user no given `stacklevel`. Both of them can get the correct lineno, showing the lineno of `info_rank0()`.  as show by the figure in #157 

So I think the key is the expected behavior, when the user explicit pass  `stacklevel=xx`.

From my perspective, I first see the `logger.info_rank0`, I would use it as an in-place replacement of `logger.info`. i.e. the rank0 printing should be same. so if there is a `logger.info(xxx, stacklevel=3)` in original code, (it maybe aiming at the get 3 upper callstack information for debugging),  the in-place relacement is `logger.info_rank0(xxx, stacklevel=3)`. So to get 3 upper from `info_rank0` caller, it should be `self.info(xxx, stacklevel=4)` to go out call stack of `def info_rank0` first. So I choose the +1 modification, rather than default value changing modification

piyifan123 (approved):

PR #162, opened 2025-10-30
fix: repeat kv bug in flash attention forward with ulysses

HaoyiZhu:
Here, the kv shape is of (bs, seq, nhead, ...), so we cannot use the repeat_kv function, which assumes an input shape of (bs, nhead, seq, ...)

### DIFF
--- a/veomni/ops/attention.py
+++ b/veomni/ops/attention.py
@@ -18,18 +18,6 @@ if is_seed_kernels_available():
 logger = logging.get_logger(__name__)
 
 
-def repeat_kv(hidden_states: torch.Tensor, n_rep: int) -> torch.Tensor:
-    """
-    This is the equivalent of torch.repeat_interleave(x, dim=1, repeats=n_rep). The hidden states go from (batch,
-    num_key_value_heads, seqlen, head_dim) to (batch, num_attention_heads, seqlen, head_dim)
-    """
-    batch, num_key_value_heads, slen, head_dim = hidden_states.shape
-    if n_rep == 1:
-        return hidden_states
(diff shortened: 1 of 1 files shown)


Juntian777:
@HaoyiZhu Could you add some comments about the shapes and remove the unused repeat_kv before merging?

veomni/ops/attention.py
```diff
@@ -81,8 +81,8 @@ def flash_attention_forward(
                 f"ulysses_size ({ulysses_size}) must be divisible by num_key_value_heads ({kv_head_num})"
             )
             n_repeat = ulysses_size // kv_head_num
-            key = repeat_kv(key, n_repeat)
-            value = repeat_kv(value, n_repeat)
+            key = torch.repeat_interleave(key, dim=2, repeats=n_repeat)
```
Juntian777, line 76: Add some comments about the shape

veomni/ops/attention.py
```diff
@@ -81,8 +81,8 @@ def flash_attention_forward(
                 f"ulysses_size ({ulysses_size}) must be divisible by num_key_value_heads ({kv_head_num})"
             )
             n_repeat = ulysses_size // kv_head_num
-            key = repeat_kv(key, n_repeat)
```
Juntian777, line 84: remove the unused repeat_kv function

HaoyiZhu: Done

veomni/ops/attention.py
HaoyiZhu: Done

PR #169, opened 2025-11-04
[model] perf: eliminate per-layer CPU-GPU sync in Qwen3-VL vision attention

Juntian777:
This PR removes per-layer CPU-GPU syncs in the Qwen3-VL vision attention path and aligns causal mask logic with Qwen3. We compute max_seqlen once per block instead of once per layer, and port Qwen3’s _update_causal_mask pipeline to ensure consistent behavior across FA2, SDPA, and cache types.
#### Compute max_seqlen once

Before: each vision attention layer did (cu_seqlens[1:] - cu_seqlens[:-1]).max() then .item(), causing one sync per layer.
After: compute max_seqlen once in the vision block, then pass it into every layer to eliminate per-layer CPU-GPU sync 

### Traces
Before:
img_v3_02rn_75c135f6-3642-4e11-89cb-cb77ecf7bdeh

After:
image

### DIFF
--- a/veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
+++ b/veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
@@ -26,11 +26,10 @@ import torch
 import torch.nn as nn
 import torch.nn.functional as F
 from transformers.activations import ACT2FN
-from transformers.cache_utils import Cache, DynamicCache, SlidingWindowCache, StaticCache
+from transformers.cache_utils import Cache, DynamicCache
 from transformers.generation import GenerationMixin
 from transformers.integrations import use_kernel_forward_from_hub
 from transformers.masking_utils import create_causal_mask
-from transformers.modeling_attn_mask_utils import AttentionMaskConverter
 from transformers.modeling_flash_attention_utils import FlashAttentionKwargs
 from transformers.modeling_layers import GradientCheckpointingLayer
 from transformers.modeling_outputs import BaseModelOutputWithPast, ModelOutput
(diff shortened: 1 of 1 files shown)


veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
```diff
@@ -873,6 +879,13 @@ def forward(
         if (input_ids is None) ^ (inputs_embeds is not None):
             raise ValueError("You must specify exactly one of input_ids or inputs_embeds")
 
+        if self.gradient_checkpointing and self.training:
```
piyifan123: what's this for? is it from upgstream transformers?

veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
```diff
@@ -907,6 +920,8 @@ def forward(
             position_ids=text_position_ids,
         )
 
+        causal_mask = self._update_causal_mask(attention_mask, inputs_embeds, cache_position, past_key_values)
```
piyifan123: same here?

veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
```diff
@@ -919,7 +934,7 @@ def forward(
         for layer_idx, decoder_layer in enumerate(self.layers):
             layer_outputs = decoder_layer(
                 hidden_states,
-                attention_mask=attention_mask,
+                attention_mask=causal_mask,
```
piyifan123: anyway, i don't think attention_mask is used when using with FA (IIUC, attention_mask is used with padding cases only). so it probably does not matter, but would like to see why we are adding this?

PR #173, opened 2025-11-05
fix: workaround duplicated AllGather for EP+FSDP2

Luosuu:
### What does this PR do?

Compute shared expert first to workaround the duplicated all-gather issue in EP+FSDP2, which seems to be a bug in PyTorch FSDP2.

Before: 

After:

### DIFF
--- a/veomni/models/transformers/deepseek_v3/modeling_deepseek.py
+++ b/veomni/models/transformers/deepseek_v3/modeling_deepseek.py
@@ -471,15 +471,19 @@ class DeepseekV3FusedMoE(DeepseekV3MoE):
         topk_idx, topk_weight = self.gate(hidden_states)
         hidden_states = hidden_states.view(-1, hidden_states.shape[-1])
 
-        final_hidden_states = torch.zeros(
-            (batch_size * sequence_length, hidden_dim), dtype=hidden_states.dtype, device=hidden_states.device
-        )
         # we compute shared expert first to workaround the duplicated AllGather issue when using EP+FSDP2
         # which seems to be a bug of fsdp2 fully_shard
         if self.config.n_shared_experts is not None:
-            final_hidden_states = final_hidden_states + self.shared_experts(identity)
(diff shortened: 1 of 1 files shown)


veomni/models/transformers/deepseek_v3/modeling_deepseek.py
```diff
@@ -472,11 +474,13 @@ def forward(self, hidden_states):
         final_hidden_states = torch.zeros(
```
piyifan123: can we delete this then? this is not fully equivalent before/after the change IIUC as the zero tensor is being added to the result,

veomni/models/transformers/deepseek_v3/modeling_deepseek.py
```diff
@@ -472,11 +474,13 @@ def forward(self, hidden_states):
...
             (batch_size * sequence_length, hidden_dim), dtype=hidden_states.dtype, device=hidden_states.device
         )
-        final_hidden_states = self.experts(hidden_states, routing_weights=topk_weight, selected_experts=topk_idx)
-
+        # we compute shared expert first to workaround the duplicated AllGather issue when using EP+FSDP2
+        # which seems to be a bug of fsdp2 fully_shard
         if self.config.n_shared_experts is not None:
             final_hidden_states = final_hidden_states + self.shared_experts(identity)
 
+        final_hidden_states = self.experts(hidden_states, routing_weights=topk_weight, selected_experts=topk_idx)
```
piyifan123: is this buggy? seems that the shared expert result is being overriden?

Luosuu: yes. thanks for pinpointing this out. fixed now.

veomni/models/transformers/deepseek_v3/modeling_deepseek.py
Luosuu: sure

PR #178, opened 2025-11-06
[model] feat: support qwen3 vl moe 

Juntian777:
This PR enables training Qwen3-VL MoE models with FSDP, Sequence Parallelism (Ulysses), and Expert Parallelism (EP). It eliminates per-layer CPU–GPU synchronization in variable-length ViT and text FlashAttention. In addition, it adds dummy image and video forward paths to prevent stalls when only a subset of ranks in the same FSDP group contain vision tokens.

### Key Changes
1.Enables FSDP, Ulysses, and EP (uses fused_moe_forward for improved performance and EP support).

2.Removes per-layer CPU–GPU sync in vision attention by computing max_seqlen once in the vision block and passing it to every layer.

3.Pre-computes variable-length FlashAttention kwargs from 3D position IDs to eliminate per-layer CPU–GPU sync in text attention.

4.Adds sequence boundary detection, padding, and slicing for Sequence Parallelism compatibility.

5.Adds dummy ViT forward paths to prevent hangs in FSDP backward reduce-scatter.

### Future Works
grad_norm aligns correctly for FSDP1+EP+SP and FSDP2+EP and other cases, but there’s a mismatch when using FSDP2 + EP + SP simultaneously. We should fix this case later.

### DIFF
--- a/tasks/omni/train_qwen_vl.py
+++ b/tasks/omni/train_qwen_vl.py
@@ -128,6 +128,7 @@ def main():
         weights_path=args.model.model_path,
         init_device=args.train.init_device,
         force_use_huggingface=args.model.force_use_huggingface,
+        moe_implementation=args.model.moe_implementation,
         attn_implementation=args.model.attn_implementation,
     )
     model_config = model.config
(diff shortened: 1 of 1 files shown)


Luosuu:
I think it is fine to have EP parallel_plan first although the grad norm issue with EP has not been resolved. But at least EP is working

PR #182, opened 2025-11-08
feat: distributed checkpointer support customized backend

Ziyi-Wang:
### What does this PR do?

support customized backend for distributed checkpointing

### DIFF
--- a/veomni/checkpoint/dcp_checkpointer.py
+++ b/veomni/checkpoint/dcp_checkpointer.py
@@ -265,6 +265,7 @@ class DistributedCheckpointer(CheckpointerBase):
         state: Dict[str, Any],
         save_async: bool = False,
         global_steps: int = None,
+        storage_writer: Optional[FileSystemWriter] = None,
     ) -> None:
         """
         save training state to distributed checkpoint
@@ -272,7 +273,9 @@ class DistributedCheckpointer(CheckpointerBase):
         args:
             path: path to save checkpoint
             state: state to save
+            save_async: whether to save asynchronously
             global_steps: global steps
+            storage_writer: storage writer backend for dcp.save and dcp.async_save. If None, will use FileSystemWriter
         return:
             None
         """
@@ -297,6 +300,14 @@ class DistributedCheckpointer(CheckpointerBase):
         if "optimizer" in state:
             save_state["optimizer"] = OptimizerState(model=state["model"], optimizer=state["optimizer"])  # type: ignore[index]
 
+        if storage_writer is None:
+            storage_writer = FileSystemWriter(
+                checkpoint_dir,
+                thread_count=16,
+                single_file_per_rank=True,
+                sync_files=False,
+            )
+
         if save_async:
             # Lazily create a dedicated Gloo process group for async DCP saves
             if cls._async_process_group is None:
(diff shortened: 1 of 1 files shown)


Luosuu:
how do we pass the dcp backend now?
I think we need to add new arguments in `helper.py` and also the trainers correspondingly?

veomni/checkpoint/checkpointer.py
```diff
@@ -304,6 +304,7 @@ def save(
         state: Dict[str, Any],
         save_async: bool = False,
         global_steps: int = None,
+        storage_writer: Optional[FileSystemWriter] = None,
```
Luosuu: Is FileSystemWriter the correct type?

Ziyi-Wang: Yes, the storage_writer must be an instance of FileSystemWriter or a derived class.

Ziyi-Wang:
> how do we pass the dcp backend now? I think we need to add new arguments in `helper.py` and also the trainers correspondingly?

The current code is forward compatible, and it uses the FileSystemReader/Writer by default.

veomni/checkpoint/checkpointer.py
```diff
@@ -348,25 +349,31 @@ def save(
                 # block until all the ranks resolve their previous dcp async saving
                 dist.barrier()
 
-            cls.dcp_save_future = dcp.async_save(
-                state_dict=save_state,
-                storage_writer=FileSystemWriter(
+            if storage_writer is None:
```
piyifan123: can this logic be moved to L339? why replicate twice?

veomni/checkpoint/checkpointer.py
piyifan123: nit: update comment about this function. and note what default value None would do?

veomni/utils/arguments.py
```diff
@@ -521,6 +521,10 @@ class TrainingArguments:
         default=None,
         metadata={"help": "Max training steps per epoch. (for debug)"},
     )
+    dcp_backend: str = field(
```
piyifan123: nit: dcp_storage_backend

veomni/utils/arguments.py
```diff
@@ -521,6 +521,10 @@ class TrainingArguments:
...
+        default="default",
+        metadata={"help": "Backend for distributed checkpointing. Uses FileSystemWriter/FileSystemReader by default."},
```
piyifan123: nit: add note that only the default FSW is supported now and we will support other storage backends like VolcEngine TOS in the future

veomni/checkpoint/checkpointer.py
Ziyi-Wang: Updated. Added documentation for `storage_writer` parameter including the default `FileSystemWriter` fallback behavior.

veomni/utils/arguments.py
```diff
@@ -521,9 +521,9 @@ class TrainingArguments:
         default=None,
         metadata={"help": "Max training steps per epoch. (for debug)"},
     )
-    dcp_backend: str = field(
+    dcp_storage_backend: str = field(
         default="default",
```
Luosuu: leaving a comment to block PR since we are still testing internally

veomni/utils/arguments.py
```diff
@@ -521,6 +521,10 @@ class TrainingArguments:
         default=None,
         metadata={"help": "Max training steps per epoch. (for debug)"},
     )
+    dcp_storage_backend: str = field(
```
piyifan123: delete this and add until we have a usage?

veomni/checkpoint/checkpointer.py
```diff
@@ -377,6 +377,7 @@ def load(
         path: str,
         state: Dict[str, Any],
         process_group=None,
+        storage_reader: Optional[FileSystemReader] = None,
```
piyifan123: nit: update docstring about this new argument what None means

piyifan123 (approved):

Issue #184, opened 2025-11-10
fsdp2 based hsdp, caculate grad_norm error in `clip_grad_norm`


ValMystletainn:
### System Info

veomni 0.1.3
ubuntu 20.04
python 3.11.4
cuda 12.6

### Information

- [x] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

the grad norm is caculate by sum the p-th power of parameter elements, first in local than all-reduce with sum in `fsdp_group`, in current version, it is `dp_sp` group. However I think it should only do in `dp_shard` group.

It cause the grad norm mismatch when use large dp_replicate size, in my training (a from scratch model). and I just write a minimum demo scripts that can reproduce the norm error with only a node with 8 gpu.

```python
# reproduce_hsdp_clip_grad.py

from dataclasses import dataclass, field

import torch
import torch.distributed as dist

from veomni.distributed.parallel_state import init_parallel_state
from veomni.distributed.torch_parallelize import build_parallelize_model
from veomni.utils import helper
from veomni.utils.arguments import TrainingArguments, parse_args
from veomni.utils.device import get_device_type, get_nccl_backend, get_torch_device


logger = helper.create_logger(__name__)

@dataclass
class Argument:
    train: "TrainingArguments" = field(default_factory=TrainingArguments)

class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.bias = torch.nn.Parameter(torch.ones(16), requires_grad=True)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        loss = (x + self.bias).sum()
        return loss

    def init_weights(self):
        self.bias.data.fill_(1.0)

def main():
    dist.init_process_group(backend=get_nccl_backend())
    args = parse_args(Argument)

    get_torch_device().set_device(f"{get_device_type()}:{args.train.local_rank}")
    init_parallel_state(
        dp_size=args.train.data_parallel_size,
        dp_replicate_size=args.train.data_parallel_replicate_size,
        dp_shard_size=args.train.data_parallel_shard_size,
        tp_size=args.train.tensor_parallel_size,
        ep_size=args.train.expert_parallel_size,
        pp_size=args.train.pipeline_parallel_size,
        cp_size=args.train.context_parallel_size,
        ulysses_size=args.train.ulysses_parallel_size,
[...]

PR #185, opened 2025-11-10
[dist] fix: refactor fsdp2 grad norm clipping

Luosuu:
Simplify fsdp2 grad norm clipping implementation to use DTensor-native approach instead of manually reducing them.

This also makes pure fsdp2 setup (including HSDP) work out-of-box.

related issue: #184 

Pure FSDP2 verification:

For EP+FSDP2, we use the native `set_gradient_divide_factor` for experts now.
But currently we still have to manually reduce gradients because experts weight dtensor has no knowledge about their EP dimension.

EP+FSDP2 verification:

Note: This PR also updates the uv and pytorch version

### DIFF
--- a/veomni/distributed/torch_parallelize.py
+++ b/veomni/distributed/torch_parallelize.py
@@ -339,6 +339,7 @@ def parallelize_model_fsdp2(
         if parallel_state.ep_enabled and experts_mod is not None:
             # shard expert
             fully_shard(experts_mod, **expert_fsdp_kwargs)
+            # average EP grads across EP ranks
             experts_mod.set_gradient_divide_factor(parallel_state.ep_size)
             layer_mod._fsdp_modules.append(experts_mod)
         # shard module that needs to ignore mixed precision control


veomni/distributed/torch_parallelize.py
```diff
@@ -339,9 +339,7 @@ def _experts_shard_placement_fn(param):
         if parallel_state.ep_enabled and experts_mod is not None:
             # shard expert
             fully_shard(experts_mod, **expert_fsdp_kwargs)
-            if hasattr(experts_mod, "set_gradient_divide_factor"):
-                # average EP grads across EP ranks
-                experts_mod.set_gradient_divide_factor(parallel_state.ep_size)
+            experts_mod.set_gradient_divide_factor(parallel_state.ep_size)
```
piyifan123, line 343: nit: add a comment to explain why we need to set a factor here.

Luosuu: sure

Issue #193, opened 2025-11-13
raise ImportError veomni_patch is not available


iqiancheng:
```sh
veomni/utils/helper.py", line 82, in save_step2token
    raise ImportError("veomni_patch is not available, please install it first")
ImportError: veomni_patch is not available, please install it first
```

Salieri0515:
same problem

iqiancheng:
```diff
diff --git a/veomni/utils/helper.py b/veomni/utils/helper.py
index 57f65c9..3550cfe 100644
--- a/veomni/utils/helper.py
+++ b/veomni/utils/helper.py
@@ -76,16 +76,23 @@ if is_veomni_patch_available():
 else:
 
     def load_step2token(*args, **kwargs):
-        raise ImportError("veomni_patch is not available, please install it first")
+        # No-op when veomni_patch is not available
+        pass
 
     def save_step2token(*args, **kwargs):
-        raise ImportError("veomni_patch is not available, please install it first")
+        # No-op when veomni_patch is not available
+        pass
 
     def is_remote_path(*args, **kwargs):
-        raise ImportError("veomni_patch is not available, please install it first")
+        # Return False as safe default when veomni_patch is not available
+        return False
 
     def convert_hdfs_fuse_path(*args, **kwargs):
-        raise ImportError("veomni_patch is not available, please install it first")
+        # Return path as-is when veomni_patch is not available
+        # This works for non-HDFS paths, but may fail for HDFS paths
+        if len(args) > 0:
```

i fixed the `helper.py` compatiable without the veomni_patch, you can try it out, just let me know is it work well.

Salieri0515:
I met this problem when training qwen2.5 with output `[INFO][veomni:30] 11/13/2025 15:18:31 >> ❌ veomni_patch is not available`.  The modification above seems no use fixing this warning. I think this package may not be necessary for training.

Luosuu:
yes indeed this is just a package for our internal utils. please simply remove them if you see such errors. Thanks for reporting

Luosuu:
btw, warning is fine.

iqiancheng:
> btw, warning is fine.

totally agree with you! warn is much better.

```diff
diff --git a/veomni/utils/helper.py b/veomni/utils/helper.py
index 57f65c9..3550cfe 100644
--- a/veomni/utils/helper.py
+++ b/veomni/utils/helper.py
@@ -76,16 +76,27 @@ if is_veomni_patch_available():
[...]

PR #197, opened 2025-11-14
helper: degrade veomni_patch functions to warnings/no-op

iqiancheng:
Fix ImportError when veomni_patch import fails during model checkpoint saving #193 

## Problem
When saving model checkpoints at the end of training, an ImportError is raised:
```sh
veomni/utils/helper.py", line 82, in save_step2token
[rank0]:     raise ImportError("veomni_patch is not available, please install it first")
[rank0]: ImportError: veomni_patch is not available, please install it first
```
This occurs even when `is_veomni_patch_available()` returns `True`, but the actual import from `veomni_patch.utils.helper` fails (e.g., due to missing dependencies or partial installation).

## Solution
Wrap the `veomni_patch` import in a try-except block to gracefully handle import failures. If the import fails, the code falls back to no-op stub functions that log warnings instead of raising exceptions, allowing training to complete successfully even when `veomni_patch` is not fully available.

## Changes
- Added try-except block around the `veomni_patch.utils.helper` import in `veomni/utils/helper.py`
- Ensures `save_step2token` and other helper functions gracefully degrade when `veomni_patch` is unavailable
- Prevents training from crashing during checkpoint saving when `veomni_patch` import fails

### DIFF
--- a/veomni/utils/helper.py
+++ b/veomni/utils/helper.py
@@ -76,16 +76,22 @@ if is_veomni_patch_available():
 else:
 
     def load_step2token(*args, **kwargs):
-        raise ImportError("veomni_patch is not available, please install it first")
+        logger.warning("veomni_patch is not available, load_step2token will be skipped")
+        pass
 
     def save_step2token(*args, **kwargs):
-        raise ImportError("veomni_patch is not available, please install it first")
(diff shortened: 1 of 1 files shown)


Coach257:
TODO: clean helper @Coach257 @FoolPlayer

PR #216, opened 2025-11-25
[optim, config] feat: add support for Muon optimizer via dion

clarkipeng:
This PR introduces support for the Muon optimizer (Momentum Orthogonalized Optimizer) using the dion backend.
It implements a hybrid optimization strategy where:

Muon is applied to high-dimensional parameters (2D+ matrices), excluding embeddings and the LM head, and with rms-normalization.

AdamW is used as the fallback for 1D tensors (vectors), embeddings, and the LM head.

This implementation requires DeviceMesh for distributed training and adds dion as a project dependency.

### Test
Screenshot 2025-11-24 at 6 59 38 PM

Validated by running qwen3_vl_8b with optimizer="muon" on a 2xH100 node with seq-len = 32784. 

(adamw ooms with same hyperparams)

### API and Usage Example

To use the Muon optimizer, simply update the optimizer argument in your configuration or CLI. No extra configuration is needed for the parameter splitting logic; it is handled automatically.

### Design & Code Changes

This PR modifies how parameter groups are constructed in `build_optimizer` without fsdp2. Before, if the parameter groups were created outside of the ```build_optimizer``` call without fsdp2, weight decay wouldn't be applied. Now, muon (if_enabled) and weight decay are automatically applied to any parameter group passed into ```build_optimizer```.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ gpu = [
   "liger-kernel",
   "flash-attn",
   "hf_transfer",
+  "dion",
 ]
 megatron = [
   "megatron-energon>=7.2.1"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,6 +152,7 @@ torchvision = [ # npu aarch64 need install torchvison from pytorch-cpu
 torchaudio = { index = "pytorch" }
 # Download flash-attn wheel directly to avoid build issues.
 flash-attn = { url = "https://github.com/Dao-AILab/flash-attention/releases/download/v2.8.3/flash_attn-2.8.3+cu12torch2.8cxx11abiTRUE-cp311-cp311-linux_x86_64.whl", marker = "extra == 'gpu' or extra == 'sglang'" }
+dion = [{ git = "https://github.com/microsoft/dion.git", rev = "0360f9b" }]
 
 [[tool.uv.index]]
 name = "pytorch"
--- a/veomni/optim/optimizer.py
+++ b/veomni/optim/optimizer.py
@@ -19,6 +19,7 @@ from typing import Any, Dict, Iterable, List, Optional, Sequence, Tuple
 
 import torch
 import torch.nn as nn
+from dion import Muon
 from torch.distributed._tensor import DTensor
 from torch.distributed.checkpoint.state_dict import (
     StateDictOptions,
@@ -239,6 +240,17 @@ def _make_param_groups_for_subset(
     return groups
 
 
+def use_muon(name: str, param: torch.Tensor) -> bool:
+    """Determine if a parameter should use Muon optimizer."""
+    if param.ndim < 2:
+        return False
+    if "lm_head" in name:
+        return False
+    if "embed_tokens" in name:
+        return False
+    return True
+
+
 # adapted from https://github.com/huggingface/transformers/blob/v4.49.0/src/transformers/trainer_pt_utils.py#L1123
 def get_parameter_names(model, forbidden_layer_types, forbidden_param_names):
     forbidden_layer_types = [] if forbidden_layer_types is None else forbidden_layer_types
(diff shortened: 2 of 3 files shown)


Luosuu:
I think in torch 2.9 muon is supported?
https://docs.pytorch.org/docs/stable/generated/torch.optim.Muon.html

clarkipeng:
Oh good point, I'll give it a go

clarkipeng:
@Luosuu Would we want to use torch 2.9? Then we probably need to rewrite many of the optimizations that [dion](https://github.com/microsoft/dion/blob/main/dion/muon.py) does.

Did a quick experiment. If I run a naive MultiOptimizer with muon and adamw the runtime is far worse... (Qwen3-1.7B with 8xH100)
W B Chart 12_13_2025, 1_52_37 AM
W B Chart 12_13_2025, 1_53_02 AM
W B Chart 12_13_2025, 1_53_16 AM

Luosuu:
Interesting, so you mean torch 2.9 muon is numerically same with dion muon but slower?

clarkipeng:
> Interesting, so you mean torch 2.9 muon is numerically same with dion muon but slower?

Yes, particularly in the parallel setting. Dion also implements triton kernels for newton-shutlz so it could be faster than shown above

Luosuu:
I am curious whether the speed diff would still be large for larger models. I thought optimizer step time would not be a very large portion in the end to end training.

VeOmni provides trace profiling by pytorch profiler which can be enabled by `enable_profiling: true`.
Could you please show the trace with dion muon and torch 2.9 muon on a larger model so that we can see the diff more clearly? Thank you very much.

clarkipeng:
Sure!
For reference my 2.9 Muon implementation is the following naive approach
```
optimizers = {}
foreach = not fused

muon_groups = [g for g in param_groups if g.get("algorithm") == "muon"]
optimizers["muon"] = Muon(muon_groups, lr=lr, weight_decay=weight_decay, eps=eps)
            
adamw_groups = [g for g in param_groups if g.get("algorithm") == "adamw"]
optimizers["adamw"] = AdamW(adamw_groups, lr, betas, eps, weight_decay, fused=fused, foreach=foreach)
            
optim = MultiOptimizer(model, optimizers, list(optimizers.keys()))
```

Here are the results on Qwen3-8B, the difference does seem to decrease.

W B Chart 12_13_2025, 11_39_41 PM

# Traces
## dion_muon:
```
  {
    "ph": "X", "cat": "user_annotation", "name": "Optimizer.step#Muon.step", "pid": 1626701, "tid": 1626701,
    "ts": 6394723773641.521, "dur": 125112.171,
    "args": {
      "External id": 145593,"Record function id": 0, "Ev Idx": 47800
    }
  },
  {
    "ph": "X", "cat": "python_function", "name": "dion/muon.py(130): step", "pid": 1626701, "tid": 1626701,
    "ts": 6394723773678.061, "dur": 125013.852,
    "args": {
      "Python parent id": 1591820, "Python id": 1591831, "Ev Idx": 1792869
    }
  },
```
## 2.9_muon:
```
  {
    "ph": "X", "cat": "user_annotation", "name": "Optimizer.step#Muon.step", "pid": 1637316, "tid": 1637316,
    "ts": 6396247796158.178, "dur": 2548154.907,
    "args": {
      "External id": 145593,"Record function id": 0, "Ev Idx": 47800
    }
  },
  {
    "ph": "X", "cat": "python_function", "name": "torch/optim/_muon.py(166): step", "pid": 1637316, "tid": 1637316,
    "ts": 6396247796195.376, "dur": 2547985.927,
    "args": {
      "Python parent id": 1591688, "Python id": 1591699, "Ev Idx": 1935281
    }
  },
```

clarkipeng:
@Luosuu Any thoughts?

Luosuu:
Dion looks cool but I am suspicious about whether Microsoft would maintain it for long term, especially considering that its functionality is overlapped by official pytorch. 
Thank you very much for your report but my personal opinion is to support muon by upgrading to torch 2.9

Luosuu:
I am requesting another reviewer for this for more thoughts on this

PR #228, opened 2025-11-28
[dist] fix: make OptimizerState EP-dim aware to fix its dcp saving

Luosuu:
Similar to `ModelState`, `OptimizerState` also needs to be EP-dim aware so that dcp can save it properly.
Now it has same workflow with `ModelState`:
* before saving with dcp, restore EP dim
* after loading state dict from dcp, drop EP dim


configs/model_configs/qwen/qwen3_moe_30a3b_4_layers.json
```diff
@@ -0,0 +1,38 @@
+{
+  "architectures": [
```
piyifan123, line 2: nit: rename it to qwen3_moe_30a3b_4_layers.json?

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -21,8 +26,8 @@
 
 
 """
-torchrun --nnodes=1 --nproc-per-node=8 --master-port=4321 tests/utils/test_trainer_saveload.py \
---model.model_path Qwen/Qwen3-4B \
+torchrun --nnodes=1 --nproc-per-node=8 --master-port=4321 tests/checkpoints/test_trainer_saveload.py \
```
piyifan123, line 25: are these commands duplicate with the following commands? if so, please delete those.

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -36,22 +41,24 @@
...
-torchrun --nnodes=1 --nproc-per-node=8 --master-port=4321 tests/utils/test_trainer_saveload.py \
---model.model_path /path/to/Qwen3-30B-A3B-Instruct-2507-merge \
+torchrun --nnodes=1 --nproc-per-node=8 --master-port=4321 tests/checkpoints/test_trainer_saveload.py \
+--model.config_path configs/model_configs/qwen/Qwen3Moe_4_layers.json \
+--model.weight_path None \
+--model.tokenizer_path /model-b/Qwen/Qwen3-30B-A3B \
 --model.moe_implementation fused \
 --model.attn_implementation flash_attention_2 \
---train.expert_parallel_size 4 \
+--train.expert_parallel_size 8 \
```
piyifan123, line 31: can we keep ep=4 so that with 8 ranks, we still have 8/4=2 sharding dimension?

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -314,6 +321,41 @@ def main():
                     "train_dataloader": train_dataloader.state_dict(),
                 },
             }
+            if getattr(optimizer, "_is_multi_optimizer", False):
```
piyifan123: what does this if block do? just checks there is no overlapping between ep/non-ep optimizer state? if so, can we write a helper method like check_ep_non_ep_optimizer_state_are_disjoint(...) so that it's self explanary?

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -314,6 +321,41 @@ def main():
...
+                non_ep_keys = set(flatten_non_ep_optim_sd.keys())
+                merged_keys = set(optimizer.state_dict().keys())
+
+                # Optional sanity check: EP and non-EP partitions are disjoint
+                overlap = ep_keys & non_ep_keys
+                assert not overlap, f"EP and non-EP optimizers share keys: {sorted(overlap)}"
+
+                # The MultiOptimizer should cover exactly the union
+                expected_keys = ep_keys | non_ep_keys
+                assert merged_keys == expected_keys, (
```
piyifan123: is the correctness of loaded saved checkpoint checked in L390/391?

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -356,5 +398,61 @@ def step_id(s):
     dist.destroy_process_group()
 
 
+def test_trainer_saveload():
```
piyifan123: can we split ep=8 and ep=4 into 2 pytest test cases and make the name self explained? e.g., test_trainer_save_load_qwen3_moe_ep8()?

piyifan123: BTW, can we also add a test case for HSDP+EP? maybe in a follow up PR?

veomni/checkpoint/checkpointer.py
```diff
@@ -188,31 +144,188 @@ class OptimizerState(Stateful):
...
+            # Find EP spec whose FQN appears in the state_dict key
+            # e.g. name = "state.model.layers.0.mlp.experts.gate_proj.step"
+            #      ep_key = "model.layers.0.mlp.experts.gate_proj"
+            matches = [ep_key for ep_key in ep_keys if ep_key in name]
+            if not matches:
+                # ignore non-ep tensor
+                continue
+
+            # each tensor in the state dict should only belong to one EP entry
+            if len(matches) > 1:
```
piyifan123: nit: change this to assert?

veomni/checkpoint/checkpointer.py
```diff
@@ -188,31 +144,188 @@ class OptimizerState(Stateful):
...
 
         # Single torch optimizer
         set_optimizer_state_dict(
             model=self.model,
             optimizers=self.optimizer,
-            optim_state_dict=optim_state,
+            optim_state_dict=optim_state_from_dcp_load,
+        )
+
+    def get_state_dict_with_ep_dim(self, state_dict):
```
piyifan123: nit: add docstring on how this is done.

veomni/checkpoint/checkpointer.py
```diff
@@ -188,31 +144,188 @@ class OptimizerState(Stateful):
...
+
+            # each tensor in the state dict should only belong to one EP entry
+            if len(matches) > 1:
+                raise RuntimeError(f"Ambiguous EP spec match for state key '{name}': {matches}")
+
+            ep_key = matches[0]
+            cur_spec_info = ep_fqn2spec_info[ep_key]
+
+            # skip non-ep params which have Replacate placement in our parallel plan spec info
+            if not isinstance(cur_spec_info.placement, Shard):
```
piyifan123: why is this needed? did we already match the name through ep_keys?

veomni/checkpoint/checkpointer.py
```diff
@@ -188,31 +144,188 @@ class OptimizerState(Stateful):
...
+                continue
+
+            tensor = state_dict[name]
+
+            # we skip param-group hyperparams like `param_groups.model.layers.0.mlp.experts.down_proj.amsgrad`
+            if not torch.is_tensor(tensor):
+                continue
+            # Skip scalars (0-D tensors) – cannot be sharded on dim 0 like `step`
```
piyifan123: do not understand on what scalars can reach there.

veomni/checkpoint/checkpointer.py
```diff
@@ -188,31 +144,188 @@ class OptimizerState(Stateful):
...
+            tensor = restore_ep_dim(tensor, cur_spec_info.ep_fsdp_mesh)
+            state_dict[name] = tensor
+
+        return state_dict
+
+    def get_state_dict_without_ep_dim(self, state_dict):
+        ep_fqn2spec_info = self.ep_fqn2spec_info
+        assert ep_fqn2spec_info is not None, "if fqn2spec_info is None it should not be patch"
+
+        ep_mesh = self.parallel_state.ep_fsdp_device_mesh["ep"]
```
piyifan123: looks like L238 to L276 can be shared between these 2 functions?

veomni/checkpoint/checkpointer.py
```diff
@@ -107,7 +107,7 @@ def get_state_dict_with_ep_dim(self, state_dict):
             if name in ep_fqn2spec_info and isinstance(ep_fqn2spec_info[name].placement, Shard):
                 cur_spec_info = ep_fqn2spec_info[name]
                 tensor = state_dict[name]
-                tensor = self._restore_ep_dim(tensor, cur_spec_info.ep_fsdp_mesh)
+                tensor = restore_ep_dim(tensor, cur_spec_info.ep_fsdp_mesh)
```
piyifan123: why is the finding ep params much simpler for model param states? BTW, can we unify the matching logic between model param and optimizer states?

veomni/checkpoint/checkpointer.py
Luosuu: `state.model.layers.0.mlp.experts.down_proj.step`

veomni/checkpoint/checkpointer.py
Luosuu: because ep params can perfetchly match the keys in the model spec info.
In the contrast, ep param names are substring of flattened model params state names, for example:
* ep param name: `model.layers.0.mlp.experts.down_proj`
* model param states: 
    * `state.model.layers.0.mlp.experts.down_proj.step` (a scalar tensor that cannot be sharded)
    * `state.model.layers.0.mlp.experts.down_proj.exp_avg`
    * `state.model.layers.0.mlp.experts.down_proj.exp_avg_sq`
    * ....

tests/checkpoints/test_trainer_saveload.py
Luosuu: sure

tests/checkpoints/test_trainer_saveload.py
Luosuu: actually we have the same check in the MultiOptimizer class state_dict(), I will simply remove these logs here

tests/checkpoints/test_trainer_saveload.py
Luosuu: checked inside MultiOptimizer. I will remove these logs here

tests/checkpoints/test_trainer_saveload.py
Luosuu: done

configs/model_configs/qwen/qwen3_moe_30a3b_4_layers.json
Luosuu: sure

veomni/checkpoint/checkpointer.py
Luosuu: done

veomni/checkpoint/checkpointer.py
Luosuu: deleted

veomni/checkpoint/checkpointer.py
Luosuu: done

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -172,31 +113,141 @@ class OptimizerState(Stateful):
...
 
         # Single torch optimizer
         set_optimizer_state_dict(
             model=self.model,
             optimizers=self.optimizer,
-            optim_state_dict=optim_state,
+            optim_state_dict=optim_state_from_dcp_load,
+        )
+
+    def get_state_dict_with_ep_dim_preprocess(self, state_dict, action):
```
piyifan123, line 153: nit: add docstring for this non-trivial function.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -172,31 +113,141 @@ class OptimizerState(Stateful):
...
+        assert global_device_mesh.ndim == 2
+
+        assert action in ["drop", "restore"]
+
+        keys = list(state_dict.keys())
+        ep_keys = list(ep_fqn2spec_info.keys())
+
+        for name in sorted(keys):
+            # Find EP spec whose FQN appears in the state_dict key
+            # e.g. name = "state.model.layers.0.mlp.experts.gate_proj.step"
```
piyifan123, line 170: nit: can you give a full example in docstring of this function to list all keys can be matched and their corresponding types? and then it's much easier to see how the filtering here works.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -172,31 +113,141 @@ class OptimizerState(Stateful):
...
+            if action == "drop":
+                tensor = drop_ep_dim(tensor, cur_spec_info.ep_fsdp_mesh)
+            elif action == "restore":
+                tensor = restore_ep_dim(tensor, cur_spec_info.ep_fsdp_mesh)
+            state_dict[name] = tensor
+
+        return state_dict
+
+
+def drop_ep_dim(loaded_tensor: torch.Tensor, device_mesh: DeviceMesh):
```
piyifan123, line 205: can this be private function as well?

veomni/checkpoint/dcp_checkpointer.py
piyifan123, line 153: is this function a member of OptimizerState class? how does ModelState be able to call it via self?

anyway, maybe make both top level private function under dcp_checkpointer.py?

piyifan123 (approved):

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -143,8 +131,9 @@ def main():
     train_dataset = build_dummy_dataset(task_type="text", size=train_data_size, max_seq_len=args.data.max_seq_len)
 
     args.train.compute_train_steps(args.data.max_seq_len, args.data.train_size)
-    train_dataloader = build_streaming_dataloader(
+    train_dataloader = build_dataloader(
         dataset=train_dataset,
+        dataloader_type="streaming",
```
Coach257, line 136: Please use native dataloader.

PR #248, opened 2025-12-02
[ckpt] refactor: remove unused output_dir parameter from ckpt_to_state_dict

TimYangst:
### What does this PR do?

> Fix `TypeError` in `dcp_ckpt_to_state_dict()` by adding missing `output_dir` parameter to match the function signature required by `ckpt_to_state_dict` interface.

### Checklist Before Starting

- [ ] Format the PR title as `[ckpt] fix: add missing output_dir parameter to dcp_ckpt_to_state_dict`

### Test

> This fix resolves the runtime error encountered during checkpoint conversion to HuggingFace format. The error occurred when training completed and attempted to save checkpoints:
> ```
> TypeError: dcp_ckpt_to_state_dict() takes 1 positional argument but 2 were given
> ```

### API and Usage Example

> No API changes for end users. The fix ensures internal consistency between the `ckpt_to_state_dict` interface and its registered implementation.

### Design & Code Changes

**Root Cause:**
- The `ckpt_to_state_dict` interface function ([checkpointer.py:51](veomni/checkpoint/checkpointer.py#L51)) calls registered converters with two positional arguments: `save_checkpoint_path` and `output_dir`
- The `dcp_ckpt_to_state_dict` implementation only accepted one positional argument, causing a signature mismatch

**Changes:**
- Updated `dcp_ckpt_to_state_dict` function signature to explicitly accept `output_dir` parameter
- Added documentation comment explaining that `output_dir` is kept for interface consistency but not used in DCP conversion (state_dict is loaded directly into memory)
- Preserved `**kwargs` for future extensibility

- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/ByteDance-Seed/VeOmni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: This is a signature fix for an existing function. The existing integration tests in training tasks will validate the fix.

### DIFF
--- a/veomni/checkpoint/checkpointer.py
+++ b/veomni/checkpoint/checkpointer.py
@@ -33,7 +33,6 @@ def build_checkpointer(ckpt_manager: str, dist_backend: str):
 
 def ckpt_to_state_dict(
     save_checkpoint_path: Union[str, os.PathLike],
-    output_dir: Union[str, os.PathLike],
     ckpt_manager: str = "dcp",
 ) -> Dict[str, Any]:
     """
@@ -43,12 +42,11 @@ def ckpt_to_state_dict(
 
     Args:
         save_checkpoint_path: Path to the checkpoint.
-        output_dir: Path to the output directory.
         ckpt_manager: Checkpoint manager.
     Returns:
         state_dict: State dict.
     """
-    return CHECKPOINT_TO_STATE_DICT_REGISTRY[ckpt_manager](save_checkpoint_path, output_dir)
+    return CHECKPOINT_TO_STATE_DICT_REGISTRY[ckpt_manager](save_checkpoint_path)
 
 
 class CheckpointerBase(ABC):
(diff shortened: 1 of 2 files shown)


veomni/checkpoint/checkpointer.py
```diff
@@ -89,11 +89,15 @@ def dcp_checkpointer(dist_backend: str):
 
 
 @CHECKPOINT_TO_STATE_DICT_REGISTRY.register("dcp")
-def dcp_ckpt_to_state_dict(save_checkpoint_path: Union[str, os.PathLike], **kwargs):
+def dcp_ckpt_to_state_dict(
+    save_checkpoint_path: Union[str, os.PathLike], output_dir: Union[str, os.PathLike], **kwargs
```
piyifan123: since output_dir is not used, why do we add it back?

TimYangst: Done

PR #256, opened 2025-12-05
[model] feat: add AutoModelForSequenceClassification support in HuggingfaceLoader

yiwzhao:
I added AutoModelForSequenceClassification support in our HuggingfaceLoader so that whenever a config’s architectures indicate a ForSequenceClassification class, we now instantiate the corresponding HF classification model instead of falling back to AutoModel.

### DIFF
--- a/veomni/models/transformers/qwen3/modeling_qwen3.py
+++ b/veomni/models/transformers/qwen3/modeling_qwen3.py
@@ -27,7 +27,7 @@ from transformers.utils import (
 
 from ....distributed.parallel_state import get_parallel_state
 from ....distributed.sequence_parallel import slice_position_embedding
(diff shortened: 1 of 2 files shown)


veomni/models/loader.py
```diff
@@ -55,6 +56,11 @@ def load_model(self, init_kwargs: dict, **kwargs):
             load_class = AutoModelForImageTextToText
         elif type(model_config) in AutoModelForVision2Seq._model_mapping.keys():  # assume built-in models
             load_class = AutoModelForVision2Seq
+        elif (
+            "ForSequenceClassification" in architecture
```
piyifan123: would this break ForCausalLM from loading? or the config would properly point to that?

anyway, is it possible to write a simple unit test for this function?

PR #272, opened 2025-12-10
[ckpt, ci] fix: refactor merge_dcp_to_hf.py to remove internal dependencies and add checkpoint verification

TimYangst:
### What does this PR do?

> Refactors the DCP-to-HuggingFace conversion script to be self-contained and adds checkpoint verification utilities to ensure conversion correctness.

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/ByteDance-Seed/VeOmni/pulls?q=is%3Apr+checkpoint+conversion
- [x] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}`: `ckpt, test`
  - `{type}`: `fix`
  - No API breaking changes

### Test

> CI tests validate the changes through existing checkpoint save/load tests (`test_trainer_saveload_ep8`, `test_trainer_saveload_ep4`, `test_trainer_saveload_no_ep`). Each test now verifies:
> - DCP checkpoint is created during training
> - HuggingFace conversion completes successfully
> - Converted checkpoint weights match original DCP weights (tensor-by-tensor comparison)

### API and Usage Example

**Command-line usage (unchanged):**
```bash
python scripts/merge_dcp_to_hf.py \
  --load-dir ./checkpoint/global_step_5 \
  --save-dir ./hf_ckpt \
  --model-assets-dir /path/to/model  # optional
```

**No breaking changes** - maintains backward compatibility.

### Design & Code Changes

**1. `scripts/merge_dcp_to_hf.py` - Self-contained script:**
- Removed dependencies on `veomni.checkpoint` and `veomni.models` modules
- Reimplemented using PyTorch DCP APIs directly (`torch.distributed.checkpoint`)
- Processes checkpoints shard-by-shard (each shard ≤2GB, configurable via `--shard-size`)
- Added model-key filtering to exclude optimizer states and training artifacts
- Improved logging and comprehensive docstrings

**2. `tests/checkpoints/checkpoint_verification_utils.py` - New verification utilities:**
- `load_dcp_checkpoint()`: Loads DCP checkpoints for verification
- `load_hf_checkpoint()`: Loads HuggingFace checkpoints (safetensors/pytorch)
- `verify_checkpoint_conversion()`: Tensor-by-tensor comparison with detailed mismatch reporting
- Supports both sharded and non-sharded checkpoints

**3. `tests/checkpoints/test_trainer_saveload.py` - Enhanced tests:**
- All tests now run `merge_dcp_to_hf.py` as subprocess after training
- Added verification step using `checkpoint_verification_utils`
- Ensures converted weights exactly match original DCP checkpoint

**Memory Efficiency:**
The implementation processes checkpoints shard-by-shard to minimize memory usage:
- Each shard (default 2GB) is loaded, converted, saved, then immediately released
- Within each shard, all tensors are loaded at once due to DCP API requirements (`torch.distributed.checkpoint.load` needs pre-allocated tensors)
- Peak memory = max(shard_size) + overhead, not total checkpoint size
- Aggressive garbage collection (`gc.collect()`) after each shard

> [!IMPORTANT]

- [x] Apply [pre-commit checks](https://github.com/ByteDance-Seed/VeOmni/blob/main/CONTRIBUTING.md?plain=1#L22): `make commit && make style && make quality`
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/ByteDance-Seed/VeOmni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: Existing end-to-end tests (`test_trainer_saveload_*`) have been updated to cover the refactored code.

### DIFF
--- a/scripts/merge_dcp_to_hf.py
+++ b/scripts/merge_dcp_to_hf.py
@@ -45,10 +45,30 @@ def get_dtype_size(dtype: torch.dtype) -> int:
 
 
 def _normalize_key(key: str) -> Optional[str]:
-    """Strip 'model.' prefix from DCP keys. Returns None for non-model weights."""
-    if key.startswith("model."):
+    """
+    Convert DCP key to HuggingFace format. Returns None for non-model weights.
+
+    Conversion rules:
+    - "model.model.*" -> "model.*" (remove first "model." prefix)
+    - "model.lm_head.weight" -> "lm_head.weight" (special case)
+    - Other "model.*" keys -> log warning and strip "model." prefix
+    """
+    if not key.startswith("model."):
+        return None
+
+    if key.startswith("model.model."):
+        # Standard case: model.model.* -> model.*
+        return key[6:]  # Remove first "model." prefix
+    elif key == "model.lm_head.weight":
+        # Special case: model.lm_head.weight -> lm_head.weight
+        return "lm_head.weight"
+    else:
+        # Other keys with single "model." prefix - log and strip prefix
+        logger.warning(
+            f"Found key with single 'model.' prefix that doesn't match expected patterns: '{key}'. "
+            f"Converting to '{key[6:]}' by stripping 'model.' prefix."
+        )
         return key[6:]
-    return None
 
 
 def _get_sharding_plan(
--- a/tests/checkpoints/checkpoint_verification_utils.py
+++ b/tests/checkpoints/checkpoint_verification_utils.py
@@ -106,13 +106,16 @@ def load_dcp_checkpoint(dcp_checkpoint_dir: str) -> Dict[str, torch.Tensor]:
         if hasattr(tensor, "full_tensor"):
             tensor = tensor.full_tensor()
 
-        # Remove the first "model." prefix for HuggingFace format
-        # DCP keys are like "model.model.layers.0.weight"
-        # HF keys should be like "model.layers.0.weight"
-        # Remove the first "model." prefix for HuggingFace format
-        # DCP keys are like "model.model.layers.0.weight"
-        # HF keys should be like "model.layers.0.weight"
-        hf_key = key[6:]  # Remove "model." prefix (6 characters)
+        # Convert DCP key to HuggingFace format:
+        # - "model.model.*" -> "model.*" (remove first "model." prefix)
+        # - "model.lm_head.weight" -> "lm_head.weight" (special case)
+        if key.startswith("model.model."):
+            hf_key = key[6:]  # Remove first "model." prefix
+        elif key == "model.lm_head.weight":
+            hf_key = "lm_head.weight"
+        else:
+            # Keep other keys as-is after removing "model." prefix
+            hf_key = key[6:]
         loaded_state_dict[hf_key] = tensor.detach().cpu()
 
         # Show progress


veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -405,10 +409,13 @@ def dcp_to_torch_state_dict(save_checkpoint_path: Union[str, os.PathLike]) -> ST
...
     state_dict: STATE_DICT_TYPE = {}
 
+    # Filter to only load model weights if requested
+    keys = ["model"] if model_only else None
+
     _load_state_dict(
         state_dict,
         storage_reader=FileSystemReader(save_checkpoint_path),
-        planner=_EmptyStateDictLoadPlanner(),
+        planner=_EmptyStateDictLoadPlanner(keys=keys),
```
piyifan123: based on the user reporting, just set keys=['models'] can work?

TimYangst: I added a flag there:  ```keys = ["model"] if model_only else None ```

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -405,10 +409,13 @@ def dcp_to_torch_state_dict(save_checkpoint_path: Union[str, os.PathLike]) -> ST
     # Load the state_dict from the DCP checkpoint
     state_dict: STATE_DICT_TYPE = {}
 
+    # Filter to only load model weights if requested
+    keys = ["model"] if model_only else None
```
piyifan123: so none means loading everything? please comment to clarify.

veomni/models/module_utils.py
```diff
@@ -119,37 +122,45 @@ class BroadcastMetadata:
     dtype: Optional["torch.dtype"]
 
 
-def _load_state_dict(weights_path: str, **kwargs) -> List["StateDictIterator"]:
+def _load_state_dict(
```
piyifan123: do not quite understand why calling this works.. from the comment and the logic, it loads a ckpt in transformers (i.e., HF) format, however, the script merge_dcp_to_hf.py loads a dcp format ckpt to state dict? 

anyway, is it possible to write a new merge_dcp_to_hf.py to make it self-contained and add a step in https://github.com/ByteDance-Seed/VeOmni/blob/main/tests/checkpoints/test_trainer_saveload.py to test?

veomni/checkpoint/dcp_checkpointer.py
TimYangst: Done

Luosuu (changes requested):
I think we can add test for this method in our CI. Maybe just in the `tests/checkpoints/test_trainer_saveload.py`?

Coach257:
rebase下master可以过ci的autolabel

FoolPlayer:
> I think we can add test for this method in our CI. Maybe just in the `tests/checkpoints/test_trainer_saveload.py`?

agree

veomni/models/module_utils.py
TimYangst: Done, rewritten the script and added the test.

TimYangst:
> I think we can add test for this method in our CI. Maybe just in the `tests/checkpoints/test_trainer_saveload.py`?

Done

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -344,43 +353,139 @@ def step_id(s):
...
+
+    # Verify the output exists
+    assert os.path.exists(hf_output_dir), f"HF checkpoint directory not found: {hf_output_dir}"
+    files = os.listdir(hf_output_dir)
+    logger.info(f"HF checkpoint files: {files}")
+    assert len(files) > 0, "No files generated in HF checkpoint directory"
+
+    # Verify HF checkpoint structure
+    logger.info("Verifying HF checkpoint structure...")
+    assert verify_hf_checkpoint(hf_output_dir, verify_structure=True, verify_weights=False), (
```
piyifan123: why is verify_weight is false

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -344,43 +353,139 @@ def step_id(s):
...
```
piyifan123: same here.

scripts/merge_dcp_to_hf.py
```diff
@@ -1,40 +1,349 @@
...
+
+    # Only load model weights (excluding optimizer states, etc.)
+    # keys=["model"] will only load model state dict
+    # See _EmptyStateDictLoadPlanner._should_include_key() in torch's default_planner.py
+    keys = ["model"]
+
+    _load_state_dict(
+        state_dict,
+        storage_reader=FileSystemReader(save_checkpoint_path),
+        planner=_EmptyStateDictLoadPlanner(keys=keys),
```
piyifan123: so it still loads the entire ckpt to the cpu mem? just asking, it's OK to have that at the beginning.

scripts/merge_dcp_to_hf.py
```diff
@@ -1,40 +1,349 @@
...
+    This function only loads model weights (ignoring optimizer states, etc.) to reduce memory usage.
+
+    Args:
+        save_checkpoint_path: Directory containing the DCP checkpoint.
+
+    Returns:
+        state_dict: Dictionary containing model weights.
+
+    Warning:
+        To avoid OOM, it's recommended to only run this function on a single rank.
```
piyifan123: what does this mean? we should only run this on a CPU machine. And even if GPU is available, we should avoid use GPU mem to load dcp as it's only 80GB most times which is insufficient for bigger models.

scripts/merge_dcp_to_hf.py
```diff
@@ -1,40 +1,349 @@
...
+
+    # Load the state_dict from the DCP checkpoint
+    state_dict: STATE_DICT_TYPE = {}
+
+    # Only load model weights (excluding optimizer states, etc.)
+    # keys=["model"] will only load model state dict
+    # See _EmptyStateDictLoadPlanner._should_include_key() in torch's default_planner.py
+    keys = ["model"]
+
+    _load_state_dict(
```
piyifan123: this is the public API from torch, wondering if we can use this https://github.com/pytorch/pytorch/blob/orig/release/2.8/torch/distributed/checkpoint/state_dict_loader.py#L53-L61

TimYangst: done

tests/checkpoints/test_trainer_saveload.py
TimYangst: Updated.

tests/checkpoints/test_trainer_saveload.py
TimYangst: Updated

scripts/merge_dcp_to_hf.py
TimYangst: Tested on cpu only machine, it should work

scripts/merge_dcp_to_hf.py
TimYangst: Yes, you're right. The current implementation does load the entire checkpoint into CPU memory at once. This is required by the `torch.distributed.checkpoint.load` API design, which needs pre-allocated tensors with correct shapes to perform in-place loading.

For future optimization, we could explore:
- Incremental loading: Process tensors shard-by-shard instead of loading everything upfront
- Streaming conversion: Load → convert → save in a pipeline to avoid holding two copies in memory
- Memory-mapped reading: Reduce memory footprint for very large checkpoints

But as an initial implementation, this approach prioritizes correctness and simplicity.

scripts/merge_dcp_to_hf.py
```diff
@@ -1,40 +1,323 @@
...
+    all_dcp_keys = []
+
+    for key, tensor_meta in metadata.state_dict_metadata.items():
+        hf_key = _normalize_key(key)
+        if hf_key:
+            # Determine dtype for size calculation
+            if save_dtype:
+                dtype = getattr(torch, save_dtype) if isinstance(save_dtype, str) else save_dtype
+            else:
+                dtype = tensor_meta.properties.dtype if hasattr(tensor_meta.properties, "dtype") else torch.float32
```
Luosuu: is it safe to assume dtype to be fp32 when we cannot read its dtype?

scripts/merge_dcp_to_hf.py
```diff
@@ -1,40 +1,323 @@
...
+        torch.float16: 2,
+        torch.bfloat16: 2,
+        torch.int64: 8,
+        torch.int32: 4,
+        torch.int16: 2,
+        torch.int8: 1,
+        torch.uint8: 1,
+        torch.bool: 1,
+    }
+    return size_map.get(dtype, 4)
```
Luosuu, line 44: same here, is default value of 32 safe?

scripts/merge_dcp_to_hf.py
Luosuu: let's assert they must have dtensor metadata

tests/checkpoints/checkpoint_verification_utils.py
```diff
@@ -0,0 +1,428 @@
...
+        if hasattr(tensor, "full_tensor"):
+            tensor = tensor.full_tensor()
+
+        # Remove the first "model." prefix for HuggingFace format
+        # DCP keys are like "model.model.layers.0.weight"
+        # HF keys should be like "model.layers.0.weight"
+        # Remove the first "model." prefix for HuggingFace format
+        # DCP keys are like "model.model.layers.0.weight"
+        # HF keys should be like "model.layers.0.weight"
+        hf_key = key[6:]  # Remove "model." prefix (6 characters)
```
piyifan123: if key.startswith(model.model) then truncate.

FoolPlayer (approved):

FoolPlayer:
LGTM! Can you help add the usage in docs in next PR, thanks!

scripts/merge_dcp_to_hf.py
TimYangst: Done, raised an exception there. And discussed with yifan last night, we will convert the weight type as bf16 as default.

scripts/merge_dcp_to_hf.py
TimYangst: As we raised an exception below, we won't go to default value there.

tests/checkpoints/checkpoint_verification_utils.py
TimYangst: Done, I add log in the merge_dcp_to_hf script, in case, there other key that only have single model. prefix.

TimYangst:
> LGTM! Can you help add the usage in docs in next PR, thanks!

Sure, will have a follow up MR for the usage doc.

Luosuu (approved):

PR #297, opened 2025-12-16
[data, model] feat: support Qwen3-VL textual token-based time encoding

GarrickBot:
### Overview
This PR integrates the specific data processing and modeling logic required for **Qwen3-VL**, specifically focusing on its unique **textual token-based time encoding strategy**.

Unlike previous model versions, Qwen3-VL requires explicit time tags (e.g., `<t 1.5 seconds>`) inserted into the context before visual tokens. This PR implements the official timestamp calculation logic, adds a dedicated data processor for Qwen3-VL video samples, and optimizes the underlying video loading utilities for better memory efficiency during training.

### Design & Code Changes

The changes are primarily concentrated in the data processing pipeline and chat template adapter:

* **`multimodal_chat_template.py`**:
    * **New Class**: Added `Qwen3VLChatTemplate` inheriting from `Qwen2VLTemplate`.
    * **Timestamp Logic**: Implemented `_calculate_timestamps` to replicate the official strategy: padding frame indices, converting to seconds, and averaging based on `temporal_patch_size=2`.
    * **Message Encoding**: Updated `encode_messages` to dynamically inject `<t {timestamp} seconds>` tags and `<|vision_start|>/<|vision_end|>` tokens into the input stream.

* **`vlm_data_process.py`**:
    * **New Processor**: Added `process_sample_qwen3_vl` to handle the specific requirements of Qwen3-VL video inputs.
    * **Data Adapter**: Added logic to automatically construct the `conversations` field from `question` and `answer` keys if the standard conversation format is missing.
    * **Metadata Fix**: Updated video metadata handling to correctly generate and assign `frames_indices`.

* **`video_utils.py`**:
    * **Workflow Optimization**: Refactored load_video_from_bytes_list to directly support pre-stored frame byte lists as input. This bypasses the need for online frame extraction (decoding), significantly reducing memory overhead.
    * **Metadata Adaptation**: Updated the function to correctly handle input VideoMetadata and ensure the output includes the modified VideoMetadata (reflecting changes in FPS or frame counts) for downstream processing.

### Data Format

The new data processor (`process_sample_qwen3_vl_video_r1`) expects input samples to follow this schema:

```python
sample = {
    # 1. 'conversations' is required (or auto-constructed from question/answer)
    "conversations": [
        {
            "role": "user", 
            "content": [
                {"type": "video"}, 
                {"type": "text", "text": "Describe this video."}
            ]
        },
        {"role": "assistant", "content": "This is a video of..."}
    ],
    
    # 2. 'fps' must be a list containing the float FPS value
    "fps": [2.0],
    
    # 3. 'videos' must be a list of raw byte frames
    "videos": [
        [b'\xff\xd8...', b'\xff\xd8...', ...] # List[bytes]
    ]
}

### DIFF
--- a/tasks/data/vlm_data_process.py
+++ b/tasks/data/vlm_data_process.py
@@ -166,7 +166,7 @@ def process_sample_qwen3_vl(
         token_num_inputs["image"] = image_token_num
         tokenized_example = chat_template.encode_messages(conversations, token_num_inputs)
     if "videos" in sample:
-        videos, metadata = fetch_videos_metadata(sample["videos"], fps=sample["fps"], **kwargs)
+        videos, metadata, _, _ = fetch_videos_metadata(sample["videos"], fps=sample["fps"], **kwargs)
         # Process videos without resizing or sampling frames initially
         video_inputs = processor.video_processor(videos=videos, video_metadata=metadata, return_tensors="pt", return_metadata=True)
         video_grid_thw = video_inputs["video_grid_thw"]
--- a/veomni/data/multimodal/video_utils.py
+++ b/veomni/data/multimodal/video_utils.py
@@ -1,18 +1,17 @@
 import math
 from io import BytesIO
-from typing import ByteString, Dict, List, Union
+from typing import ByteString, Dict, List, Optional, Tuple, Union
 
 import av
 import librosa
 import numpy as np
-import PIL
+import PIL.Image
 import torch
 import torchvision
 from torchvision.io.video import _read_from_stream
-from torchvision.transforms import InterpolationMode, functional
+from torchvision.transforms import InterpolationMode, functional as F
 
 from ...utils import logging
-
 
 logger = logging.get_logger(__name__)
 
@@ -41,50 +40,6 @@ def save_video_bytes_to_file(video_bytes, output_path):
         f.write(video_bytes)
 
 
-def smart_video_nframes(
-    video: torch.Tensor,
-    video_fps: Union[int, float],
-    fps: int = 2.0,
-    frame_factor: int = None,
-    min_frames: int = None,
-    max_frames: int = None,
-    **kwargs,
-) -> torch.Tensor:
-    total_frames = video.shape[0]
-    nframes = total_frames / video_fps * fps
-
-    if min_frames is not None:
-        if frame_factor is not None:
-            min_frames = math.ceil(min_frames / frame_factor) * frame_factor
-        nframes = max(min_frames, nframes)
-
-    if max_frames is not None:
-        if frame_factor is not None:
-            max_frames = math.floor(max_frames / frame_factor) * frame_factor
-        nframes = min(max_frames, nframes)
-
-    nframes = min(nframes, total_frames)
-    if frame_factor is not None:
-        nframes = math.floor(nframes / frame_factor) * frame_factor
-        nframes = max(nframes, frame_factor)
-
-    if nframes > total_frames:
-        pad_count = nframes - total_frames
-        last_frame = video[-1:].expand(pad_count, -1, -1, -1)  # shape: (pad_count, C, H, W)
-        video = torch.cat([video, last_frame], dim=0)
-        total_frames = video.shape[0]
-
-    idx = torch.linspace(0, total_frames - 1, int(nframes)).round().long()
-    video = video[idx]
-    return video
-
-
-def smart_audio_nframes(audio: np.ndarray, audio_fps: int, sample_rate: int = 16000, **kwargs):
-    if audio is not None:
-        audio = librosa.resample(audio, orig_sr=audio_fps, target_sr=sample_rate)
-    return audio
-
-
 def smart_resize(
     video: torch.Tensor,
     scale_factor: int = None,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: veomni/data/multimodal/video_utils.py [hunk 1]
```diff
@@ -216,3 +269,35 @@ def fetch_videos(videos: List[VideoInput], **kwargs):
         smart_audio_nframes(audio, audio_fps, **kwargs) for audio, audio_fps in zip(audio_inputs, audio_fps_list)
     ]
     return video_inputs, audio_inputs
+
+def fetch_videos_metadata(videos: List[VideoInput], **kwargs):
```
Coach257: Can we merge this function into `fetch_videos`? It seems return metadata will be necessary, so `fetch_videos` returns videos, video_metadata, audios, audio_metadata

veomni/data/multimodal/video_utils.py
```diff
@@ -216,3 +269,35 @@ def fetch_videos(videos: List[VideoInput], **kwargs):
...
+
+    for v_raw, v_fps in zip(video_inputs_raw, video_fps_list):
+        processed_v, processed_fps = smart_video_nframes(
+            smart_resize(v_raw, **kwargs), v_fps, **kwargs
+        )
+        final_video_inputs.append(processed_v)
+        final_fps_inputs.append(processed_fps)
+
+    
+    return final_video_inputs, final_fps_inputs
```
Coach257: As the returns[1] is metadata, the second return should be a dict.

tasks/data/vlm_data_process.py
```diff
@@ -135,6 +135,79 @@ def process_sample_qwen2_5_vl(
 
     return [tokenized_example]
 
+def process_sample_qwen3_vl(
```
Coach257, line 138: @piyifan123 Please check if there's any conflict with the origin `process_sample_qwen3_vl`. If no, use this new function?

veomni/data/multimodal/multimodal_chat_template.py
```diff
@@ -420,6 +420,174 @@ def encode_messages(
...
+        return system_message
+
+    # ================= [New: Official Timestamp Calculation Logic] =================
+    def _calculate_timestamps(self, indices: List[int], video_fps: float, merge_size: int = 2):
+        """
+        Replicates Qwen3-VL official logic: Pad -> Convert to Seconds -> Average.
+        """
+        # 1. Pad frame indices to be divisible by merge_size
+        if len(indices) % merge_size != 0:
+            indices.extend([indices[-1]] * (merge_size - len(indices) % merge_size))
```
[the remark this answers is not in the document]
GarrickBot: Although `indices` is passed as an argument, I've verified that the caller does not reference or reuse it after passing it to this function. Therefore, the in-place modification (extend) is safe in this context and avoids unnecessary copying, while aligning with the official Qwen3-VL implementation.

tasks/data/vlm_data_process.py
```diff
@@ -135,6 +135,79 @@ def process_sample_qwen2_5_vl(
...
+    tokenized_example.update(video_inputs)
+    
+    if record_process_time:
+        process_time = time.time() - start_time
+        tokenized_example["process_sample_time_sec"] = process_time
+
+    return [tokenized_example]
+
 
 def process_sample_qwen3_vl(
```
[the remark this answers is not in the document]
GarrickBot: Fixed. The duplicate function definition has been removed.

veomni/data/multimodal/multimodal_chat_template.py
```diff
@@ -420,6 +420,174 @@ def encode_messages(
...
+                if value[0] == "text":
+                    content += value[1]
+                elif value[0] == "image":
+                    data_type = "t2i" if role == "assistant" else "i2t"
+                    # Assumes self.image_pattern returns Qwen2-VL style image padding
+                    content += self.image_pattern(next(image_token_num_list))
+                
+                elif value[0] == "video":
+                    # --- [Core Modification: Video Timestamp Processing] ---
+                    total_video_tokens = next(video_token_num_list)
```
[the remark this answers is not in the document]
GarrickBot: Fixed. Added a try-except block to safely handle `StopIteration` as suggested.

veomni/data/multimodal/video_utils.py
GarrickBot: I have updated the logic to support returning the full set of 4 values (video, video_metadata, audio, audio_metadata). However, I decided to keep `fetch_videos_metadata` separate from `fetch_videos`.
This design ensures isolation: `fetch_videos_metadata` is for the new pipeline requiring full metadata, while `fetch_videos` falls back to the original behavior (returning 2 values) to ensure backward compatibility for other tasks.

veomni/data/multimodal/video_utils.py
GarrickBot: The return value was actually already a `List[Dict]`, but the previous variable name `final_fps_inputs` was indeed misleading (it sounded like a list of floats).
I have renamed the variables in the latest commit to explicitly reflect that they are lists of metadata dictionaries, which should clear up the confusion.

Coach257:
#386

PR #304, opened 2025-12-17
[ops] feat: veomni_fa_function

Coach257:
### What does this PR do?

1. patch `flash_attention_2` and `flash_attention_3` in `ALL_ATTENTION_FUNCTIONS`
2. define new mask_func in `ALL_MASK_ATTENTION_FUNCTIONS`
3. add external fa ops in `ALL_FLASH_ATTENTION_FUNCTIONS`, using `transformers.modeling_flash_attention_utils._flash_attention_forward` for default
    - delete seed_kernels related, for internal use only

4. `flash_attention_forward` up to date (transformers==4.57.3)
5. models support fa3 now: qwen2, qwen3, qwen3moe

### DIFF
--- a/tests/models/utils.py
+++ b/tests/models/utils.py
@@ -107,7 +107,7 @@ def prepare_models_modes(is_moe: bool = False):
                 ),
                 ModelMode(
                     force_use_huggingface=False,
-                    attn_implementation="veomni_flash_attention_3",
+                    attn_implementation="flash_attention_3",
                     attn_case="position_ids",
                     moe_implementation="fused",
                 ),


Luosuu:
Hi, let's coordinate on this since I am support fa3 now #305

veomni/models/transformers/deepseek_v3/modeling_deepseek.py
```diff
@@ -53,6 +53,7 @@
 
 from ....ops import causallm_loss_function, fused_moe_forward
 from ....utils.import_utils import is_liger_kernel_available
+from ...modeling_utils import ALL_ATTENTION_FUNCTIONS
```
piyifan123: wondering if we can keep the current impl that just register the transformers ALL_ATTENTION_FUNCTIONS? which requires less modifications to the modeling code.

veomni/models/modeling_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+logger = logging.get_logger(__name__)
+
+# Global AttentionInterface shared by all models which do not need to overwrite any of the existing ones
+ALL_ATTENTION_FUNCTIONS: AttentionInterface = AttentionInterface()
+
+ALL_ATTENTION_FUNCTIONS["flash_attention_3"] = flash_attention_forward
+ALL_ATTENTION_FUNCTIONS["flash_attention_2"] = flash_attention_forward
+
+ALL_FLASH_ATTENTION_FUNCTIONS["flash_attention_2"] = _flash_attention_forward
```
piyifan123: nit: a nitpick is that it's kind of confusing for users since they still confuses this "flash_attention_2" with the HF "flash_attention_2" impl. One idea is to give it another name like "veomni_ulysses_fa2".

veomni/models/modeling_utils.py
```diff
@@ -0,0 +1,19 @@
...
+from ..ops.attention import ALL_FLASH_ATTENTION_FUNCTIONS, flash_attention_forward
+from ..utils import logging
+
+
+logger = logging.get_logger(__name__)
+
+# Global AttentionInterface shared by all models which do not need to overwrite any of the existing ones
+ALL_ATTENTION_FUNCTIONS: AttentionInterface = AttentionInterface()
+
+ALL_ATTENTION_FUNCTIONS["flash_attention_3"] = flash_attention_forward
```
piyifan123: i feel it's a little bit confusing for users that fa2 and fa3 shares the same interface and calling arguments. IIUC, this just uses the transformers logic that when impl=None, it just decides based on which package is installed which I think it's not very clear to users and developers. 

Can we instead bind impl=fa2 and impl=fa3 to the functions?

veomni/ops/attention.py
```diff
@@ -28,28 +30,56 @@ def flash_attention_forward(
...
 
-    # FA2 always relies on the value set in the module, so remove it if present in kwargs to avoid passing it twice
-    kwargs.pop("is_causal", None)
+    # In PEFT, usually we cast the layer norms in float32 for training stability reasons
+    # therefore the input hidden states gets silently casted in float32. Hence, we need
+    # cast them back in the correct dtype just to be sure everything works as expected.
+    # This might slowdown training & inference so it is recommended to not cast the LayerNorms
+    # in fp32. (usually our RMSNorm modules handle it correctly)
+    target_dtype = None
+    if query.dtype == torch.float32:
```
piyifan123, line 64: wondering if we can just have a PEFT specific implementation variant like "veomni_peft_fa2" so it's clear for developers and users that this logic is only triggered for peft cases.

veomni/models/modeling_utils.py
Coach257: fa2 and fa3 shares the same Ulysses patched `flash_attention_forward` but different `_flash_attention_forward`. The difference is controlled by the `attn_implementation`. See https://github.com/huggingface/transformers/blob/bdc85cb85c8772d37aa29ce447860b44d7fad6ef/src/transformers/modeling_flash_attention_utils.py#L608

veomni/ops/attention.py
Coach257: It's just a copy of https://github.com/huggingface/transformers/blob/bdc85cb85c8772d37aa29ce447860b44d7fad6ef/src/transformers/integrations/flash_attention.py#L68
I will update this to 5.0.0

veomni/models/modeling_utils.py
Coach257: But to use `veomni_ulysses_fa2`， the user should load model by `AutoModel.from_pretrained(attn_implementation="veomni_ulysses_fa2")`. I think it's strange.

Coach257: When `sp_size>1`, use `veomni_ulysses_fa2`; when `sp_size=1 and seed_kernels_availabale`, also use `veomni_ulysses_fa2`. So I think when using a veomni patched model, all the functions should be replaced with veomni patched fa forward.

veomni/models/transformers/deepseek_v3/modeling_deepseek.py
Coach257: See https://github.com/ByteDance-Seed/VeOmni/pull/304#discussion_r2629406111

Coach257:
> Hi, let's coordinate on this since I am support fa3 now #305

Qwen3Moe; Qwen3; Qwen2 fa3 supported and ci added @Luosuu

veomni/models/modeling_utils.py
KKZ20: I think it is the same implemention for ulysses?

KKZ20: you can let transformers to decide attn_implementation by analyzing your env

veomni/models/modeling_utils.py
Coach257: renameed as veomni_flash_attention_2

veomni/models/transformers/deepseek_v3/modeling_deepseek.py
Coach257: register to hf

veomni/models/transformers/llama/modeling_llama.py
```diff
@@ -640,7 +640,7 @@ def _update_causal_mask(
         past_key_values: Cache,
         output_attentions: bool,
     ):
-        if self.config._attn_implementation == "flash_attention_2":
+        if "flash_attention_2" in self.config._attn_implementation:
```
Luosuu: do we need to handle fa3 here as well

veomni/models/transformers/llama/modeling_llama.py
Luosuu, line 635: I see we deleted `_update_causal_mask` for Qwen3 model to match the latest huggingface transformers. Do we still keep it for deepseek models?

veomni/utils/arguments.py
```diff
@@ -82,9 +82,17 @@ class ModelArguments:
...
+            "sdpa",
+            "flash_attention_2",
+            "flash_attention_3",
+            "veomni_flash_attention_2",
+            "veomni_flash_attention_3",
+            "native-sparse",
+        ]
     ] = field(
-        default="flash_attention_2",
+        default="veomni_flash_attention_2",
```
Luosuu: maybe we need to explain what is the diff between `flash_attention_2` and `veomni_flash_attention_2`. Also, does this mean all the job yamls need to be updated if they are using SP?

pyproject.toml
```diff
@@ -67,6 +67,7 @@ gpu = [
   "nvidia-nccl-cu12",
   "liger-kernel",
   "flash-attn",
+  "flash-attn-3",
```
Luosuu, line 70: maybe make flash-attn-3 as a single option so that we install fa3 like `uv sync --extra gpu --extra fa3`. I am not sure whether it is fine to install fa3 for non-Hopper GPUs.

veomni/models/transformers/llama/modeling_llama.py
Coach257: Prioritize updating the Qwen series models first. All subsequent models will be updated to transformers 5.0.0 by applying patches to the corresponding `modeling.py`.

veomni/utils/arguments.py
Coach257: The default value for argument `attn_implementation` is now `veomni_flash_attention_2`. If users want to run **native transformers modeling**, they can set `flash_attention_2`.

pyproject.toml
Coach257: It's runtime error. Set `veomni_flash_attention_2` instead of `fa3`.

veomni/models/transformers/llama/modeling_llama.py
Coach257: Yes

Luosuu:
now it looks good to me. But let @piyifan123 also take a look

veomni/models/transformers/qwen3_moe/modeling_qwen3_moe.py
```diff
@@ -758,8 +758,14 @@ def forward(
         if position_ids is None:
             position_ids = cache_position.unsqueeze(0)
 
-        causal_mask = self._update_causal_mask(
-            attention_mask, inputs_embeds, cache_position, past_key_values, output_attentions
+        mask_function = create_causal_mask if self.config.sliding_window is None else create_sliding_window_causal_mask
```
piyifan123, line 761: is this following the tf 5.0 convention?

piyifan123 (approved):

FoolPlayer (approved):

veomni/models/transformers/qwen3_moe/modeling_qwen3_moe.py
Coach257: Yes

PR #167, opened 2025-11-03
[ops] feat: Optimize Qwen3-Moe Performance on Ascend NPU with Patches

zhihaofang1017:
### What does this PR do?

This PR introduces targeted performance optimizations for the Qwen3-Moe model for Ascend NPU environments. The changes leverage NPU-native fused operators, custom kernel integrations, and NPU patches to address key performance bottlenecks in RMSNorm, Rotary Position Embedding (RoPE) — critical components of Qwen3-Moe’s training pipeline.
Also, we improve the performance foe eager moe_implementation.

## Key Changes
The optimization spans 3 core files,  focusing on NPU compatibility, operator fusion, and performance-critical code paths:

| File Path | Modification Details | Purpose |
|-----------|----------------------|---------|
| `veomni/__init__.py` | 1. Added `is_torch_npu_available()` utility to check NPU runtime availability<br>2. Auto-imports NPU patches when NPU is detected | Ensure runtime environment meets NPU requirements; enable auto-patching for NPU |
| `veomni/ops/npu_patch.py` | Implemented NPU-native rms_norm_forward() with torch_npu.npu_rms_norm, replaced RoPE with apply_rotary_pos_emb_qwen3_npu() , and added custom GmmFunction for MoE expert computations. | Replace PyTorch native ops with NPU-optimized versions; reduce MoE computation latency |
| `veomni/ops/npu_patch.py` | It uses NPU MoE primitives and grouped GEMM to reduce overhead, fuse activation, and improve performance while preserving original routing behavior and numerical equivalence. | Optimize Qwen3-Moe Performance for **eager** moe_implementation |

## Performance Improvements
We performed performance verification on the qwen3-30B model. Under the condition that the accuracy meets the standard
- **Hardware**: 16P Ascend NPU (tested on Ascend 910B, A+X).
- **Software Dependencies**:
- PyTorch&torch_npu =2.7.1
- Transformers =4.57.1 
- **Main parameters**: 
- Seqlen = 8k
- gbs/mbs = 32/1
- FSDP1
- EP/SP = 1/1
- dataset：fineweb-edu/sample/10BT
1. **RMSNorm**: End-to-end performance improvement is 7%.
2. **RoPE**: End-to-end performance improvement is 5%.
3. **Eager moe_implementation**: Huge performance improvements

## Testing
All changes have been validated with:
1. **Functional Tests**: Qwen3-Moe passes inference/training correctness checks on Ascend NPU.
[...]

PR #314, opened 2025-12-20
[ops] fix: mv qwen3moe patch

Coach257:
### What does this PR do?

Runtime patch qwen3moe to aviod import error. mv npu patch to qwen3moe modeling.


Coach257:
#167  @zhihaofang1017

PR #317, opened 2025-12-22
[model] feat: add Qwen3ForSequenceClassification to support sequence embedding classification

yiwzhao:
[model] feat: add Qwen3ForSequenceClassification to support sequence embedding classification

### DIFF
--- a/veomni/models/transformers/qwen3/modeling_qwen3.py
+++ b/veomni/models/transformers/qwen3/modeling_qwen3.py
@@ -741,8 +741,7 @@ class Qwen3ForSequenceClassification(Qwen3PreTrainedModel):
     def __init__(self, config):
         super().__init__(config)
         self.num_labels = config.num_labels
-        # we're not using AutoModel here, but our own Qwen3Model.
-        setattr(self, self.base_model_prefix, Qwen3Model(config))
+        self.model = Qwen3Model(config)
         self.score = nn.Linear(config.hidden_size, self.num_labels, bias=False)
         self.loss_fct = CrossEntropyLoss(ignore_index=-100, reduction="none")
         # Initialize weights and apply final processing


veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,95 @@ def forward(
...
+    - The current loss function directly uses single-label CrossEntropyLoss.
+    """
+
+    base_model_prefix = "model"
+
+    def __init__(self, config):
+        super().__init__(config)
+        self.num_labels = config.num_labels
+        # we're not using AutoModel here, but our own Qwen3Model.
+        setattr(self, self.base_model_prefix, Qwen3Model(config))
```
piyifan123: I don't quite understand this? can it just be?        self.model = Qwen3Model(config)

like https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3/modeling_qwen3.py#L464C1-L465C1

yiwzhao: Yes, in this class they are equal.

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,94 @@ def forward(
         )
 
 
+class Qwen3ForSequenceClassification(Qwen3PreTrainedModel):
```
piyifan123, line 730: nit: add modification note.

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,94 @@ def forward(
...
+        input_ids: Optional[torch.LongTensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Cache] = None,
+        inputs_embeds: Optional[torch.FloatTensor] = None,
+        labels: Optional[torch.LongTensor] = None,
+        use_cache: Optional[bool] = None,
+        cu_seqlens: Optional[torch.Tensor] = None,
+        max_seqlen: Optional[int] = None,
+        **kwargs: Unpack[FlashAttentionKwargs],
```
piyifan123, line 763: check if it conflicts with the unpack FlashAttentionKwargs

veomni/ops/loss.py
```diff
@@ -90,3 +91,23 @@ def causallm_loss_function(
...
+) -> torch.Tensor:
+    # local sum loss
+    # CrossEntropyLoss(reduction="none") + mask + sum
+    per = loss_fct(logits, labels)  # [N] if reduction="none"
+    valid = labels != ignore_index
+    loss_sum = (per * valid).sum()
+    cnt = valid.sum().to(dtype=loss_sum.dtype)
+
+    if sp_group is not None:
+        dist.all_reduce(loss_sum, op=dist.ReduceOp.SUM, group=sp_group)
```
piyifan123, line 111: see if we can reuse reduce_sequence_parallel_loss

veomni/ops/loss.py
```diff
@@ -90,3 +91,23 @@ def causallm_loss_function(
         loss = reduce_sequence_parallel_loss(loss, num_valid_tokens)
 
     return loss, logits
+
+
+def seqcls_token_loss_sp_aware(
```
piyifan123, line 96: add docstring and unit tests.

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,94 @@ def forward(
...
+            **kwargs,
+        )
+
+        hidden_states = transformer_outputs.last_hidden_state
+        logits = self.score(hidden_states)
+
+        loss = None
+        if labels is not None:
+            # labels are token-level now, shape must match logits tokens
+            if logits.dim() == 3:
```
piyifan123, line 784: can we assume always we have B?

veomni/ops/loss.py
Coach257, line 96: why not use the `causallm_loss_function`

piyifan123: Shift label is not needed for seq classification task in general. And it's more readable and less error prone to have a separate loss function for classification from ntp loss imo

veomni/models/transformers/qwen3/modeling_qwen3.py
yiwzhao: There won't be conflict because the key names are different. FlashAttentionKwargs only contains: cu_seq_lens_q, cu_seq_lens_k, max_length_q, max_length_k. My key names here are use_cache, cu_seqlens, and max_seqlen.

veomni/models/transformers/qwen3/modeling_qwen3.py
yiwzhao: Yes, we can. I‘ll remove the branch where logits.dim() == 2.

veomni/models/transformers/qwen3/modeling_qwen3.py
yiwzhao: Added

veomni/ops/loss.py
yiwzhao: Added.

veomni/ops/loss.py
yiwzhao: Yes, we can. I implemented it in the new MR: https://github.com/ByteDance-Seed/VeOmni/pull/322

veomni/models/transformers/qwen3/modeling_qwen3.py
yiwzhao: Yes, in this scenario they are the same.

veomni/models/transformers/qwen3/modeling_qwen3.py
piyifan123: why not use FlashAttentionKwargs instead? i don't see the reason why you need to define your own cu_seqlens, and max_seqlen which could cause HF transformers fa_util does not read the proper FA kwargs and cause cpu sync calculation.

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,94 @@ def forward(
...
+        position_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Cache] = None,
+        inputs_embeds: Optional[torch.FloatTensor] = None,
+        labels: Optional[torch.LongTensor] = None,
+        use_cache: Optional[bool] = None,
+        cu_seqlens: Optional[torch.Tensor] = None,
+        max_seqlen: Optional[int] = None,
+        **kwargs: Unpack[FlashAttentionKwargs],
+    ) -> SequenceClassifierOutputWithPast:
+        transformer: Qwen3Model = getattr(self, self.base_model_prefix)
```
piyifan123, line 765: why can't we just simply write self.model here?

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,98 @@ def forward(
...
+            input_ids,
+            attention_mask=attention_mask,
+            position_ids=position_ids,
+            past_key_values=past_key_values,
+            inputs_embeds=inputs_embeds,
+            use_cache=use_cache,
+            cu_seqlens=cu_seqlens,
+            max_seqlen=max_seqlen,
+            **kwargs,
+        )
```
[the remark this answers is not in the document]
piyifan123: yes, same issue here. why define your own cu_seqlens and max_seqlen while the underlying model does not even understand it...

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -725,10 +727,94 @@ def forward(
...
+
+        hidden_states = transformer_outputs.last_hidden_state
+        logits = self.score(hidden_states)
+
+        loss = None
+        if labels is not None:
```
piyifan123, line 782: i still think it's too complex while we can basically go with the linearcrossentropy path as the NTP head does by making sure that labels are set with proper ignore_index there.

veomni/ops/loss.py
piyifan123, line 96: i think you should have the exact same input format as causallm_loss_function, i.e., the part I don't think makes a lot of sense is: the current definition requires model to calculate logits which will cause 2 problems:
1. hidden_states -> logits projection logic is leaked to the modeling code leads to more complexity. 
2. lose the chance to use a fusedlinearcrossentropy kernel for performance.
3. introduce a different pattern from ntp loss while they can be pretty much very similar for people to understand.

PR #339, opened 2025-12-29
[dist] fix: loss when sp enabled

Coach257:
### What does this PR do?

1. fix loss when sp>1
- loss in sp group must be the same: `loss = all_reduce(loss, group = sp_group) / sp_size`
2. fix token count when use rmpad_with_position_id
- padding tokens should not be count in meter
3. register `ForConditionalGeneration`
4. fix some bugs in qwen3vlmoe

### DIFF
--- a/veomni/utils/loss_utils.py
+++ b/veomni/utils/loss_utils.py
@@ -43,11 +43,10 @@ def mean_global_loss(
         if get_parallel_state().sp_enabled:
             cur_token_len = all_reduce(cur_token_len.item(), op="sum", group=get_parallel_state().sp_group)
 
-        cur_loss = cur_loss * cur_token_len
         all_reduced_len = all_reduce((micro_batches_token_len[f"{loss_name}_tokens"].item()), op="sum")
 
         if all_reduced_len != 0:
-            cur_loss = cur_loss / all_reduced_len * get_parallel_state().fsdp_size
+            cur_loss = cur_loss * cur_token_len / all_reduced_len * get_parallel_state().fsdp_size
         else:  # no loss
             assert torch.allclose(cur_loss, torch.zeros_like(cur_loss)), f"cur_loss: {cur_loss}"
 


veomni/utils/loss_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        losses = {"foundation_loss": losses}
+
+    for key, cur_loss in losses.items():
+        loss_name = key.split("_loss")[0]  # foundation/image_decoder/**
+
+        cur_token_len = micro_batch_token_len[f"{loss_name}_tokens"]
+        if get_parallel_state().sp_enabled:
+            cur_token_len = all_reduce(cur_token_len.item(), op="sum", group=get_parallel_state().sp_group)
+
+        cur_loss = cur_loss * cur_token_len
```
FoolPlayer: Maybe it's better to get `cur_loss` per rank with line 50

Coach257: `cur_loss` here is loss per dp_rank

TimYangst (approved):
Looks good to me, thanks!

veomni/utils/loss_utils.py
```diff
@@ -0,0 +1,61 @@
...
+import torch
+
+from ..data.constants import IGNORE_INDEX
+from ..distributed.parallel_state import get_parallel_state
+from .device import get_device_type
+from .dist_utils import all_reduce
+
+
+def count_loss_token(batches: Union[list[dict[str, torch.Tensor]], dict[str, torch.Tensor]]):
+    """Calculate the total number of text_tokens/image_tokens/** for loss in a global batch, or one micro batch."""
```
piyifan123, line 13: nit: maybe add a note that when calling with pattern with micro_batches, it would require dataloader to finalize all data in a global batch for a rank, which can take more time?

veomni/utils/loss_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    if isinstance(batches, dict):
+        batches = [batches]
+    token_len = defaultdict(int)
```
piyifan123, line 16: nit: wondering if we can use a "mini" dataclass here for better readability about these 2 fields?

veomni/utils/loss_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    return token_len
+
+
+def mean_global_loss(
+    losses: Union[dict[str, torch.Tensor], torch.Tensor],
+    micro_batch_token_len: dict[str, torch.Tensor],
+    micro_batches_token_len: dict[str, torch.Tensor],
+):
+    """Calcuate the global mean loss. Avg on all_reduced_token_num instead of on dp_size.
+    - cur_losses[key] = cur_loss * cur_token_num / global_batches_token_num * get_parallel_state().fsdp_size
```
piyifan123, line 30: nit: maybe explain the last * get_parallel_state().fsdp_size which might not be clear to some users who do not know fsdp by default divides gradients by its size.

veomni/utils/loss_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        if get_parallel_state().sp_enabled:
+            cur_token_len = all_reduce(cur_token_len.item(), op="sum", group=get_parallel_state().sp_group)
+
+        cur_loss = cur_loss * cur_token_len
+        all_reduced_len = all_reduce((micro_batches_token_len[f"{loss_name}_tokens"].item()), op="sum")
+
+        if all_reduced_len != 0:
+            cur_loss = cur_loss / all_reduced_len * get_parallel_state().fsdp_size
+        else:  # no loss
+            assert torch.allclose(cur_loss, torch.zeros_like(cur_loss)), f"cur_loss: {cur_loss}"
```
piyifan123, line 52: nit: https://google.github.io/styleguide/pyguide.html#:~:text=Do%20not%20use%20assert,verify%20expectations.%20For%20example%3A

Coach257: fix these in next pr

PR #352, opened 2025-12-31
[task] feat: preforward

Coach257:
### What does this PR do?

1. Add preforward pipeline. Preforwrad now supports:
- packing when `rmpad_with_position_ids=True`
- padding when `rmpad_with_position_ids=False`
- sp_slice when `sp_enabled`
2. core logic change:
- all model precalculate `position_ids`
- when `attn_imple=fa`, precalculate `cu_seqlens` & `max_seqlens`
- attention_masks always all_ones when `attn_imple=fa`, refer to [note](https://github.com/ByteDance-Seed/VeOmni/blob/7494f61cda82b48cc7f265afa351dcd80ec68cda/veomni/trainer/preforward.py#L108) for details.
- support tensor list return (anyres images t,c,h,w)
3. demo task: qwen3vlmoe
- log preforward info
image

- sp aligned with rmpad_true or rmpad_false with bs=1 or bs > 1
image
image
All tested on a toy two-layer qwen3vlmoe.

### DIFF
--- a/tasks/omni/train_qwen_vl.py
+++ b/tasks/omni/train_qwen_vl.py
@@ -233,7 +233,6 @@ def main():
         enable_reentrant=args.train.enable_reentrant,
         enable_forward_prefetch=args.train.enable_forward_prefetch,
     )
-    model.language_model.layers = torch.nn.ModuleList(model.language_model.layers[:2])  # clip layers
     optimizer = build_optimizer(
         model,
         lr=args.train.lr,


veomni/trainer/preforward.py
```diff
@@ -0,0 +1,342 @@
...
+            assert self.sp_pad_value is not None and self.sp_pad_scale is not None, (
+                "sp_pad_value and sp_pad_scale must be specified when sp_slice is True"
+            )
+
+        assert (self.sp_pad_value is None) == (self.sp_pad_scale is None), (
+            "sp_pad_value and sp_pad_scale must be specified together or None"
+        )
+
+
+class Preforward:
```
Luosuu, line 70: is it possible to add a unit test for this class

tasks/vlm/train_qwen3_vl_moe.py
```diff
@@ -0,0 +1,420 @@
+import json
```
Luosuu, line 1: why we need a new trainer? I think the original train_qwen_vl trainer should work?

veomni/utils/helper.py
```diff
@@ -96,16 +96,17 @@ def _compute_seqlens(
         rmpad_with_pos_ids (bool): Whether to remove the padding tokens using the position ids.
         enable_multisource (bool): Whether to enable the multi-source dataloader.
     """
-    attention_mask = micro_batch["attention_mask"]
-    if rmpad:
+    attention_mask = micro_batch["attention_mask"]  # rm this, attention_mask will be all ones
+    if rmpad:  # rm this, veomni will only have rmpad_with_pos_ids
```
Luosuu, line 100: shall we directly do assertion here to avoid any case that is not using `rmpad_with_pos_ids`?

tasks/vlm/train_qwen3_vl_moe.py
piyifan123: +1. better to simplify with a single trainer IMO.

veomni/trainer/preforward.py
piyifan123: +1

tasks/vlm/train_qwen3_vl_moe.py
```diff
@@ -0,0 +1,420 @@
...
+        data_iterator = iter(train_dataloader)
+        for _ in range(start_step, args.train.train_steps):
+            global_step += 1
+            try:
+                micro_batches: List[Dict[str, Any]] = next(data_iterator)
+            except StopIteration:
+                logger.info(f"epoch:{epoch} Dataloader finished with drop_last {args.data.drop_last}")
+                break
+
+            micro_batches = preforward_func(micro_batches)
```
piyifan123, line 292: wondering if we gauge the performance change? asking because i'm not sure if the packing/sp_splitting logic was running on the separate dataloader process but now it's blocking the main training process that creates a list of non-trivial cpu operations for the main training process to execute which result in GPU idle times.

i think this preforward way makes things simpler but wondering if it hits the performance significantly.

Coach257: Yes, this indeed introduce performance regression. But in the upcoming verl+veomni integration, verl's rollout pipeline will pass fully-sequence tensor dict directly to VeOmni. VeOmni has to perform packing and sp slicing in `preforward`.

veomni/utils/helper.py
Coach257: Remove rmpad in the future as too many tasks related currently.

tasks/vlm/train_qwen3_vl_moe.py
Coach257: This is a demo task for verl+veomni integration. I have moved this pr to [branch:verl-trainer](https://github.com/ByteDance-Seed/VeOmni/tree/verl_trainer)

PR #355, opened 2025-12-31
[ops] fix: RoPE fused op caused freqs shape bug

zbian99:
### What does this PR do?

In PR #336, we attempted to add the RoPE NPU fused operator for the Wan2.1 model, which required adjusting the shapes of `cos` and `sin` (see [code](https://github.com/zbian99/VeOmni/blob/8d1c318ab716d149be2e13599ba7eb067ffa6ead/veomni/models/transformers/wan/modeling_wan.py#L625C1-L626C83). Since RoPE is called twice, modifying the shapes of `cos` and `sin` directly within the `forward` pass could save some time; however, this behavior presents a risk:

Although the [rope_apply](https://github.com/zbian99/VeOmni/blob/8d1c318ab716d149be2e13599ba7eb067ffa6ead/veomni/models/transformers/wan/modeling_wan.py%23L244C1-L250C29) function in `modeling_wan.py` does not explicitly use `cos` and `sin` (it uses `freqs` instead), the implementation of `rope_apply` might be replaced when applying `veomni.ops.dit.rope_wan.rotary` or other patches. Consequently, modifying `cos` and `sin` could lead to dimension errors.

Therefore, to ensure code correctness and maintainability, we have moved the shape adjustment of `cos` and `sin` into `npu_patch.py`.

### DIFF
--- a/veomni/models/transformers/wan/modeling_wan.py
+++ b/veomni/models/transformers/wan/modeling_wan.py
@@ -622,8 +622,8 @@ class WanModel(PreTrainedModel):
             .reshape(f * h * w, 1, -1)
             .to(x.device)
         )
-        cos = freqs.real.unsqueeze(0).unsqueeze(-1).contiguous().to(torch.float32)
-        sin = freqs.imag.unsqueeze(0).unsqueeze(-1).contiguous().to(torch.float32)
+        cos = freqs.real.squeeze().contiguous()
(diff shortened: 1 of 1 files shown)


zbian99:
ascend

zbian99:
label ascend

onehaitao:
lgtm

PR #322, opened 2025-12-24
[model, ops] feat: add Qwen3 sequence classification model and loss for embedding classification.

yiwzhao:
feat: Add a data collator to support embedding classification.

PR #376, opened 2026-01-07
[data] feat: add data collators for embedding classification.

yiwzhao:
feat: Add data collators to support embedding classification.
split the data collator part from https://github.com/ByteDance-Seed/VeOmni/pull/322

### DIFF
--- a/veomni/data/data_collator.py
+++ b/veomni/data/data_collator.py
@@ -156,6 +156,7 @@ class DataCollatorWithPositionIDs(DataCollator):
     """
 
     mask_boundary_labels: bool = True
+
     def __post_init__(self):
         self.sp_enabled = get_parallel_state().sp_enabled
 


veomni/data/data_collator.py
```diff
@@ -325,3 +325,91 @@ def __call__(self, batch: Sequence[Dict[str, "torch.Tensor"]]) -> Dict[str, "tor
             add_flash_attention_kwargs_from_position_ids(batch)
 
         return batch
+
+
+@dataclass
+class ClassificationDataCollatorWithPositionIDs(DataCollator):
```
Coach257: maybe reuse DataCollatorWithPositionIDs if no diff here

veomni/data/data_collator.py
```diff
@@ -325,3 +325,91 @@ def __call__(self, batch: Sequence[Dict[str, "torch.Tensor"]]) -> Dict[str, "tor
...
+      - NO label shift
+      - NO masking of last token
+    Keep everything else identical.
+    """
+
+    def __call__(self, batch: Sequence[Dict[str, torch.Tensor]]) -> Dict[str, torch.Tensor]:
+        input_ids = batch.pop("input_ids")
+
+        # CHANGED: do NOT shift labels for seq-cls token-level labels
+        labels = batch.pop("labels").contiguous()
```
Coach257: i think we cound add a parameter named ‘shift_labels=True` in TextSequenceShardCollator to control whether to keep the labels or shift and mask the labels

veomni/data/data_collator.py
yiwzhao: There is a difference: DataCollatorWithPositionIDs masks the labels corresponding to the "boundary tokens" of each packed subsequence, while ClassificationDataCollatorWithPositionIDs does not perform this masking step. However, I think we can make "whether to mask boundary labels" a configurable option, so we don't need to create a new class.

yiwzhao: Updated.

veomni/data/data_collator.py
yiwzhao: I agree. I added two switches to the original class, which by default maintain the old behavior while also supporting sequence classification.

veomni/data/data_collator.py
```diff
@@ -275,18 +291,25 @@ def sp_padding(
...
-            labels[batch["cu_seqlens"][1:-1] - 1] = IGNORE_INDEX
+        if self.shift_labels:
+            labels = batch.pop("labels")[..., 1:].contiguous()  # shift labels
+            labels = F.pad(labels, (0, 1), "constant", IGNORE_INDEX)
+        else:
+            labels = batch.pop("labels").contiguous()
+
+        if self.mask_boundary_labels:
+            if self.rmpad_with_pos_ids:  # mask the last token of each sequence
+                cu_seqlens = pos2culen(batch["position_ids"])
```
piyifan123, line 305: don't we need cu_seqlens for not mask_boundary_labels case as well?

tests/data/test_seqcls_collators.py
```diff
@@ -0,0 +1,168 @@
...
+        # return ((cu_q, max_q), (cu_k, max_k)) or whatever your real fn returns
+        return (cu, 3), (cu, 3)
+
+    monkeypatch.setattr(m, "prepare_fa_kwargs_from_position_ids", fake_prepare)
+
+    collator = m.DataCollatorWithPositionIDs(mask_boundary_labels=False)
+    batch = collator(features_two_samples)
+
+    assert batch["input_ids"].shape == (1, 5)
+    assert "position_ids" in batch
```
piyifan123: can we assert actual values and in other tests?

tests/data/test_seqcls_collators.py
```diff
@@ -0,0 +1,168 @@
+import types
+
+import pytest
+import torch
+
+
+IGNORE_INDEX = -100
```
piyifan123: can we import veomni constant

veomni/data/data_collator.py
yiwzhao: No, we don't need it, because calculating cu_seqlens is an auxiliary step needed only for creating the boundary mask; it's not batch["cu_seqlens"], and it won't be used again later.

tests/data/test_seqcls_collators.py
yiwzhao: Yes, updated.

tests/data/test_seqcls_collators.py
yiwzhao: Sure, updated.

PR #379, opened 2026-01-08
[trainer] feat: trainer framework

Coach257:
### What does this PR do?

1. base trainer (seperate training loop and callbackfuncs)
- callbacks: functions that not related to training precision & efficiency. For example, environ meter, checkpoint, wandb.
2. text_trainer & vlm_trainer
- support qwen3; qwen3vlmoe
- only small modifications of base_trainer
3. a demo of vlm_rl_trainer, used for verl
4. alignment:
- rmpad sp2 & sp1 aligned
- padding bs1sp1 & bs2sp1 & bs1sp2 & bs2sp2 aligned
- qwen3 & qwen3vlmoe aligned with previous `tasks`
- vlm_rl_trainer aligned with vlm_trainer after postforward

### Modification of veomni modules
As trainier is parallel to origin training scripts, the veomni modules mostly kept except following:
1. environ meter: environ_meter.add() is called before preforward. So the data is one origin sample that have not been packed or padded. So calculate the `seqlens` can be simple as `len(batch['attention_mask'])`
2. loss utils: mean_global_batch now returns a Dict[torch.Tensor], the total loss for backward is calcuated outside.

These changes may cause origin `tasks` failure.

However, all the `tasks` will be replaced with the new trainer framework.

### DIFF
--- a/.github/workflows/docker-build-ascend-a2.yml
+++ b/.github/workflows/docker-build-ascend-a2.yml
@@ -47,10 +47,28 @@ jobs:
         id: version
         run: |
           BRANCH_NAME=$(echo "${{ github.ref }}" | sed 's/refs\/heads\///g' | sed 's/[^a-zA-Z0-9._-]/_/g')
+          BASE_IMAGE_NAME="${{ steps.base_image.outputs.new_image_name }}"
+          QUAY_REPO="quay.io/ascend/veomni"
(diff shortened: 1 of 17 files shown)


Coach257:
Thanks @Fazziekey  #327 for trainer design.

Issue #369, opened 2026-01-05
😓😓😓


Yangr116:
Very not clear:

Qwen3vl code:
https://github.com/ByteDance-Seed/VeOmni/blob/b6fa6ccf52783e138a06072b2759b4792f8b6178/veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py#L1370-L1396

Qwen2.5vl code:
https://github.com/ByteDance-Seed/VeOmni/blob/b6fa6ccf52783e138a06072b2759b4792f8b6178/veomni/models/transformers/qwen2_5vl/modeling_qwen2_5_vl.py#L2022-L2039

能让他们保持一致吗？😓

Luosuu:
I helped you to make the title more informative. Please keep your communication professional and helpful.

As for the issue, please follow the Qwen3VL modeling since Qwen2.5VL is still under refactoring. Thank you for your understanding.

PR #399, opened 2026-01-15
[model] feat: patch qwen3vlmoe; qwen3vl; qwen25vl

Coach257:
### What does this PR do?
1. Patch transformers 4.57.3
- qwen3vlmoe
- qwen3vldense
- qwen25vl

2. fix train_qwen_vl training joint image/video/text data
3. fix qwen25vl video data training #369 

All patched modeling loss aligned with hf native modeling.

### DIFF
--- a/veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
+++ b/veomni/models/transformers/qwen3_vl/modeling_qwen3_vl.py
@@ -72,7 +72,8 @@ logger = logging.get_logger(__name__)
 
 # ================================================================
 # Patch: Qwen3VLVisionAttention.forward
-# 1. use precomputed max_seqlen in advance to avoid per-layer cpu-gpu sync
+# 1. add flash_attention_3 support
+# 2. use precomputed max_seqlen in advance to avoid per-layer cpu-gpu sync
 # ================================================================
 def Qwen3VLVisionAttention_forward(
     self: Qwen3VLVisionAttention,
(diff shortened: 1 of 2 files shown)


Luosuu (approved):

Luosuu:
codex:

I’m concerned about `prepare_fa_kwargs_from_position_ids` in `tasks/omni/train_qwen_vl.py:349` now using the HF helper in `veomni/utils/seqlen_pos_transform_utils.py:57`, which treats `every position_id == 0` as
  a new sequence head. 

For Qwen2.5/3 VL, vision position ids often contain many zeros within a sequence (not just at the start), so this looks like it would create spurious boundaries and incorrect cu_seq_lens/max_length. Can you clarify the new position-id semantics (are zeros now only used for sequence starts)? If not, I think we should keep the previous “strict drop” boundary detection or
  compute cu_seq_lens from true packed lengths/attention masks instead.

TimYangst (approved):

Coach257:
> codex:
> 
> I’m concerned about `prepare_fa_kwargs_from_position_ids` in `tasks/omni/train_qwen_vl.py:349` now using the HF helper in `veomni/utils/seqlen_pos_transform_utils.py:57`, which treats `every position_id == 0` as a new sequence head.
> 
> For Qwen2.5/3 VL, vision position ids often contain many zeros within a sequence (not just at the start), so this looks like it would create spurious boundaries and incorrect cu_seq_lens/max_length. Can you clarify the new position-id semantics (are zeros now only used for sequence starts)? If not, I think we should keep the previous “strict drop” boundary detection or compute cu_seq_lens from true packed lengths/attention masks instead.

The origin implementation cause differ when useing padding & bs>1 (long padding 0 sequence).

Issue #402, opened 2026-01-16
[perf] pad packed input to avoid triton kernel compiling overhead


Luosuu:
### Feature request

As prerequisite of #401 

### Motivation

Packed input size can vary due to dynamic batching, which would trigger Triton kernel to compile for the new input shapes during training, blocking GPU execution.

Example:

<img width="650" height="3706" alt="Image" src="https://github.com/user-attachments/assets/71c32bd5-d370-4554-9645-622ea4449a47" />

### Your contribution

* explore design to pad packed input

Luosuu:
Resolved by #410

Issue #401, opened 2026-01-16
RFC: torch.compile & CUDAGraph


Luosuu:
### Feature request

* support for torch.compile & cudagraph

### Motivation

Latest platforms like GB200/910C are equipped with ARM CPUs, which are expected to be slower and can cause significant kernel launch overheads. To mitigate that, we need torch.compile & CUDAGraph to eliminate the kernel launch overheads. 

Furthermore, with torch.compile & cudagraph, we can also avoid heavy triton compiling overheads when input size varies. For example, in the following figure, currently LigerRMSNorm compiling is long enough to block kernel execution on GPU.

<img width="650" height="3706" alt="Image" src="https://github.com/user-attachments/assets/53db8119-7a22-4f11-9d23-39869ee347c8" />

To achieve this, it is enforced to pad the packed input which needs to be considered after we decide the trainer design (#326)

### Your contribution

contribute to
*  trainer design and implementation
* torch.compile integration

Luosuu:
cc @piyifan123 @Coach257 @FoolPlayer

piyifan123:
IIUC, implementing padding is a prerequisite for enabling torch.compile/cuda graph? If so, can we add another issue for that and link to this?

Luosuu:
> IIUC, implementing padding is a prerequisite for enabling torch.compile/cuda graph? If so, can we add another issue for that and link to this?

Sure

TueVNguyen:
Your graph is so cool! Can I ask which tool/scripts you used.

Luosuu:
@TueVNguyen Hi, you can get torch profiler trace by setting `enable_profiling: True`

Luosuu:
Now `pad_packed_input` has been implemented. Next step would be looking into per-block torch.compile to workaround communication operations. 

For example, when enabling FSDP2 for dense models, AllGather happens in between decoder layers and we compile the decoder layer forward only so that no communication is involved in torch.compile().

PR #410, opened 2026-01-19
[data, perf, ops] feat: option to pad packed input to a fixed shape for text-only models

Luosuu:
### What does this PR do?

Related issues: #401 #402 

VeOmni expects data to be feed with `rmpad_with_pos_ids` which would pack sequences into one batch and lead to different input length for every training iteration. This will consequently trigger recompile of the LigerRMSNorm backward triton kernel, blocking GPU execution (often observed in the first backward layer).

We introduce a new data collator that remove padding with pos_ids and then pad the packed input to a fixed shape for convenience of CUDAGraph & torch.compile()

### Test

- [x] new data tests
- [x] varlen FA output accuracy test: padded vs. non-padded 
- [x] e2e smoke tests
  - [x] runnable
  - [x] accuracy 
  - [ ] perf 

### API and Usage Example

> Demonstrate how the API changes if any, and provide usage example(s) if possible.

```yaml
train:
  pad_packed_input: true
  pad_packed_to_length: 4096 # if not set this would be micro_batch_size * max_seq_len
```

### Design & Code Changes

Similar to SP, we directly pass the padded input into varlen FA kernel since we are providing `cu_seqlens`.
We add a test to verify that varlen FA can produce exact same outputs between padded and non-padded inputs with the same correct `cu_seqlens`

Effects: 
1. all the modeling are not touched
4. RMSNorm and MLP see the same shapes at the cost of additional compute for padded token

### DIFF
--- a/veomni/data/data_collator.py
+++ b/veomni/data/data_collator.py
@@ -224,6 +224,8 @@ class DataCollatorWithPositionIDsAndPadding(DataCollator):
             return torch.cat([tensor, pad], dim=-1)
 
         batch["input_ids"] = pad_last_dim(batch["input_ids"], 0)
+        if "attention_mask" in batch:
+            batch["attention_mask"] = pad_last_dim(batch["attention_mask"], self.attention_mask_pad_value)
         if "labels" in batch:
             batch["labels"] = pad_last_dim(batch["labels"], IGNORE_INDEX)
         if "position_ids" in batch:
--- a/tests/data/test_seqcls_collators.py
+++ b/tests/data/test_seqcls_collators.py
@@ -142,6 +142,7 @@ def test_data_collator_padded_packed_length_is_static(features_two_samples):
     tail = slice(5, pad_to_length)
     assert torch.equal(out["input_ids"][0, tail], torch.zeros(pad_to_length - 5, dtype=torch.long))
     assert torch.equal(out["attention_mask"][0, tail], torch.ones(pad_to_length - 5, dtype=torch.long))
+    assert torch.equal(out["attention_mask"][0, tail], torch.ones(pad_to_length - 5, dtype=torch.long))
     assert torch.equal(out["labels"][0, tail], torch.full((pad_to_length - 5,), IGNORE_INDEX, dtype=torch.long))
     assert torch.equal(
         out["position_ids"][0, tail],


veomni/data/data_collator.py
```diff
@@ -188,6 +189,46 @@ def __call__(self, features: Sequence[Dict[str, "torch.Tensor"]]) -> Dict[str, "
...
+            pad = torch.full(pad_shape, pad_value, dtype=tensor.dtype, device=tensor.device)
+            return torch.cat([tensor, pad], dim=-1)
+
+        batch["input_ids"] = pad_last_dim(batch["input_ids"], self.pad_token_id)
+        if "attention_mask" in batch:
+            batch["attention_mask"] = pad_last_dim(batch["attention_mask"], self.attention_mask_pad_value)
+        if "labels" in batch:
+            batch["labels"] = pad_last_dim(batch["labels"], IGNORE_INDEX)
+        if "position_ids" in batch:
+            batch["position_ids"] = pad_last_dim(batch["position_ids"], self.position_id_pad_value)
```
piyifan123: do we need to care about cu_seqlens for padded tokens?

veomni/data/data_loader.py
```diff
@@ -71,6 +72,8 @@ def build_native_dataloader(
     bsz_warmup_ratio: float = 0.02,
     bsz_warmup_init_mbtoken: int = 200,
     dyn_bsz: bool = True,
+    pad_packed_to_length: Optional[int] = None,
+    pad_packed_token_id: int = 0,
```
Coach257: Do we really need this arg? input id always 0; labels always -100; position ids always 0; attention mask always 1.

Luosuu: sure. thanks for the suggestion

veomni/data/data_collator.py
Luosuu: no, please see `tests/ops/test_flash_attn_varlen_padding.py`

docs/usage/rmpad_with_pos_ids_and_dyn_bsz.md
```diff
@@ -73,3 +73,14 @@ Example fixed batch size = 2 samples:
 * Batch 2: S3 + S4
 * input_ids = [G H I J]
 * position_ids = [0 1 2 0]
+
+## Padding packed inputs (pad_packed_input)
+
+When `rmpad_with_pos_ids` is enabled, `pad_packed_input` can pad the packed sequence to a fixed length
+(`pad_packed_to_length`). This is useful to avoid uneven lengths that can trigger kernel recompilation.
+
```
piyifan123, line 82: can we add an example here? Or link to the original github issue and pasted the before/after comparison trace there.

veomni/data/data_collator.py
```diff
@@ -188,6 +189,53 @@ def __call__(self, features: Sequence[Dict[str, "torch.Tensor"]]) -> Dict[str, "
...
+        if pad_len == 0:
+            return batch
+
+        def pad_last_dim(tensor: "torch.Tensor", pad_value: int) -> "torch.Tensor":
+            pad_shape = list(tensor.shape)
+            pad_shape[-1] = pad_len
+            pad = torch.full(pad_shape, pad_value, dtype=tensor.dtype, device=tensor.device)
+            return torch.cat([tensor, pad], dim=-1)
+
+        batch["input_ids"] = pad_last_dim(batch["input_ids"], 0)
```
piyifan123, line 226: should we follow the similar logic in L118 to pad?

veomni/arguments/arguments_types.py
```diff
@@ -750,7 +758,22 @@ class VeOmniArguments:
     train: TrainingArguments = field(default_factory=TrainingArguments)
 
     def __post_init__(self):
-        pass
+        if self.train.pad_packed_input:
```
piyifan123, line 761: should we just assert that rmpad_with_pos_id must be true to have pad_packed_input=true?

docs/usage/rmpad_with_pos_ids_and_dyn_bsz.md
Luosuu: not sure how to share a pytorch trace with links. I will link the [#402](https://github.com/ByteDance-Seed/VeOmni/issues/402) here.

veomni/data/data_collator.py
Luosuu: I find that we cannot use pad_sequence as it does not accept a given length

PR #419, opened 2026-01-22
[parallel] fix: Correct k normalization backward pass order in async_ulysses_dit

DaiShiResearch:
### Summary
Fix a critical bug in `async_ulysses_dit.py` where k normalization backward was executed before collecting gradients from all-to-all communication, causing incorrect gradient computation. This bug severely impacts Wan model sequence parallel training.

### Changes

1. **`veomni/distributed/sequence_parallel/async_ulysses_dit.py`**:
   - Moved `grad_k = grad_k_res()` before k normalization backward
   - Fixed `needs_input_grad` index checks (off-by-one errors)

2. **`tests/parallel/ulysses/test_async_ulysses_dit.py`** (new file):
   - Added comprehensive test for `async_ulysses_dit` forward/backward correctness
   - Compares SP (sequence parallel) path vs DP (data parallel) reference
   - Tests both divisible and non-divisible sequence lengths
   - Includes non-fatal assertions with diagnostic output (ratio, max_abs_diff)

### Bug Details

**Before (incorrect order):**
```python
# k normalization backward - using wrong grad_k!
if norm_type is not None:
    grad_k, grad_norm_k_weight = rms_backward_affine(grad_k, ...)

# k grad collect - too late!
grad_k = grad_k_res()
```

**After (correct order):**
```python
# k grad collect - collect first!
grad_k = grad_k_res()

# k normalization backward - now using correct grad_k
if norm_type is not None:
    grad_k, grad_norm_k_weight = rms_backward_affine(grad_k, ...)
```

### Minor Fix: `needs_input_grad` Index Correction

Fixed off-by-one errors in `needs_input_grad` index checks. Example:

```python
# Before (incorrect)
if k_bias is not None and ctx.needs_input_grad[5]:

# After (correct)
if k_bias is not None and ctx.needs_input_grad[6]:
```

The index should match the position of each parameter in `forward()` signature. These errors could cause gradient computation to be skipped incorrectly.

### Test Methodology

The test script (`test_async_ulysses_dit.py`) validates correctness by:

1. **Forward pass**: Comparing SP output (gathered) vs DP output
2. **Backward pass**: Comparing all gradients between SP and DP paths:
   - `proj_o.weight.grad` - output projection
   - `q_proj.weight.grad`, `k_proj.weight.grad`, `v_proj.weight.grad` - QKV projections
   - `q_norm.weight.grad`, `k_norm.weight.grad` - RMSNorm weights (critical checkpoint for this bug)
   - `input.grad` - input gradients

3. **Diagnostic Metrics**:
   - `ratio = dot(sp, dp) / dot(dp, dp)`: Scale factor in least-squares sense. Should be 1.0 for correct implementation.
   - `max_abs_diff`: Maximum absolute difference between tensors.

### Verification

Run the test:
```bash
torchrun --nproc_per_node=2 -m pytest tests/parallel/ulysses/test_async_ulysses_dit.py -v -s
```

**Before fix**: `k_proj` and `k_norm` gradients show large errors (ratio ≠ 1.0)
**After fix**: All gradients pass (ratio ≈ 1.0, max_diff within tolerance)

---

## Appendix: Understanding the `ratio` Metric

The `ratio` is calculated as:
```python
ratio = torch.dot(sp.flatten(), dp.flatten()) / torch.dot(dp.flatten(), dp.flatten())
```

This is the scale factor in the least-squares sense:
- **ratio = 1.0**: SP and DP gradients are identical (correct)
- **ratio ≠ 1.0**: Systematic bias exists (incorrect)
- **ratio < 0**: Gradient direction is completely reversed (severe error)

In this bug, `k_norm.weight.grad` had ratio = -0.02, meaning the gradient was nearly orthogonal to the correct value. This would cause training to completely diverge from the correct direction.

### DIFF
--- a/tests/parallel/ulysses/test_async_ulysses_dit.py
+++ b/tests/parallel/ulysses/test_async_ulysses_dit.py
@@ -1,0 +1,456 @@
+"""
+Test script for async_ulysses_dit.py
+
+This test specifically validates the fix for the backward pass bug where
+k normalization backward was executed BEFORE the all-to-all communication
+collect (grad_k_res()), which caused incorrect gradient computation.
+
+The fix ensures the correct order:
+1. grad_k = grad_k_res()  # collect gradients first
+2. k norm backward        # then compute norm backward
+
+Run with pytest:
+    torchrun --nproc_per_node=2 -m pytest tests/parallel/ulysses/test_async_ulysses_dit.py -v -s
+
+Run directly (without pytest):
+    python tests/parallel/ulysses/test_async_ulysses_dit.py
+"""
+
+import sys
+
+import torch
+import torch.distributed as c10d
+
+from veomni.utils.device import get_device_type, get_dist_comm_backend, get_torch_device
+
+
+if not c10d.is_available() or not c10d.is_backend_available(get_dist_comm_backend()):
+    print("c10d NCCL not available, skipping tests", file=sys.stderr)
+    sys.exit(0)
+
+import pytest
+import torch.distributed as dist
+import torch.nn as nn
+import torch.nn.functional as F
+from einops import rearrange
+from torch.testing._internal.common_utils import run_tests
+
+from veomni.distributed.sequence_parallel import gather_heads_scatter_seq, gather_seq_scatter_heads
+from veomni.distributed.sequence_parallel.async_ulysses_dit import (
+    async_ulysses_output_projection as async_ulysses_dit_output_projection,
+)
+from veomni.distributed.sequence_parallel.async_ulysses_dit import (
+    async_ulysses_qkv_projection as async_ulysses_dit_qkv_projection,
+)
+from veomni.distributed.sequence_parallel.comm import (
+    get_ulysses_sequence_parallel_group,
+    set_ulysses_sequence_parallel_group,
+)
+from veomni.distributed.sequence_parallel.data import gather_outputs, slice_input_tensor
+from veomni.distributed.sequence_parallel.utils import unpadding_tensor_for_seqeunce_parallel
+from veomni.utils.helper import enable_high_precision_for_bf16, set_seed
+from veomni.utils.import_utils import is_torch_npu_available
+
+from .utils import (
+    SequenceParallelTest,
+    sync_tensor,
+)
+
+
+def _scale_ratio(sp_t: torch.Tensor, dp_t: torch.Tensor, eps: float = 1e-12) -> float:
+    """Calculate scale ratio between two tensors (sp_t / dp_t in least-squares sense)."""
+    spf = sp_t.detach().float().reshape(-1)
+    dpf = dp_t.detach().float().reshape(-1)
+    denom = torch.dot(dpf, dpf).item()
+    if denom <= eps:
+        return float("nan") if torch.dot(spf, spf).item() > eps else 1.0
+    num = torch.dot(spf, dpf).item()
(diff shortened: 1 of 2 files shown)


Luosuu:
please take a look @Crystal-jiang

Crystal-jiang:
Could you please provide the training loss results of the WAN model after this bug fix?

DaiShiResearch:
> Could you please provide the training loss results of the WAN model after this bug fix?

Hi @Crystal-jiang,

Thank you for reviewing this PR. Also thanks to @Luosuu for bringing this to your attention.

Due to project privacy constraints, I am unable to share the actual training loss data. However, based on my measurements, **the convergence curve after this fix aligns almost identically with the non-async_ulysses (standard) implementation**.

## Root Cause Analysis

The root cause of this bug is actually simpler than it might initially appear. 

By adding a shape assertion on `grad_k` at [L255](https://github.com/ByteDance-Seed/VeOmni/blob/3868ebb5e5fb366c7153d1654f9d2dfc0635c2be/veomni/distributed/sequence_parallel/async_ulysses_dit.py#L255) before the normalization backward, we can immediately detect the bug:

```python
# L255: qk normalization backward (if needed)
if norm_type is not None:
    # Add shape assertion here to detect the bug
    assert grad_k.shape == k.shape and grad_k.shape[-len(normalized_shape):] == normalized_shape, (
        f"Shape mismatch in norm backward: "
        f"grad_k.shape={grad_k.shape}, k.shape={k.shape}, normalized_shape={normalized_shape}"
    )
    if norm_type == "rmsnorm":
        grad_k, grad_norm_k_weight = fused_layer_norm_cuda.rms_backward_affine(
            grad_k,
            invvar_k,
            k,
            normalized_shape,
            norm_k_weight,
            eps,
            False,
        )
```

Running the [test script](https://github.com/DaiShiResearch/VeOmni/blob/fix/async-ulysses-dit-grad-error/tests/parallel/ulysses/test_async_ulysses_dit.py) immediately reveals the issue:

```
AssertionError: Shape mismatch in norm backward: 
  grad_k.shape=torch.Size([2, 8192, 512]), 
  k.shape=torch.Size([2, 4096, 1024]), 
  normalized_shape=torch.Size([1024])
```

**The `grad_k` tensor entering the APEX RMSNorm/LayerNorm backward has an incorrect shape**, which violates the original code's intention:
- `grad_k`: `[2, 8192, 512]` — last dim is `512`, **does NOT match `normalized_shape`**
- `k`: `[2, 4096, 1024]` — last dim is `1024`, matches `normalized_shape`
- `normalized_shape`: `[1024]`

The RMSNorm backward expects `grad_k` and `k` to have the same shape, with the last dimension equal to `normalized_shape`. However, the bug causes `grad_k` to have a completely different shape (after all-to-all transformation but before collect), breaking this invariant.

The critical issue is that **APEX's fused norm kernels do not perform shape validation** on their inputs. When fed mismatched tensors, they silently compute incorrect gradients rather than raising an error. This caused the bug to go undetected until gradient correctness was explicitly verified against a reference implementation.

## After the Fix

After applying this PR's fix (moving `grad_k = grad_k_res()` before the normalization backward), adding the same shape assertion at the same location passes successfully, confirming that `grad_k` now has the correct shape matching `k` and `normalized_shape`.

Best regards

Luosuu:
I think the tests look good? @Crystal-jiang

PR #426, opened 2026-01-26
[model] feat: Use a patch to add the modeling code for Qwen3 for Sequence Classification.

yiwzhao:
Use a patch to add the modeling code for Qwen3 for Sequence Classification.

### DIFF
--- a/veomni/models/transformers/qwen3/modeling_qwen3.py
+++ b/veomni/models/transformers/qwen3/modeling_qwen3.py
@@ -19,7 +19,7 @@ from typing import Optional, Union
 
 import torch
 import transformers.models.qwen3.modeling_qwen3 as hf_qwen3
-from transformers import Qwen3ForCausalLM, Qwen3Model, Qwen3ForSequenceClassification
+from transformers import Qwen3ForCausalLM, Qwen3ForSequenceClassification, Qwen3Model
 from transformers.cache_utils import Cache, DynamicCache
 from transformers.masking_utils import create_causal_mask, create_sliding_window_causal_mask
 from transformers.modeling_outputs import (


veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -211,11 +212,85 @@ def qwen3forcausallm_forward(
...
+            "token-level labels" convention: positions not supervised should be set to `-100`, and only the
+            supervised token(s) (e.g., the last valid token of each sample) carry a real class id in
+            `[0, ..., num_labels - 1]`. Tokens with label `-100` are ignored.
+
+            Note: `labels` should be provided for classification training tasks.
+
+    Returns:
+
+    """
+    output_attentions = output_attentions if output_attentions is not None else self.config.output_attentions
```
piyifan123: if it's useless just remove?

veomni/models/transformers/qwen3/modeling_qwen3.py
```diff
@@ -211,11 +212,85 @@ def qwen3forcausallm_forward(
...
+    )
+    hidden_states = outputs.last_hidden_state
+
+    loss = None
+    logits = None
+    if labels is not None:
+        loss, logits = self.loss_function(
+            logits=logits,
+            labels=labels,
+            hidden_states=hidden_states,
```
piyifan123, line 265: can we pass down self.num_labels to the loss_function?

veomni/models/transformers/qwen3/modeling_qwen3.py
yiwzhao: Good suggestion, it has been deleted.

veomni/models/transformers/qwen3/modeling_qwen3.py
yiwzhao: I agree, and I've changed it to pass the self.num_labels from the model.

PR #427, opened 2026-01-26
[data] feat: add arguments for sequence classification

yiwzhao:
add arguments for sequence classification


Crystal-jiang:
Could you please add corresponding configuration item documentation and update it in the following file?  
👉 https://github.com/ByteDance-Seed/VeOmni/blob/main/docs/usage/arguments.md

veomni/arguments/arguments_types.py
```diff
@@ -264,6 +264,10 @@ class DataArguments:
         default=False,
         metadata={"help": "Whether to ignore exceptions using byted dataset. Defaults to ``False``"},
     )
+    num_labels: Optional[int] = field(
```
piyifan123: can we remove this? we should get the num_labels from the model's `config.json` of the model, i.e., how many labels a model can classify should be an inherent property of the model instead of the data.

yiwzhao:
> Could you please add corresponding configuration item documentation and update it in the following file? 👉 https://github.com/ByteDance-Seed/VeOmni/blob/main/docs/usage/arguments.md

Thanks for pointing this out, it was added now.

veomni/arguments/arguments_types.py
yiwzhao: Yes, I agree, this parameter has been removed. Users need to set num_labels themselves in the model's config file.

Issue #439, opened 2026-01-30
Feature Request: Upgrade to Torch 2.9.1+cu129


piyifan123:
### Feature request

- [x] Add optional dependency for `torch==2.9.1+cu129`
- [x] Pin/confirm FA2 compatible build for torch 2.9 + CUDA 12.9
- [x] Pin/confirm FA3 compatible build for torch 2.9 + CUDA 12.9
- [x] Fix torch 2.9 regressions (incl. SP collator deprecation)
- [x] Update CI image to CUDA 12.9
- [x] Run CI with torch 2.9.1+cu129
- [x] Promote torch 2.9.1+cu129 to default
- [ ] Clean up transitional code/docs

Note:
- There is no official torch2.9 compatible prebuilt wheels for FA2 and FA3, so we will use the following 3P prebuilt wheels:
   -  FA2: https://github.com/mjun0812/flash-attention-prebuild-wheels
   -  FA3: https://github.com/windreamer/flash-attention3-wheels

#### SP collator using a non-tuple sequence for multidimensional indexing

```
/opt/tiger/modelchef/submodules/Open-VeOmni/veomni/distributed/sequence_parallel/data.py:86: UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at /pytorch/torch/csrc/autograd/python_variable_indexing.cpp:306.)
  return x[slc].contiguous()
```

### Motivation

We should upgrade VeOmni to **PyTorch 2.9.1 + CUDA 12.9 (cu129 wheels)** for two reasons:

1. **Policy / freshness**: Align with our **-1 policy** now that **Torch 2.10** has been published recently. Torch 2.9.1 is the latest “-1” target and should be supported as a first-class configuration.
2. **Kernel enablement**: Move from **CUDA 12.8 → 12.9** so we can reliably use **cuTE-based kernels** (e.g. **FlashAttention-4 (FA4)** and **SonicMoE**).

### Your contribution

TBA

piyifan123:
Done.

PR #442, opened 2026-02-02
[misc] feat: add cu129 Dockerfile for CUDA 12.9 builds

piyifan123:
Part of https://github.com/ByteDance-Seed/VeOmni/issues/439.

Get a new cu129 docker file.

the only diff from the existing cu128 dockerfile is the base image is changed to nvcr.io/nvidia/pytorch:25.06-py3.

we will delete the old cu128 one once the cu129 upgrade is done.

### DIFF
--- a/docker/cuda/Dockerfile.cu129
+++ b/docker/cuda/Dockerfile.cu129
@@ -1,0 +1,83 @@
+# https://docs.nvidia.com/deeplearning/frameworks/pytorch-release-notes/rel-25-06.html
+from nvcr.io/nvidia/pytorch:25.06-py3
+
+# Define environments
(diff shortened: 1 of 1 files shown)


FoolPlayer (approved):

FoolPlayer:
Do we need to change the Python version for uv?

piyifan123:
> Do we need to change the Python version for uv?

I don't think so.

1) https://github.com/ByteDance-Seed/VeOmni/blob/main/pyproject.toml#L14 forces to python 3.11 so uv would download a 3.11 in its venv
2) i think it's how it works today as the docker image is basically the same as the cu128 one except the base image (and both image image use py3.12).

PR #454, opened 2026-02-04
[data, task] feat: Integrate the data processor for the Qwen 2.5 and 3 Omni models from trainer into data files.

yiwzhao:
Originally, the data processors for Qwen2.5 and Qwen3 were located at the beginning of the trainer. To optimize the code and simplify the trainer, we followed the approach of Qwen-VL and moved them to the data folder.
Reference: https://github.com/ByteDance-Seed/VeOmni/blob/main/tasks/data/vlm_data_process.py

### DIFF
--- a/tasks/omni/train_qwen25_omni.py
+++ b/tasks/omni/train_qwen25_omni.py
@@ -3,13 +3,14 @@ import os
 import time
 from dataclasses import asdict, dataclass, field
 from functools import partial
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
+from typing import Any, Dict, List, Optional
 
 import torch
 import torch.distributed as dist
 import wandb
 from tqdm import trange
 
+from tasks.data.omni_data_process import process_sample_qwen_omni
 from veomni.arguments import DataArguments, ModelArguments, TrainingArguments, parse_args, save_args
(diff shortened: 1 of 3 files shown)


tasks/omni/train_qwen25_omni.py
```diff
@@ -43,141 +40,8 @@
...
-    position_id_func: "Callable",
-    **kwargs,
-):
-    """
-    Processes multimodal example with qwen2_5_vl's pre-processor.
-    """
-    source = (
-        kwargs["source_name"] if "source_name" in kwargs else sample["source_name"]
-    )  # source_name if use multisource_dataset
-    conversations = sample["conversations"] if ("conversations" in sample and sample["conversations"]) else sample
```
yiwzhao, line 70: When sample["conversations"] is a NumPy array, directly using if sample["conversations"] will cause an error. Therefore, I changed it to using len(sample["conversations"]) > 0 for checking.

tasks/omni/train_qwen3_omni.py
TimYangst, line 1: 可不可以比对一下这两个文件，看看能不能合成一个呀？

Coach257:
These two are now merged in train_vlm.py

PR #456, opened 2026-02-05
[model, data] feat: support audio processing in Qwen3OmniMoe with SP and FSDP fixes

TimYangst:
### What does this PR do?

Add audio processing support for Qwen3OmniMoe model during training, including:
- Sequence Parallelism (SP) support for the audio encoder
- FSDP dummy forward to avoid reduce-scatter hangs when ranks lack audio
- Minimum audio length padding to prevent Whisper STFT errors
- Processor kwarg aliasing for `audios` / `input_audios` compatibility

### Checklist Before Starting

- [ ] Format the PR title as `[{modules}] {type}: {description}`

### Test

Validated by running Qwen3OmniMoe training with audio data under FSDP + SP configuration. Training completes without NCCL hang or STFT shape errors.

### API and Usage Example

No public API changes. Internal audio pipeline now accepts `audios` and `input_audios` as aliases for `audio` in `Qwen3OmniMoeProcessor.__call__`.

### Design & Code Changes

**`tasks/data/vlm_data_process.py`**
- Add `_ensure_min_audio_len` helper: pads short or empty audio arrays to at least `n_fft=400` samples to satisfy Whisper's STFT constraint.
- Apply the helper to both video-extracted and standalone audio clips before passing to the processor.

**`veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py`**
- `Qwen3OmniMoeAudioEncoder.forward`:
  - Before chunking: gather `input_features` across SP ranks along the time dimension, strip SP padding.
  - Before encoder layers: slice `hidden_states` along seq dim across SP ranks, extend `cu_seqlens` for the padded tail.
  - After encoder layers: gather `hidden_states`, strip padding, then re-slice final audio tokens for downstream injection.
- `Qwen3OmniMoeAudioEncoder.dummy_forward`: new method that runs a minimal valid forward pass (one silent chunk) so FSDP reduce-scatter stays in sync on ranks without audio data.
- `Qwen3OmniMoeThinkerForConditionalGeneration.get_audio_features`:
  - Remove `NotImplementedError` for SP; delegate SP handling to `audio_tower`.
  - Add compatibility branch for pre-processed 2-D input without `feature_attention_mask`.
- `Qwen3OmniMoeThinkerForConditionalGeneration.forward`:
  - After audio encode: gather seq / scatter heads to match `inputs_embeds` layout; truncate to actual audio placeholder count.
  - When no audio is present but FSDP is active: call `dummy_forward()` and add zero contribution to keep all ranks in the allreduce.

**`veomni/models/transformers/qwen3_omni_moe/processing_qwen3_omni_moe.py`**
- Accept `audios` (Qwen2.5-Omni style) and `input_audios` (VeOmni pipeline) as aliases for `audio`, popped before `_merge_kwargs` to suppress unknown-kwarg warnings.

> [!IMPORTANT]

- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/ByteDance-Seed/VeOmni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: audio SP/FSDP integration requires a multi-GPU cluster setup not available in standard CI.

### DIFF
--- a/veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
+++ b/veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
@@ -897,7 +897,7 @@ class Qwen3OmniMoeAudioEncoder(Qwen3OmniMoePreTrainedModel):
         cu_seqlens = torch.tensor(cu_chunk_lens, device=aftercnn_lens.device).cumsum(-1, dtype=torch.int32)
 
         # Modification: SP support — slice hidden_states along seq dim (dim=0) before encoder layers.
-        # Track total unpadded length from cu_seqlens for unpadding after gather.
+        # Track total unpadded length to compute SP padding size for cu_seqlens extension.
         if get_parallel_state().sp_enabled:
             unpadded_hidden_len = cu_seqlens[-1]
             hidden_states = slice_input_tensor(hidden_states, dim=0, group=get_parallel_state().sp_group)
@@ -914,22 +914,10 @@ class Qwen3OmniMoeAudioEncoder(Qwen3OmniMoePreTrainedModel):
 
             hidden_states = layer_outputs[0]
 
-        # Modification: SP support — gather hidden_states back across ranks, then strip padding.
-        if get_parallel_state().sp_enabled:
-            hidden_states = gather_outputs(hidden_states, gather_dim=0, group=get_parallel_state().sp_group)
-            sp_hidden_padding = hidden_states.size(0) - unpadded_hidden_len
-            if sp_hidden_padding > 0:
-                hidden_states = unpad_tensor(hidden_states, dim=0, padding_size=sp_hidden_padding)
-
         hidden_states = self.ln_post(hidden_states)
         hidden_states = self.proj1(hidden_states)
         hidden_states = self.act(hidden_states)
         hidden_states = self.proj2(hidden_states)
-
-        # Modification: SP support — slice final audio tokens along seq dim for downstream injection.
-        if get_parallel_state().sp_enabled:
-            hidden_states = sp_pad_and_slice(hidden_states, dim=0, pad_value=0)
-
         return BaseModelOutput(last_hidden_state=hidden_states)
 
     def dummy_forward(self):


veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
```diff
@@ -884,6 +896,16 @@ def forward(
                 cu_chunk_lens += [remainder]
         cu_seqlens = torch.tensor(cu_chunk_lens, device=aftercnn_lens.device).cumsum(-1, dtype=torch.int32)
 
+        # Modification: SP support — slice hidden_states along seq dim (dim=0) before encoder layers.
+        # Track total unpadded length from cu_seqlens for unpadding after gather.
+        if get_parallel_state().sp_enabled:
+            unpadded_hidden_len = cu_seqlens[-1]
+            hidden_states = slice_input_tensor(hidden_states, dim=0, group=get_parallel_state().sp_group)
```
piyifan123, line 903: curious on why we don't call sp_pad_and_slice here?

veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
```diff
@@ -892,12 +914,40 @@ def forward(
...
+            if sp_hidden_padding > 0:
+                hidden_states = unpad_tensor(hidden_states, dim=0, padding_size=sp_hidden_padding)
+
         hidden_states = self.ln_post(hidden_states)
         hidden_states = self.proj1(hidden_states)
         hidden_states = self.act(hidden_states)
         hidden_states = self.proj2(hidden_states)
+
+        # Modification: SP support — slice final audio tokens along seq dim for downstream injection.
+        if get_parallel_state().sp_enabled:
```
piyifan123: so the process is slice -> gather -> run the L924-L927 and then slice again? why do we need to gather and slice again?

veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
TimYangst: `slice_input_tensor` is used here (not `sp_pad_and_slice`) because the audio encoder uses varlen attention with cu_seqlens. After slicing, we manually compute `pad_seq_len` and extend `cu_seqlens` to cover the SP padding tail, so the encoder's varlen attention correctly handles padded tokens. `sp_pad_and_slice` only operates on the tensor and doesn't update `cu_seqlens`, so using it here alone wouldn't be sufficient.

veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
TimYangst: The gather at line 919 is necessary to strip SP padding correctly — each rank's slice may contain padded tail tokens, and we need the full gathered tensor to know exactly how many real tokens there are (line 920-922). 
After stripping padding, ln_post/proj1/proj2 run on the complete clean features. The re-slice at line 930 is then needed because inputs_embeds (from line 2373) is in the seq-parallel sliced format, so audio_features must be in the same layout for the masked_scatter injection to work correctly.

piyifan123: hidden_states = self.ln_post(hidden_states)
        hidden_states = self.proj1(hidden_states)
        hidden_states = self.act(hidden_states)
        hidden_states = self.proj2(hidden_states) 

looks like these 4 operations are on the individual feature/token level (instead of the audio sequence level). can it still be done without gather but SP?

TimYangst: Updated, yes, it totally makes sense, thanks for catching that.

piyifan123 (approved):

tasks/data/vlm_data_process.py
```diff
@@ -254,15 +255,27 @@ def _process_sample_omni(
     else:
         audio_audios = []
 
+    # Whisper STFT requires input length > n_fft/2 (pad=200, so need len > 200).
+    # Use n_fft=400 as the minimum safe length; pad short/empty audio with silence.
+    _WHISPER_MIN_AUDIO_LEN = 400
+
+    def _ensure_min_audio_len(audio):
+        if audio is None:
+            return np.zeros(_WHISPER_MIN_AUDIO_LEN, dtype=np.float32)
```
Coach257, line 264: Why we need init zero audio here

PR #461, opened 2026-02-06
[ckpt] feat: support distributed huggingface safetensor save via HuggingFaceStorageWriter

Ziyi-Wang:
## What does this PR do?
                                                                                                                                                                                                                                                                             
  ### Summary                                                                                                                                                                                                                                                                 

  Add distributed HuggingFace safetensors saving using PyTorch's HuggingFaceStorageWriter (torch >= 2.9). On supported versions, model weights are saved directly from the live FSDP model in parallel across all ranks, replacing the legacy rank-0-only checkpoint conversion path

  ### Changes

  - veomni/utils/save_safetensor_utils.py: Refactor save_hf_safetensor into a dispatcher that routes between two modes:
    - Distributed mode (_save_hf_safetensor_distributed): Uses HuggingFaceStorageWriter with save_distributed=True and consolidation for parallel saving across all ranks (torch >= 2.9, ckpt_manager="dcp", non-LoRA).
    - Legacy mode (_save_hf_safetensor_legacy): Falls back to rank-0-only ckpt_to_state_dict conversion on older PyTorch versions, non-DCP managers, or LoRA training.
  - veomni/checkpoint/dcp_checkpointer.py: Add HuggingFaceStorageWriter support in DistributedCheckpointer.save() — uses flat state dict to match HF weight-map FQNs, casts fp32 params to bf16 on a copy (preserving original model dtypes), filters out tied weights not in the weight map, and cleans up intermediate sharded directories after synchronous save.
  - veomni/arguments/arguments_types.py: Add safetensor_idx_path field to ModelArguments with auto-resolution from model_path; parse model.safetensors.index.json into fqn_to_index_mapping for sharded output. Add directory layout comments to TrainingArguments.
  - tasks/train_torch.py, tasks/omni/train_*.py (6 files): Update all training entry points to call save_hf_safetensor on all ranks (not just rank 0) with the new parameters (model, fqn_to_index_mapping, save_hf_safetensor_path). Initialize model_assets = None before the rank-0 guard where needed.
  - tests/checkpoints/test_trainer_saveload.py: Add dtype preservation assertions (capture_param_dtypes / assert_param_dtypes_unchanged) to verify the save operation does not mutate model parameter dtypes in-place. Update test to call save_hf_safetensor on all ranks with verification on rank 0.
### Known Limitations

  - Distributed mode currently does not support LoRA (train_architecture="lora"); LoRA workloads fall back to the legacy rank-0 path. Distributed LoRA saving will be added in a follow-up PR.


### example output hf_ckpt folder
hf_ckpt
├── added_tokens.json
├── chat_template.jinja
├── config.json
├── merges.txt
├── model-00001-of-00003.safetensors
├── model-00002-of-00003.safetensors
├── model-00003-of-00003.safetensors
├── model.safetensors.index.json
├── sharded, (saved by each rank individually before merge, we let the user decide whether to delete it for safety)
│   ├── shard-00001-model-00001-of-00003.safetensors
│   ├── shard-00001-model-00002-of-00003.safetensors
│   ├── shard-00001-model-00003-of-00003.safetensors
│   ├── shard-00002-model-00001-of-00003.safetensors
│   ├── shard-00002-model-00002-of-00003.safetensors
│   ├── shard-00002-model-00003-of-00003.safetensors
│   ├── shard-00003-model-00001-of-00003.safetensors
│   ├── shard-00003-model-00002-of-00003.safetensors
│   ├── shard-00003-model-00003-of-00003.safetensors
│   ├── shard-00004-model-00001-of-00003.safetensors
│   ├── shard-00004-model-00002-of-00003.safetensors
│   ├── shard-00004-model-00003-of-00003.safetensors
│   ├── shard-00005-model-00001-of-00003.safetensors
│   ├── shard-00005-model-00002-of-00003.safetensors
│   ├── shard-00005-model-00003-of-00003.safetensors
│   ├── shard-00006-model-00001-of-00003.safetensors
│   ├── shard-00006-model-00002-of-00003.safetensors
│   ├── shard-00006-model-00003-of-00003.safetensors
│   ├── shard-00007-model-00001-of-00003.safetensors
│   ├── shard-00007-model-00002-of-00003.safetensors
│   ├── shard-00007-model-00003-of-00003.safetensors
│   ├── shard-00008-model-00001-of-00003.safetensors
│   ├── shard-00008-model-00002-of-00003.safetensors
│   └── shard-00008-model-00003-of-00003.safetensors
├── special_tokens_map.json
├── tokenizer_config.json
├── tokenizer.json
└── vocab.json

### DIFF
--- a/veomni/utils/save_safetensor_utils.py
+++ b/veomni/utils/save_safetensor_utils.py
@@ -29,6 +29,7 @@ from veomni.utils.import_utils import is_torch_version_greater_than
 logger = helper.create_logger(__name__)
 
 
+@torch.no_grad()
 def get_model_save_state(
     model: torch.nn.Module,
     fqn_to_index_mapping: Optional[Dict[str, int]],
@@ -104,6 +105,7 @@ def _save_hf_safetensor_distributed(
         state_dict=save_state,
         storage_writer=storage_writer,
     )
+    del save_state  # Free copied tensors (e.g. fp32->bf16) to reduce peak memory
     if dist.is_initialized():
         dist.barrier()
     gc.collect()


tasks/train_torch.py
```diff
@@ -391,14 +391,35 @@ def main():
...
     # save model in huggingface's format
-    if args.train.global_rank == 0 and args.train.save_hf_weights and save_checkpoint_path is not None:
-        hf_weights_path = os.path.join(save_checkpoint_path, "hf_ckpt")
-        model_state_dict = ckpt_to_state_dict(
-            save_checkpoint_path=save_checkpoint_path,
-            ckpt_manager=args.train.ckpt_manager,
-        )
-        save_model_weights(hf_weights_path, model_state_dict, model_assets=model_assets)
-        logger.info_rank0(f"Huggingface checkpoint saved at {hf_weights_path} successfully!")
+    if args.train.save_hf_weights and args.train.save_safetensor_path is not None:
```
piyifan123: can we hide logic in save_model_weights (or refactor it?) it's painful to get this impl details in trainer IMO

Ziyi-Wang: yes I will extract the logic into a function and let all trainers to call it

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -288,9 +290,27 @@ def save(
         # saving extra_state first to gurantee that every saved model/optimizer ckpts have their extra_state saved before them
         cls._save_extra_state(checkpoint_dir=checkpoint_dir, state=state)
 
-        save_state = {"model": ModelState(state["model"])}
-        if "optimizer" in state:
-            save_state["optimizer"] = OptimizerState(model=state["model"], optimizer=state["optimizer"])  # type: ignore[index]
+        if isinstance(storage_writer, HuggingFaceStorageWriter):
+            for param in state["model"].parameters():
+                if param.data.dtype == torch.float32:
+                    param.data = param.data.to(torch.bfloat16)
```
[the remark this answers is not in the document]
Luosuu: I think this may make sense actually.. I am not sure whether this is true but we'd better perform copy first so that we only do hard conversion on copied params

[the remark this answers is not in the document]
Luosuu: can you add a test for this as well? Thanks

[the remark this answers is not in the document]
Ziyi-Wang: Sounds good. I've moved this part after save_state = ModelState(state["model"]).state_dict() so the conversion is performed on a copy of the state dict, avoiding any side effects. Also added an assert_param_dtypes_unchanged test in test_trainer_saveload.py to verify that dtypes remain unchanged after calling the HF safetensor save.

veomni/arguments/arguments_types.py
```diff
@@ -147,6 +154,28 @@ def __post_init__(self):
         if self.tokenizer_path is None:
             self.tokenizer_path = self.config_path
 
+        # Auto-resolve safetensor_idx_path from model_path if not specified
```
piyifan123, line 157: can we do this only if save_hf_weights is enabled?

veomni/arguments/arguments_types.py
```diff
@@ -147,6 +154,28 @@ def __post_init__(self):
...
+                    "saved safetensor will be a single file instead of sharded."
+                )
+
+        # Parse fqn_to_index_mapping from safetensor index json
+        self.fqn_to_index_mapping = None
+        if self.safetensor_idx_path is not None:
+            with open(self.safetensor_idx_path) as f:
+                weight_map = json.load(f)["weight_map"]
+            self.fqn_to_index_mapping = {fqn: int(filename.split("-")[1]) for fqn, filename in weight_map.items()}
+        if self.fqn_to_index_mapping is None:
```
piyifan123, line 169: the logic is convoluted here. you have already checked self.safetensor_idx_path before and have another 2 if branch to do the same logic. can we refator to keep it simple?

veomni/arguments/arguments_types.py
```diff
@@ -716,6 +745,12 @@ def __post_init__(self):
             )
 
         # save paths
+        # output_dir/
+        # ├── checkpoints/          # DCP training checkpoints (model + optimizer + extra_state)
```
piyifan123, line 744: where will hf weight be saved?

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -10,6 +10,14 @@
     FileSystemReader,
     FileSystemWriter,
 )
+
+from ..utils.import_utils import is_torch_version_greater_than
+
+
+if is_torch_version_greater_than("2.9"):
```
piyifan123: nit: add a comment on why checking torch 2.9 here

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -288,15 +296,61 @@ def save(
...
+            save_state = {"model": ModelState(state["model"])}
+            if "optimizer" in state:
+                save_state["optimizer"] = OptimizerState(model=state["model"], optimizer=state["optimizer"])  # type: ignore[index]
 
         if storage_writer is None:
             storage_writer = cls._create_storage_writer(checkpoint_dir)
 
         cls.execute_save(save_state=save_state, storage_writer=storage_writer, save_async=save_async)
 
+        # Clean up the intermediate per-shard directory created by HuggingFaceStorageWriter.
```
piyifan123: can we keep it as is? deletion is dangerous IMO.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -288,15 +296,61 @@ def save(
         # saving extra_state first to gurantee that every saved model/optimizer ckpts have their extra_state saved before them
         cls._save_extra_state(checkpoint_dir=checkpoint_dir, state=state)
 
-        save_state = {"model": ModelState(state["model"])}
-        if "optimizer" in state:
-            save_state["optimizer"] = OptimizerState(model=state["model"], optimizer=state["optimizer"])  # type: ignore[index]
+        save_using_hf_storage_writer = HuggingFaceStorageWriter is not None and isinstance(
```
piyifan123: can we just write a standalone logic to save instead of changing dcp_checkpointer? 

reasons:
1. let's keep this simple to only do the standard dcp save/load
2. the logic to save hf tensor can be encapsulated in save_safetenor_utils.py imo.
3. the checkpointer interface asks for load as well  but here we don't have load ability which makes the logic asymmetric here. (i.e., can we just implement the save without using the checkpointer interface?)

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -288,15 +296,61 @@ def save(
...
+            storage_writer, HuggingFaceStorageWriter
+        )
+        if save_using_hf_storage_writer:
+            # When saving HF safetensor files, only sync mode is supported
+            assert not save_async, "Async save is not supported for HuggingFaceStorageWriter."
+            # Use flat state dict so DCP FQNs match the original HF weight_map keys
+            # (e.g. "model.embed_tokens.weight" instead of "model.model.embed_tokens.weight")
+            save_state = ModelState(state["model"]).state_dict()
```
piyifan123: would this cause a all_gather on weights? @Luosuu cc

veomni/utils/save_safetensor_utils.py
```diff
@@ -12,40 +12,122 @@
...
+        train_architecture: Training architecture type. Used for routing (legacy when "lora")
+            and to filter LoRA weights in legacy mode.
+        save_checkpoint_path: [Legacy only] Path to the distributed checkpoint for conversion.
+        output_dir: [Legacy only] Output directory passed to ``ckpt_to_state_dict``.
+            Required by non-dcp checkpoint managers (e.g., omnistore).
+        model: [Distributed only] Live FSDP model for distributed save.
+        fqn_to_index_mapping: [Distributed only] Maps FQNs to safetensors file indices
+            for multi-file output.
+    """
+    use_distributed = is_torch_version_greater_than("2.9") and train_architecture != "lora" and ckpt_manager == "dcp"
```
Ziyi-Wang, line 175: Distributed mode for LoRA will be supported in a following MR to minimize the code review effort.

veomni/arguments/arguments_types.py
Ziyi-Wang: Can't guard this — fqn_to_index_mapping is in ModelArguments and save_hf_weights is in TrainingArguments, so neither dataclass can see the other's field. We could put the logic in VeOmniArguments.__post_init__ which has access to both, but only train_torch.py uses  
  VeOmniArguments from arguments_types.py — the other trainers define their own Arguments class locally.

veomni/arguments/arguments_types.py
Ziyi-Wang: simplified

veomni/arguments/arguments_types.py
Ziyi-Wang: updated the chart.  HF safetensors saved under the last checkpoint folder

veomni/checkpoint/dcp_checkpointer.py
Ziyi-Wang: No, ModelState.state_dict() calls get_model_state_dict which returns local shards (DTensor) per rank, no all_gather. The consolidation happens downstream in HuggingFaceStorageWriter during execute_save, where each rank writes its shard and rank 0 consolidates.

veomni/checkpoint/dcp_checkpointer.py
Ziyi-Wang: Sure, we can keep it. Just a heads-up: the hf_ckpt folder will be ~2x the model size since the intermediate sharded/ directory contains per-rank safetensor files alongside the consolidated ones.

veomni/checkpoint/dcp_checkpointer.py
Ziyi-Wang: sounds good, refactored

veomni/utils/save_safetensor_utils.py
```diff
@@ -12,40 +12,172 @@
...
+from veomni.models import save_model_assets, save_model_weights
 from veomni.utils import helper
+from veomni.utils.import_utils import is_torch_version_greater_than
 
 
 logger = helper.create_logger(__name__)
 
 
-def save_hf_safetensor(
+def get_model_save_state(
```
piyifan123, line 33: check if we should annotate no_grad() somewhere so that we don't mess up the auto grad graph.

veomni/utils/save_safetensor_utils.py
```diff
@@ -12,40 +12,172 @@
...
+    logger.info_rank0("Starting distributed HuggingFace safetensors save...")
+    dist.barrier()
+    start_time = time.time()
+    dcp.save(
+        state_dict=save_state,
+        storage_writer=storage_writer,
+    )
+    if dist.is_initialized():
+        dist.barrier()
+    gc.collect()
```
piyifan123, line 111: would this properly delete the copied bf16 tensors?

tests/checkpoints/test_trainer_saveload.py
```diff
@@ -324,19 +340,31 @@ def main():
     # After training on all epochs, save model in huggingface's format and verify against DCP checkpoint
     if Checkpointer.dcp_save_future is not None:
         Checkpointer.dcp_save_future.result()
-    if args.train.global_rank == 0 and args.train.save_hf_weights and save_checkpoint_path is not None:
+    if args.train.save_hf_weights and save_checkpoint_path is not None:
```
piyifan123, line 343: can you remind me which tests cases have save_hf_weights=True/False respectively?

veomni/utils/save_safetensor_utils.py
Ziyi-Wang: What gc.collect() does here is clean up any unreachable cyclic garbage. It will clean things like internal temporary objects that dcp.save() may have created and abandoned, but it won't free save_state. save_state will be freed after the function returns. 

I added  del save_state at L108 to do it explicitly and for better readability

tests/checkpoints/test_trainer_saveload.py
Ziyi-Wang: This test file has two test suites, the test suite test_trainer_saveload won't save hf weights. The test_trainer_save_hf_safetensor will save hf weights and only runs on qwen3_moe model. 

I can do a follow up PR to make save_hf_weights as the paramter of the _run_trainer* function to further enhance the readbility of the tests.

veomni/utils/save_safetensor_utils.py
Ziyi-Wang: No worries, state_dict() returns detached tensors, so the .to() calls won't affect the autograd graph. Added @torch.no_grad() anyway as a defensive measure since it's zero cost.

Only added it to get_model_save_state since it's the only function that directly operates on model tensors. The others either call it(_save_hf_safetensor_distributed), load from disk (_save_hf_safetensor_legacy), or just route (save_hf_safetensor) — none of them touch model parameters directly.

veomni/utils/save_safetensor_utils.py
```diff
@@ -12,40 +12,174 @@
...
+        fqn_to_index_mapping: [Distributed only] Maps FQNs to safetensors file indices
+            for multi-file output.
+    """
+    use_distributed = is_torch_version_greater_than("2.9") and train_architecture != "lora" and ckpt_manager == "dcp"
+
+    if use_distributed:
+        _save_hf_safetensor_distributed(model, save_hf_safetensor_path, fqn_to_index_mapping, model_assets)
+    else:
+        # Legacy path is rank-0 only
+        if dist.is_initialized() and dist.get_rank() != 0:
```
piyifan123, line 181: can we just pass in the is_rank0 as an argument?

tests/checkpoints/test_trainer_saveload.py
piyifan123: SGTM. please ensure that tests cover EP case and tied_weight case

veomni/utils/save_safetensor_utils.py
Ziyi-Wang: sure, resolved at #479

PR #463, opened 2026-02-06
[task] refactor: Centralize HF safetensor saving logic into shared util

Ziyi-Wang:
### What does this PR do?

Extract duplicated HF safetensor saving logic from 7 training scripts into a single shared utility `save_hf_safetensor()` in `veomni/utils/save_safetensor_utils.py`. This consolidates checkpoint-to-safetensors conversion, model asset saving, and LoRA weight filtering        
  (previously only in `train_flux.py` / `train_wan.py`) into one place.                                                                                                                                                                                                            

  **Files changed:**
  - **New:** `veomni/utils/save_safetensor_utils.py`
  - **Updated:** `tasks/train_torch.py`, `tasks/omni/train_qwen_vl.py`, `tasks/omni/train_qwen25_omni.py`, `tasks/omni/train_qwen3_omni.py`, `tasks/omni/train_omni_model.py`, `tasks/omni/train_flux.py`, `tasks/omni/train_wan.py` — replaced inline save blocks and removed local
  `save_hf_weights` helpers with calls to the shared utility.

### DIFF
--- a/tests/checkpoints/test_trainer_saveload.py
+++ b/tests/checkpoints/test_trainer_saveload.py
@@ -17,8 +17,6 @@ import torch
 import torch.distributed as dist
 from tqdm import trange
 
-from veomni.utils.save_safetensor_utils import save_hf_safetensor
-
 from veomni.arguments import DataArguments, ModelArguments, TrainingArguments, parse_args, save_args
 from veomni.checkpoint import build_checkpointer
 from veomni.data import build_dataloader, build_dummy_dataset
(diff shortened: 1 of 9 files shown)


tests/checkpoints/test_trainer_saveload.py
```diff
@@ -336,6 +337,24 @@ def main():
             dist.barrier()
             logger.info_rank0(f"Distributed checkpoint saved at {save_checkpoint_path} successfully!")
 
+    # After training on all epochs, save model in huggingface's format and verify against DCP checkpoint
+    if Checkpointer.dcp_save_future is not None:
+        Checkpointer.dcp_save_future.result()
+    if args.train.global_rank == 0 and args.train.save_hf_weights and save_checkpoint_path is not None:
```
piyifan123, line 343: args.train.save_hf_weights IIUC defaults to true. please check if it will trigger on all test cases.

Ziyi-Wang: Confirmed.
save_hf_weights defaults to True in TrainingArguments, but get_checkpoint_test_command in utils.py defaults its save_hf_weights param to False and always explicitly passes --train.save_hf_weights to the CLI.
So test_trainer_saveload  runs with False, and only test_trainer_save_hf_safetensor runs with True. 

The dataclass default is never implicitly used.

PR #464, opened 2026-02-06
[model] fix: fix weight init

FoolPlayer:
### What does this PR do?

fix weight init for moe models


Luosuu:
LGTM. Please fix CI.

PR #467, opened 2026-02-10
[model] fix: handle audio_seqlens=None in video branch

yiwzhao:
Now, we set audio_seqlens=None when the video sample doesn't have audio features at all. 
During training process, I noticed the video branch did not guard against this None case. And it caused a crash on any video-without-audio sample in LLaVA_Video dataset.
So I add a none check before access.


Luosuu (approved):

Coach257:
When training video, audio_seqlens must be given. If no audio, the audio_seqlens will be [0]. This is fixed in trainer.

PR #474, opened 2026-02-12
[model] chore: respect USE_LIGER_KERNEL flag in GPU model patch

Luosuu:
Now env var USE_LIGER_KERNEL is only used for cross entropy loss but we should respective this flag for general model patch as well, including RoPE, RMSNorm and SwiGLUMLP.

### DIFF
--- a/veomni/utils/env.py
+++ b/veomni/utils/env.py
@@ -8,7 +8,7 @@ logger = logging.get_logger(__name__)
 
 ENV_DEFAULTS = {
     "MODELING_BACKEND": "veomni",
-    "VEOMNI_VEOMNI_USE_LIGER_KERNEL": "1",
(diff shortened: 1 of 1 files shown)


veomni/utils/env.py
```diff
@@ -8,7 +8,7 @@
 
 ENV_DEFAULTS = {
     "MODELING_BACKEND": "veomni",
-    "USE_LIGER_KERNEL": "1",
+    "VEOMNI_VEOMNI_USE_LIGER_KERNEL": "1",
```
Coach257: VEOMNI_USE_LIGER_KERNEL？

Coach257 (changes requested):
fix key error

Coach257 (approved):

PR #479, opened 2026-02-13
[ckpt] feat: make save_hf_safetensor self-contained with internal synchronization

Ziyi-Wang:
## Summary

  - Refactor `save_hf_safetensor` to be self-contained w.r.t. synchronization: add `synchronize()` at entry, move async DCP save wait logic from `_save_hf_safetensor_distributed` to the top-level function, and add `dist.barrier()` at exit so callers no longer need external sync.
  - Add `is_rank_0` parameter to internalize the rank-0 guard for legacy save path, removing the need for callers to wrap the call in `if global_rank == 0`.
  - Update all training scripts (`train_flux`, `train_wan`, `train_omni_model`, `train_qwen25_omni`, `train_qwen3_omni`, `train_qwen_vl`, `train_torch`) and test to pass `is_rank_0` and remove outer rank checks.
  - Add `helper.empty_cache()` after distributed HF safetensors save to free GPU memory.

### DIFF
--- a/veomni/utils/save_safetensor_utils.py
+++ b/veomni/utils/save_safetensor_utils.py
@@ -100,7 +100,8 @@ def _save_hf_safetensor_distributed(
     save_state = get_model_save_state(model, fqn_to_index_mapping)
 
     logger.info_rank0("Starting distributed HuggingFace safetensors save...")
-    dist.barrier()
+    if dist.is_initialized():
+        dist.barrier()
     start_time = time.time()
     dcp.save(
         state_dict=save_state,
@@ -110,6 +111,7 @@ def _save_hf_safetensor_distributed(
     if dist.is_initialized():
         dist.barrier()
     gc.collect()
+    helper.empty_cache()
     elapsed_time = time.time() - start_time
     logger.info_rank0(f"Distributed HuggingFace safetensors save took {elapsed_time:.2f}s")
 


Luosuu:
please make PR title more informative

veomni/utils/save_safetensor_utils.py
```diff
@@ -187,8 +190,7 @@ def save_hf_safetensor(
     if use_distributed:
         _save_hf_safetensor_distributed(model, save_hf_safetensor_path, fqn_to_index_mapping, model_assets)
     else:
-        # Legacy path is rank-0 only
-        if dist.is_initialized() and dist.get_rank() != 0:
+        if not is_rank_0:
```
piyifan123: nit: add a comment saying the legacy save_hfsafetensor is done by rank0 only.

veomni/utils/save_safetensor_utils.py
piyifan123: can we make sure that every path here have synchorinize() and dist barrier()?

veomni/utils/save_safetensor_utils.py
Ziyi-Wang: Done. Added a comment clarifying that the legacy save_hfsafetensor is done by rank0 only.

veomni/utils/save_safetensor_utils.py
Ziyi-Wang: Yes, all paths now have synchronize() and dist.barrier() to ensure consistency across ranks.

PR #482, opened 2026-02-13
[model] feat: Support Qwen3-Next with additional support for Ulysses-based Linear Attention Sequence Parallelism

Luosuu:
### What does this PR do?

Support Qwen3-Next in VeOmni, with additional support for Ulysses-based Linear Attention Sequence Parallelism

### Test

- [x] unit model test.


Coach257:
maybe after trainer merged?

Luosuu:
> maybe after trainer merged?

Sure. I can just leave this PR here in case anyone needs it.

veomni/models/transformers/qwen3_next/modeling_qwen3_next.py
```diff
@@ -0,0 +1,1574 @@
+#                🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨
```
Zhang1Sheng, line 1: In the current code repository, all transformers models use `patch` to integrate new models. Should we avoid directly copying the entire file here?

Luosuu:
Since Qwen3.5 is supported. It seems Qwen3-Next is not very necessary anymore

PR #458, opened 2026-02-05
[BREAKING][data,model,task] refactor: use trainer to handle all tasks

Coach257:
### What does this PR do?
1. use trainer to handle all tasks (text, vlm, omni models) (DiT TODO...) #326 
- callbacks: functions that not related to training precision & efficiency. For example, environ meter, checkpoint, wandb.
- base trainer: build_models, build_parallelized_model, build_data, training loop
- vlm_trainer, text_trainer: build specific data transform functions
- rl_trainer: preforward & post forward
2. batching data strategy #436 #408 
- del args: rmpad, rmpad_with_position_ids
3. task & trainer & rl_trainer aligned
- origin tasks in `task/deprecated_tasks/`
- new trainer in `task/train_xxx.py`
- use task or trainer or rl_trainer all aligned on the same config.yaml
4. merge tasks
- merge qwen2vl, qwen25omni, qwen25vl, qwen3vl, qwe3vlmoe to task `train_qwen_vl.py` and trainer `train_vlm.py`
5. all tasks precompute position_ids, cu_seqlens, max_seqlens
6. position_ids sp slice in collate_fn or preforward
- no need to sp patch in modeling.py
7. cis:
- add e2e parallel align ci for text_models/vlm_models/omni_models. sp+ep align
- add model_patch ci for vlm_models/omni_models.
- ug models todo

PR #488, opened 2026-02-18
[data] feat: add DynamicBatchingSizeDataset for stateful multi-worker dynamic batching

LiuzcEECS:
### What does this PR do?

Add `DynamicBatchingSizeDataset`, a new dataset-level dynamic batching approach that constructs micro batches inside DataLoader worker processes (as opposed to the existing `DynamicBatchSizeDataLoader` which does it in the main process). This enables proper checkpoint/resume support via `StatefulDataLoader`'s `state_dict()`/`load_state_dict()` for dynamic batching, and is controlled by the new `dyn_bsz_in_worker_loop` flag.

### Test

All 9 tests pass (`pytest tests/data/test_dynamic_batching_dataset.py`):

- 5 unit tests: basic dynamic batching (shuffle on/off), force long sequence, last batch draining, without `get_item`
- 4 distributed tests via `torchrun --nproc_per_node=2`: shuffle × save_by_idx (2×2 combinations), each running 2 epochs of training with checkpoint save at step 2 of epoch 1 and resume verification

### API and Usage Example

New arguments:

```bash
# Use dataset-level dynamic batching (DynamicBatchingSizeDataset) instead of main-process batching (DynamicBatchSizeDataLoader)
--train.dyn_bsz_in_worker_loop=false

# Control whether to save buffer by index (way smaller checkpoint) or by full sample
--train.dyn_bsz_dataset_save_by_idx=true
```

When `dyn_bsz_in_worker_loop=false`, `build_native_dataloader` wraps the dataset with `DynamicBatchingSizeDataset`, which yields pre-batched micro batches from each worker.

When `dyn_bsz_in_worker_loop=true` (default, existing behavior), the original `DynamicBatchSizeDataLoader` path is used — no behavior change.

### Design & Code Changes

**Core: `DynamicBatchingSizeDataset`** (`veomni/data/dynamic_batching.py`)
- New `IterableDataset` subclass that buffers samples in each worker and yields micro batches when the buffer has ≥ `ready_for_micro_batch_threshold` samples and ≥ `micro_batch_seq_length` total tokens.
- Greedy bin-packing in `_get_micro_batch()`: iterates buffer, selects samples fitting within the token budget.
- Supports `state_dict()` / `load_state_dict()` for `StatefulDataLoader` checkpoint/resume:
  - `save_by_idx=True`: saves only sample indices (smaller checkpoint), requires dataset to support `get_item()` and `output_refetch_idx`.
  - `save_by_idx=False`: saves full buffer contents via `deepcopy`.

**Integration** (`veomni/data/data_loader.py`)
- `build_native_dataloader` now branches on `dyn_bsz_in_worker_loop`:
  - `True`: existing `TextBatchingStrategy` + `DynamicBatchSizeDataLoader` path.
  - `False`: wraps dataset in `DynamicBatchingSizeDataset`, uses `NoopDataCollator`, returns `StatefulDataLoader` directly (no `DynamicBatchSizeDataLoader` wrapper).

**Arguments** (`veomni/arguments/arguments_types.py`)
- `dyn_bsz_in_worker_loop: bool = True` — controls which dynamic batching path to use.
- `dyn_bsz_dataset_save_by_idx: bool = True` — controls checkpoint buffer serialization strategy.
- Updated `dataloader_batch_size` calculation for `dyn_bsz_in_worker_loop=False` case.

**Minor refactor** (`veomni/data/batching_strategy.py`)
- Renamed `BaseBatchingStrategy.is_full_filled()` → `is_ready_for_micro_batch()` for clarity.

**Tests** (`tests/data/test_dynamic_batching_dataset.py`, `tests/data/utils.py`)
- `DummyMappingDataset` / `DummyIterableDataset`: test fixtures with configurable sequence lengths, rank/worker sharding (like `StreamingDataset` in ByteDance), and `state_dict()` support.
- Unit tests covering basic batching, long sequence handling, dataset exhaustion.
- Distributed tests running multi-epoch training with checkpoint save/resume verification across shuffle × save_by_idx combinations.

### DIFF
--- a/tests/data/test_dynamic_batching_dataset.py
+++ b/tests/data/test_dynamic_batching_dataset.py
@@ -16,8 +16,6 @@ The test suite includes:
     Unit tests (run without distributed setup, CPU-compatible):
         - ``test_dynamic_batching_basic`` – core batching logic and expected batch
           contents for shuffled and non-shuffled data.
-        - ``test_force_long_sequence`` – overlong samples are emitted rather than
-          dropped when ``force_generate_long_sequence=True``.
         - ``test_last_batch_on_dataset_end`` – remaining buffer items are yielded
           after upstream exhaustion.
         - ``test_dynamic_batching_without_get_item`` – ``ValueError`` is raised when
@@ -152,14 +150,13 @@ def test_dynamic_batching_basic(shuffle, seed):
     ]
 
     # Expected input_ids for shuffle=True (seed=42)
-    # Note: force_generate_long_sequence=True allows batches to exceed micro_batch_seq_length
-    # when the buffer is empty and only has one long sample
+    # Samples longer than MICRO_BATCH_SEQ_LENGTH (32) are skipped (force_generate_long_sequence=False).
     expected_input_ids_shuffle = [
-        [43] * 43,  # 43 tokens (exceeds 32 due to force_generate_long_sequence)
-        [18] * 18 + [1] + [9] * 9,  # 28 tokens
+        [18] * 18 + [1] + [9] * 9 + [4] * 4,  # 32 tokens
         [31] * 31,  # 31 tokens
-        [35] * 35,  # 35 tokens (exceeds 32 due to force_generate_long_sequence)
-        [21] * 21 + [4] * 4 + [3] * 3 + [2] * 2,  # 30 tokens
+        [21] * 21 + [3] * 3 + [2] * 2,  # 26 tokens
+        [26] * 26 + [6] * 6,  # 32 tokens
+        [19] * 19 + [7] * 7,  # 26 tokens
     ]
 
     expected_input_ids = expected_input_ids_shuffle if shuffle else expected_input_ids_no_shuffle
@@ -172,8 +169,6 @@ def test_dynamic_batching_basic(shuffle, seed):
         assert batch["attention_mask"].sum() > 0
 
         # Calculate total tokens in batch
-        # Note: With force_generate_long_sequence=True, batches can exceed micro_batch_seq_length
-        # when the buffer only has one long sample
         total_tokens = batch["attention_mask"].sum()
 
         # Check expected input_ids
@@ -199,49 +194,6 @@ def test_dynamic_batching_basic(shuffle, seed):
 
     assert batch_count > 0, "Should produce at least one batch"
     logger.info(f"test_dynamic_batching_basic (shuffle={shuffle}) passed! Produced {batch_count} batches")
-
-
-def test_force_long_sequence():
-    """Test that force_generate_long_sequence allows batches exceeding micro_batch_seq_length.
-
-    When force_generate_long_sequence=True and the buffer only has one long sample,
-    the batch should be allowed to exceed micro_batch_seq_length.
-    """
-
-    mapping_dataset = DummyMappingDataset(size=DATASET_SIZE)
-    iterable_dataset = DummyIterableDataset(mapping_dataset, shuffle=False)
-
-    # Test with force_generate_long_sequence=True (default)
-    dynamic_ds = DynamicBatchingSizeDataset(
-        dataset=iterable_dataset,
-        micro_batch_seq_length=MICRO_BATCH_SEQ_LENGTH,
-        ready_for_micro_batch_threshold=READY_FOR_MICRO_BATCH_THRESHOLD,
-        dynamic_batching_collate_fn=DataCollatorWithPositionIDs(),
-        get_length_fn=get_length_fn,  # Use the global get_length_fn that works with dict
-        force_generate_long_sequence=True,
-    )
-
-    # Iterate through batches to find one that exceeds micro_batch_seq_length
-    found_long_batch_value = None
-
-    batch_idx = 0
-    for batch in dynamic_ds:
-        total_tokens = batch["attention_mask"].sum().item()
-        input_ids = batch["input_ids"].tolist()[0]
-        unique_values = set(input_ids)
-        is_single_sample = len(unique_values) == 1
-
-        if total_tokens > MICRO_BATCH_SEQ_LENGTH and is_single_sample:
-            sample_value = unique_values.pop()
-            found_long_batch_value = sample_value
-            break
-
-        batch_idx += 1
-
-    # Verify the found batch
-    assert found_long_batch_value == MICRO_BATCH_SEQ_LENGTH + 1, (
-        f"Expected long batch to contain value {MICRO_BATCH_SEQ_LENGTH + 1}, got value={found_long_batch_value}"
-    )
 
 
 def test_last_batch_on_dataset_end(setup_dynamic_batching_dataset):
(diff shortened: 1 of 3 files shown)


Luosuu:
should we simply replace the original one in this case? Why would we need two if their difference is just where to perform batching

veomni/arguments/arguments_types.py
```diff
@@ -420,10 +396,22 @@ class TrainingArguments:
         default="worker",
         metadata={"help": "Use main process or worker process to run dynamic batch size."},
     )
+    dyn_bsz_in_worker_loop: bool = field(
```
Luosuu: remove this. After integration with trainer we can expose this to end users

LiuzcEECS: sounds good, let's discuss the APIs when we are reviewing https://github.com/ByteDance-Seed/VeOmni/pull/458

LiuzcEECS: I created an issue to track the status https://github.com/ByteDance-Seed/VeOmni/issues/491

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,283 @@
...
+
+        Yields:
+            Collated micro batch when buffer conditions are met.
+
+        Raises:
+            Exception: Re-raises any exception other than StopIteration encountered
+                during iteration.
+        """
+        if not self._data_iter:
+            self._data_iter = iter(self.dataset)
```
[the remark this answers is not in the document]
LiuzcEECS: I think you are wrong, in the pytorch dataloader, when __iter__ is started, it will copy the state of the dataset inited in the main process, in the main process, we only init it and the _data_iter is always None in the main process. Then the __iter__ func will try to assign value to it in the beginning of looping so we should be fine.

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,309 @@
...
+
+        while True:
+            try:
+                item = next(self._data_iter)
```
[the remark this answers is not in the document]
LiuzcEECS: @gemini-code-assist I think you are wrong, in the pytorch dataloader, when iter is started, it will copy the state of the dataset inited in the main process, in the main process, we only init the dataloader and dataset itself but not the iter and the _data_iter is always None in the main process. Then the iter func will try to assign value to it in the beginning of looping so we should be fine.

tests/data/test_dynamic_batching_dataset.py
```diff
@@ -0,0 +1,665 @@
...
+import argparse
+import os
+import subprocess
+import sys
+import time
+from dataclasses import dataclass, field
+from unittest.mock import patch
+
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
```
piyifan123, line 39: nit: double check to see if we need this.

veomni/data/data_loader.py
```diff
@@ -122,14 +125,26 @@ def build_native_dataloader(
...
     if use_rmpad and dyn_bsz:
-        batching_strategy = TextBatchingStrategy(
-            token_micro_bsz=token_micro_bsz - dyn_bsz_margin * max_seq_len,
-            buffer_size=dyn_bsz_buffer_size,
-            bsz_warmup_steps=bsz_warmup_steps if bsz_warmup_steps else -1,
-            bsz_warmup_init_mbtoken=bsz_warmup_init_mbtoken,
-        )
         dyn_bsz_collate_fn = collate_fn
-        collate_fn = UnpackDataCollator()
+        if dyn_bsz_in_dataloader:
```
piyifan123, line 130: nit: add docstring to what  dyn_bsz_in_dataloader means.

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,309 @@
...
+    performs batching inside each DataLoader worker process.
+    It is also compatible with ``StatefulDataLoader``'s per-worker ``state_dict()`` /
+    ``load_state_dict()`` mechanism, enabling exact checkpoint / resume for dynamic-batching workloads.
+
+    Internally each worker maintains a sample buffer.  A micro batch is emitted once
+    the buffer holds at least ``ready_for_micro_batch_threshold`` samples **and** their
+    combined token count reaches ``micro_batch_seq_length``.  When the upstream dataset
+    is exhausted, remaining buffer contents are drained and emitted as final batches
+    regardless of the threshold.
+
```
piyifan123, line 47: add a doc under docs to explain DynamicBatchingSizeDataset

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,309 @@
...
+    def __init__(
+        self,
+        dataset: IterableDataset,
+        micro_batch_seq_length: int,
+        ready_for_micro_batch_threshold: int,
+        dynamic_batching_collate_fn: Optional[Callable] = DataCollatorWithPositionIDs(),
+        save_by_idx: bool = True,
+        get_length_fn: Optional[Callable] = len,
+        force_generate_long_sequence: bool = True,
```
piyifan123: can we default this to False? this is a major cause of OOM.

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,309 @@
...
+
+        while True:
+            try:
+                item = next(self._data_iter)
+
+                if self.save_by_idx:
+                    item, refetch_idx = item[0], item[1]
+
+                length = self.get_length_fn(item)
+                if length > self.micro_batch_seq_length and not self.force_generate_long_sequence:
```
piyifan123, line 156: can we simplify to support only force_generate_long_sequence=False.

piyifan123: add a TODO to also record num of thrown too long examples.

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,309 @@
...
+        """
+        state = {
+            "save_by_idx": self.save_by_idx,
+            "buffer_token_count": self._buffer_token_count,
+        }
+
+        # the state_dict might be called frequently with StatefulDataloaders(see more details of snapshot_every_n_steps)
+        # so we try to not include extra calculations here.
+        if self.save_by_idx:
+            state["buffer"] = copy.copy(self._buffer_of_refetch_idx)
```
piyifan123: can we deepcopy to align?

veomni/data/dynamic_batching.py
```diff
@@ -17,15 +17,309 @@
...
+                sum of token lengths recomputed from the reconstructed buffer.
+            ValueError: If ``save_by_idx`` is True on the current instance but the
+                checkpoint buffer holds some full samples instead of indices (incompatible
+                checkpoint format).
+        """
+        # prev_save_by_idx does not have to be equal to self.save_by_idx, however, we still need to resume the buffer according to it.
+        prev_save_by_idx = state_dict["save_by_idx"]
+        if prev_save_by_idx:
+            self._buffer = []
+            self._buffer_of_refetch_idx = []
```
piyifan123, line 291: can we log the time and progress of the replaying?

tests/data/test_dynamic_batching_dataset.py
```diff
@@ -0,0 +1,665 @@
...
+            break
+
+        batch_idx += 1
+
+    # Verify the found batch
+    assert found_long_batch_value == MICRO_BATCH_SEQ_LENGTH + 1, (
+        f"Expected long batch to contain value {MICRO_BATCH_SEQ_LENGTH + 1}, got value={found_long_batch_value}"
+    )
+
+
```
piyifan123: can we also add a unit test to test save/load?

piyifan123 (approved):

tests/data/test_dynamic_batching_dataset.py
LiuzcEECS: yes, we do. Otherwise if we use `from ..tools.launch_utils import find_free_port` when using torchrun to call that, we will get `ImportError: attempted relative import with no known parent package`

veomni/data/dynamic_batching.py
LiuzcEECS: Yes, it will be part of our work in future for metrics in wandb

veomni/data/dynamic_batching.py
LiuzcEECS: no problem, just FYI, in byted-streaming, even setting this to False will still cause OOM because their codes don't clean up the memory for too long samples

PR #493, opened 2026-02-19
[omni, model] fix: support mixed video_w_audio & video_w/o_audio in Qwen3OmniMoe get_rope_index

TimYangst:
### What does this PR do?

> Fix three bugs in `Qwen3OmniMoeThinker` and `Qwen3OmniMoePreTrainedModel`:
> 1. `get_rope_index`: support batches with mixed videos-with-audio and videos-without-audio, mirroring the per-video check in `modeling_qwen2_5_omni.py`.
> 2. `get_audio_features`: fix `audio_feature_lengths` being overwritten with `None` in the `else` branch, causing `AttributeError` at runtime.
> 3. `forward`: replace the call to the removed `get_placeholder_mask` with the pre-computed `audio_mask` from `kwargs`, consistent with how `image_mask` and `video_mask` are already handled.

### Checklist Before Starting

- [ ] Format the PR title as `[{modules}] {type}: {description}` (This will be checked by the CI)
  - `{modules}` include `misc`, `ci`, `config`, `docs`, `core`, `data`, `dist`, `omni`, `logging`, `model`, `optim`, `ckpt`, `release`, `task`
  - If this PR involves multiple modules, separate them with `,` like `[megatron, torch, liger-kernals,fsdp, doc]`
  - `{type}` is in `feat`, `fix`, `refactor`, `chore`, `test`
  - If this PR breaks any API (CLI arguments, config, function signature, etc.), add `[BREAKING]` to the beginning of the title.
  - Example: `[BREAKING][fsdp, torch, liger-kernals] feat: dynamic batching`

### Test

> Correctness validated by code inspection against `modeling_qwen2_5_omni.py` and the existing `image_mask`/`video_mask` pattern in this file. Full training run CI required to validate end-to-end.

### API and Usage Example

No API change. All fixes are internal to `modeling_qwen3_omni_moe.py`. Callers must ensure `audio_mask` is included in the kwargs passed to `forward` (alongside the existing `image_mask` and `video_mask`), which is already handled by `vlm_data_process.py`.

### Design & Code Changes

**Fix 1 — `get_rope_index`: per-video audio check** (`~L417`)

Replaced the global `use_audio_in_video` flag with a per-video check inside the "Video Only" branch, using `audio_seqlens[audio_idx]` (not `video_idx`) to correctly handle samples where pure-audio tokens precede videos:

```python
if audio_seqlens is not None:
    if audio_seqlens[audio_idx] == 0:
        use_audio_in_video = False
        audio_idx += 1  # consume zero-length placeholder
    else:
        use_audio_in_video = True
else:
    use_audio_in_video = False
```

Also simplified the interleaving logic (merge-sort loop → `torch.cat` + `torch.argsort`) in both the new and existing video-with-audio branches.

**Fix 2 — `get_audio_features`: preserve caller-supplied `audio_feature_lengths`** (`~L2111`)

Removed the `else: audio_feature_lengths = None` branch that overwrote the argument passed by the caller, causing an `AttributeError` on line 2117 when `feature_attention_mask` was absent.

**Fix 3 — `forward`: use pre-computed `audio_mask`** (`~L2258`)

`get_placeholder_mask` was commented out but still called for audio. Replaced with `audio_mask` from `kwargs` (same pattern as `image_mask`/`video_mask`), and added the corresponding `assert` guard.

> [!IMPORTANT]

- [ ] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/ByteDance-Seed/VeOmni/tree/main/.github/workflows) to cover all the code. If not feasible, explain why: Bugs only manifest at runtime with specific audio/video combinations; a targeted unit test for `get_rope_index` and `get_audio_features` with mixed inputs is recommended.

### DIFF
--- a/veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
+++ b/veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
@@ -455,7 +455,7 @@ class Qwen3OmniMoePreTrainedModelForConditionalGeneration(Qwen3OmniMoePreTrained
                             video_idx += 1
                             remain_videos -= 1
                         else:
-                            # Video with audio track: interleave video and audio position ids
+                            # --- Patch: Video with audio track: interleave video and audio position ids
                             audio_len = _get_feat_extract_output_lengths(audio_seqlens[audio_idx])
                             audio_llm_pos_ids = torch.arange(audio_len).view(1, -1).expand(3, -1) + st_idx
                             grid_t = video_grid_thw[video_idx][0]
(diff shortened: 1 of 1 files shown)


veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
```diff
@@ -417,25 +417,70 @@ def get_rope_index(
...
+                        # video will have already advanced audio_idx independently of video_idx.
+                        if audio_seqlens is not None:
+                            if audio_seqlens[audio_idx] == 0:
+                                use_audio_in_video = False
+                                audio_idx += 1  # consume the zero-length placeholder ONLY when there is no audio
+                            else:
+                                use_audio_in_video = True
+                        else:
+                            use_audio_in_video = False
+                        # --- Patch end ---
```
piyifan123, line 437: nit: the annotation here.

veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
```diff
@@ -417,25 +417,70 @@ def get_rope_index(
...
+
+                        if not use_audio_in_video:
+                            grid_t = video_grid_thw[video_idx][0]
+                            grid_hs = video_grid_thw[:, 1]
+                            grid_ws = video_grid_thw[:, 2]
+                            t_index = (
+                                torch.arange(grid_t)
+                                * second_per_grids[video_idx].cpu().float()
```
piyifan123, line 445: why .cpu() here?

TimYangst: It is necessory, otherwise there are maybe device mismatch issue (second_per_grids may be initialized in gpu)

veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py
TimYangst: Added a new patch block blow

PR #504, opened 2026-02-24
[misc] fix: optimize npu profiling uploading

yicheng-gong:
fix: optimize npu profiling uploading

### DIFF
--- a/veomni/utils/helper.py
+++ b/veomni/utils/helper.py
@@ -664,7 +664,16 @@ def create_profiler(
         if VEOMNI_UPLOAD_CMD:
             try:
                 logger.info_rank0(f"upload trace file {trace_file}")
-                command2 = f"{VEOMNI_UPLOAD_CMD} {trace_file}"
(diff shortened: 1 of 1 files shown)


veomni/utils/helper.py
```diff
@@ -664,7 +664,16 @@ def handler_fn(p):
         if VEOMNI_UPLOAD_CMD:
             try:
                 logger.info_rank0(f"upload trace file {trace_file}")
-                command2 = f"{VEOMNI_UPLOAD_CMD} {trace_file}"
+                if IS_NPU_AVAILABLE:
```
Luosuu, line 667: I think such utils should only be patched internally?

heidongxianhua: 这里本来就是个try逻辑，似乎不用单独区分？

Luosuu (approved):

veomni/utils/helper.py
Luosuu: ok

PR #505, opened 2026-02-24
[parallel] feat: Vision Data Parallel — O(1) communication alternative to patch-level SP

aoshen524:
## Summary

This PR introduces **Vision DP** — an alternative to patch-level Sequence Parallelism for Vision Transformers that trades per-layer communication for a simpler, communication-free ViT execution with a single post-ViT all-gather.

### Why this is NOT traditional Data Parallelism

> **Key distinction**: Vision DP operates across **SP ranks** (which share the same micro-batch), NOT DP ranks (which have different data).

In Ulysses SP, all SP ranks must hold the **identical text sequence** for the sequence-parallel attention to work correctly. Since vision embeddings are embedded inline within the text sequence, all SP ranks also receive **all images** in the batch. This creates a unique problem:

| | Traditional DP | Vision DP (this PR) |
|--|----------------|---------------------|
| **Ranks** | DP ranks (different data) | SP ranks (same data) |
| **Input** | Each rank has different micro-batch | All ranks have the **same** micro-batch |
| **Fix via dataloader?** | Yes — dataloader shards data | **No** — SP ranks must share the same input |
| **What's distributed** | Entire forward/backward | Only the ViT forward; embeddings are all-gathered back |

**Why you can't just change the dataloader**: SP ranks are constrained to process the same sequence. Giving them different images would break text-side sequence parallelism — the attention all-to-all would receive mismatched KV from different ranks.

### How the two approaches compare

Both patch-level SP (current) and Vision DP achieve **the same ~1/sp_size ViT activation memory reduction** — they just partition along different axes:

| | Patch-level SP (current) | Vision DP (this PR) |
|--|--------------------------|---------------------|
| **Partition axis** | Each rank processes 1/sp_size patches of ALL images | Each rank processes ALL patches of 1/sp_size images |
| **ViT activation memory** | ~total_patches/sp_size per rank | ~total_patches/sp_size per rank |
| **ViT internal comms** | Model-specific (see below) | **Zero** — each rank runs ViT independently |
| **Post-ViT** | — | 1 all-gather to collect all embeddings |

The key difference is in **ViT-internal communication**:

#### Qwen2.5-VL: Vision DP eliminates 4 boundary all-to-alls

Qwen2.5-VL's ViT uses **window attention**, which requires seeing the full sequence to compute `window_index` / `reverse_indices`. With patch-level SP, this forces 4 boundary all-to-all operations (gather→reindex→scatter, before and after the transformer block loop).

| | Patch-level SP | Vision DP |
|--|----------------|-----------|
| ViT internal | 4 all-to-all (window attention boundary gather/scatter) | 0 |
| Post-ViT | — | 1 all-gather |
| Model fusion | 3–4 all-to-all | similar |
| **Net ViT comms** | **4 all-to-all → 1 all-gather** |

#### Qwen3-VL: Comparable — current approach already has zero ViT-internal comms

Qwen3-VL has **no window attention**, so patch-level SP already achieves zero ViT-internal all-to-all. Vision DP adds 1 all-gather but may simplify model fusion.

| | Patch-level SP | Vision DP |
|--|----------------|-----------|
| ViT internal | 0 | 0 |
| Post-ViT | — | 1 all-gather |
| Model fusion | 3 all-to-all + N×all-gather (deepstack) | similar |
| **Net ViT comms** | **Comparable — no clear winner** |

### When to use Vision DP

- **Qwen2.5-VL**: Recommended — eliminates 4 window attention boundary all-to-alls, simpler ViT implementation
- **Qwen3-VL**: Optional — similar communication cost, but simpler ViT code (no `cu_seqlens` SP slicing or `sp_pad_and_slice` inside ViT)
- **General**: Useful when you want a model-agnostic ViT SP solution that doesn't require per-model attention patching

### Key design choices

- **Image-level distribution**: avoids breaking ViT's internal `cu_seqlens` tracking — no need to manipulate attention boundaries across SP ranks
- **Load-balanced contiguous bin-packing**: assigns images to ranks by patch count, preserving order so no reordering is needed after gather
- **Correct gradient routing**: `all_reduce(SUM)` in backward before slicing recovers complete gradients from all SP ranks' sequence shards
- **Backward-compatible**: controlled via `vision_dp=True` flag. Default behavior is unchanged.

### Changes

- **New**: `veomni/distributed/sequence_parallel/vision_dp.py` — core utilities (assign, prepare, gather, `create_dp_vision_forward` wrapper)
- **Modified**: `modeling_qwen3_vl.py` — `_vision_dp` flag to skip SP patches; `apply_vision_dp_patch()`; outer model skips post-ViT `gather_seq_scatter_heads` when Vision DP is active
- **Modified**: `modeling_qwen2_5_vl.py` — same pattern, eliminating 4 all-to-all ops in ViT
- **Modified**: `veomni/arguments/arguments_types.py` — added `vision_dp: bool = False` to `TrainingArguments`
- **Modified**: `veomni/models/auto.py` — `build_foundation_model()` accepts `vision_dp` and dispatches to model-specific patch
- **Modified**: `tasks/train_torch.py`, `tasks/omni/train_qwen_vl.py` — pass `vision_dp` from training args
- **New**: 21 CPU-only unit tests for all Vision DP utilities

### Usage

```bash
# Enable via training argument
--train.vision_dp=true --train.ulysses_parallel_size=2
```

### Precision Alignment: Vision DP On vs Off (verl reference experiment)

Precision alignment was validated in the verl framework (volcengine/verl#5230) under controlled conditions. The same Vision DP algorithm is shared across frameworks.

#### Experiment Setup

| Config | Value |
|--------|-------|
| Model | Qwen2.5-VL-7B (SFT checkpoint) |
| GPUs | 4 × H100 (1 node) |
| SP size | 2 |
| Strategy | FSDP2 |
| Train batch size | 4 |
| Rollout N | 2 |
| Max assistant turns | 4 |
| Dummy mode | full (fixed multi-image prompts with meaningful screenshots + fixed text) |
| Determinism | enabled (seed=42, TF32 disabled, CUBLAS_WORKSPACE_CONFIG=:4096:8) |
| Advantage | hardcoded to 1.0 to remove reward randomness |
| Training steps | 1 |

#### Results: Hash-Level (62 groups × 3 phases)

| Phase | Param SHA256 | Grad SHA256 Match | Vision Grad MM | Language Grad MM |
|-------|:------------:|:-----------------:|:--------------:|:----------------:|
| **pre_clip** (raw grad) | 62/62 ✅ | 29/62 | 32 | **0** ✅ |
| **before** (post-clip) | 62/62 ✅ | 0/62 | 32 | 29 |
| **after** (post-step) | 0/62 | 0/62 | 32 | 29 |

> **Language gradients are bitwise identical at pre_clip** — Vision DP does not affect the language model's gradient computation. The `before`/`after` phase divergence is expected: global grad norm (used for clipping) includes vision gradients, so a different clip factor propagates to all parameters.

#### Results: Per-Parameter Element-wise Diff

| Scope | Params | max_diff | mean_diff | cosine_sim |
|-------|-------:|:--------:|:---------:|:----------:|
| **vision** | 390 | 4.70e-05 | 2.93e-08 | 0.9991 |
| **language** | 338 | 9.50e-08 | 1.15e-10 | 1.0020 |
| **other** | 1 | 9.13e-08 | 2.25e-13 | 1.0001 |

#### Conclusion

- **Vision DP is numerically lossless**: all differences fall within bf16 floating-point precision (~1e-05 max).
- **Root cause**: `all_reduce(SUM)` in the backward pass changes floating-point accumulation order across SP ranks (SP=2), causing ~1e-05 level differences in vision gradients only.
- **Language is unaffected**: pre-clip language gradients are bitwise identical; post-clip differences are solely due to global grad norm clipping propagation.

## Test plan

- [x] 21 CPU-only unit tests pass (`pytest tests/parallel/test_vision_dp.py -v`)
- [x] `ruff check` and `ruff format` pass on all modified files
- [ ] Multi-GPU distributed test with SP enabled (requires GPU cluster)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tasks/omni/train_qwen_vl.py
+++ b/tasks/omni/train_qwen_vl.py
@@ -129,6 +129,7 @@ def main():
         attn_implementation=args.model.attn_implementation,
         encoder_data_balance=args.model.encoder_data_balance,
         encoder_data_balance_sorting_algo=args.model.encoder_data_balance_sorting_algo,
+        vision_dp=args.train.vision_dp,
     )
     model_config = model.config
     helper.print_device_mem_info("VRAM usage after building model")
--- a/tasks/train_torch.py
+++ b/tasks/train_torch.py
@@ -150,6 +150,7 @@ def main():
         attn_implementation=args.model.attn_implementation,
         moe_implementation=args.model.moe_implementation,
         init_device=args.train.init_device,
+        vision_dp=args.train.vision_dp,
     )
     model_config = model.config
     helper.print_device_mem_info("VRAM usage after building model")
--- a/veomni/arguments/arguments_types.py
+++ b/veomni/arguments/arguments_types.py
@@ -554,6 +554,10 @@ class TrainingArguments:
         default=1,
         metadata={"help": "Ulysses sequence parallel size."},
     )
+    vision_dp: bool = field(
+        default=False,
+        metadata={"help": "Enable Vision DP: distribute ViT across Ulysses SP ranks."},
+    )
     context_parallel_size: int = field(
         default=1,
         metadata={"help": "Ring-attn context parallel size."},
--- a/veomni/models/auto.py
+++ b/veomni/models/auto.py
@@ -80,6 +80,7 @@ def build_foundation_model(
     config_kwargs: Optional[Dict[str, Any]] = None,
     encoder_data_balance: Optional[bool] = False,
     encoder_data_balance_sorting_algo: Optional[str] = "post_mbs_balancing_greedy_without_pad",
+    vision_dp: bool = False,
 ) -> "PreTrainedModel":
     """
     Builds the foundation model.
@@ -185,6 +186,20 @@ def build_foundation_model(
             "Still evaluating HF kernels hub integration with VeOmni patches; keep use_kernels disabled for now "
             "to avoid unexpected kernel loading side effects."
         )
+
+    if vision_dp and get_parallel_state().sp_enabled:
+        if config.model_type == "qwen2_5_vl":
+            from .transformers.qwen2_5vl.modeling_qwen2_5_vl import apply_vision_dp_patch_qwen25
+
+            apply_vision_dp_patch_qwen25()
+            logger.info_rank0("Applied Vision DP patch for Qwen2.5-VL")
+        elif config.model_type == "qwen3_vl":
+            from .transformers.qwen3_vl.modeling_qwen3_vl import apply_vision_dp_patch
+
+            apply_vision_dp_patch()
+            logger.info_rank0("Applied Vision DP patch for Qwen3-VL")
+        else:
+            logger.warning_rank0(f"Vision DP requested but model_type={config.model_type} is not supported")
 
     model_class_path = f"{model.__class__.__module__}.{model.__class__.__name__}"
     logger.info_rank0(f"Built foundation model class: {model_class_path}")


veomni/models/transformers/qwen2_5vl/modeling_qwen2_5_vl.py
```diff
@@ -157,10 +158,16 @@ def Qwen2_5_VisionTransformerPretrainedModel_forward(
             The final hidden states of the model.
         grid_thw (`torch.Tensor` of shape `(num_images_or_videos, 3)`):
             The temporal, height and width of feature shape of each image in LLM.
+        _vision_dp: When True, skip patch-level SP logic (Vision DP handles
```
Luosuu, line 161: what if there are not enough images to distribute among vit dp ranks?

PR #515, opened 2026-02-26
[misc] chore: check if ckpt.load_path starts with ckpt.output_dir in post_init

xzzWZY:
In class TrainingArguments, check if "load_checkpoint_path" starts with "output_dir", forcing VeOmni to load checkpoints from output_dir (previous saved checkpoint).

### DIFF
--- a/veomni/arguments/arguments_types.py
+++ b/veomni/arguments/arguments_types.py
@@ -16,6 +16,7 @@ import json
 import math
 import os
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import Dict, List, Literal, Optional
 
 from ..utils import logging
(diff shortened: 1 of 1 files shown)


veomni/arguments/arguments_types.py
```diff
@@ -747,6 +748,17 @@ def __post_init__(self):
                 output_dir=self.output_dir, is_local_rank0=self.local_rank == 0, ckpt_manager=self.ckpt_manager
             )
 
+        if self.load_checkpoint_path:
+            load_path = Path(self.load_checkpoint_path).resolve()
+            output_dir = Path(self.output_dir).resolve()
+
+            try:
+                load_path.relative_to(output_dir)
```
Luosuu, line 539: please remind me why we have to do this?

PR #519, opened 2026-02-26
[ci] refactor: enable ruff 'B' rule and fix code quality issues

LiuzcEECS:
### Summary

Enable ruff's `B` (flake8-bugbear) linting rule to catch common Python anti-patterns and fix all violations across the codebase.

### Changes

**Linting Configuration**
- Add `"B"` to ruff's `select` rules in `pyproject.toml`

**Code Quality Fixes** (51 files, 229 insertions, 131 deletions)

The following flake8-bugbear violations were fixed:

1. **B006 - Mutable default arguments**
   - Replace `defaultdict(list)` defaults with `None` and initialize inside functions
   
2. **B024/B027 - Abstract base class issues**
   - Add `@abstractmethod` decorators to `BaseModelLoader.__init__` and `load_model`
   - Remove unnecessary `ABC` inheritance from utility classes

3. **B904 - Exception chaining**
   - Use `raise ... from e` for proper exception context
   
4. **B905 - zip() without strict parameter**
   - Add explicit `strict=False` where length mismatch is expected
   
5. **B009/B010 - getattr/setattr anti-patterns**
   - Replace `getattr(model, "module")` with `model.module`
   
6. **B007 - Unused loop variables**
   - Rename to `_variable` convention

### DIFF
--- a/tests/data/multimodal/test_video_utils.py
+++ b/tests/data/multimodal/test_video_utils.py
@@ -275,7 +275,7 @@ def test_fetch_videos_multiple_inputs():
     assert len(audios) == 3, f"Expected 3 audios, got {len(audios)}"
 
     # Verify each output
-    for i, (video, audio) in enumerate(zip(videos, audios)):
+    for _i, (video, audio) in enumerate(zip(videos, audios)):
         assert_video_output_valid(video, audio, **kwargs)
 
 
--- a/tests/ops/test_comp.py
+++ b/tests/ops/test_comp.py
@@ -30,7 +30,7 @@ def fast_pos_embed_interpolate_ref(self, grid_thw):
 
     device = self.pos_embed.weight.device
     dtype = self.pos_embed.weight.dtype
-    for t, h, w in zip(grid_ts, grid_hs, grid_ws):
+    for _t, h, w in zip(grid_ts, grid_hs, grid_ws, strict=True):
         h_idxs = torch.linspace(0, self.num_grid_per_side - 1, h, dtype=torch.float64)
         w_idxs = torch.linspace(0, self.num_grid_per_side - 1, w, dtype=torch.float64)
 
--- a/veomni/data/multimodal/multimodal_chat_template.py
+++ b/veomni/data/multimodal/multimodal_chat_template.py
@@ -32,7 +32,7 @@ logger = logging.get_logger(__name__)
 class MultimodalChatTemplate(ChatTemplate):
     @abstractmethod
     def encode_messages(
-        self, messages: Sequence[Dict[str, str]], num_tokens: Dict[str, List[int]] = defaultdict(list), **kwargs
+        self, messages: Sequence[Dict[str, str]], num_tokens: Dict[str, List[int]] = None, **kwargs
     ) -> Dict[str, List[int]]:
         """
         Encodes messages to a dictionary of input_ids, attention_mask, labels, and mm with mm_seqlens.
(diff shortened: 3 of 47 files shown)


tests/data/multimodal/test_video_utils.py
```diff
@@ -274,7 +274,7 @@ def test_fetch_videos_multiple_inputs():
     assert len(audios) == 3, f"Expected 3 audios, got {len(audios)}"
 
     # Verify each output
-    for i, (video, audio) in enumerate(zip(videos, audios)):
+    for _i, (video, audio) in enumerate(zip(videos, audios, strict=False)):
```
piyifan123: can we let AI to instead exempt with noqa for strict check? by reading the doc, it seems that the recommended fix is to change to True, but change to True in this PR is too large and risky of breaking things, let's use noqa to mark instead.

veomni/data/multimodal/video_utils.py
```diff
@@ -367,7 +369,7 @@ def _load_and_process_video_with_codec(video_input: VideoInput, use_audio_in_vid
                 sampled_indices = [0]
                 _, pad_count = calculate_frame_indices(total_frames=1, video_fps=video_fps, **kwargs)
             except Exception:
-                raise RuntimeError(f"Failed to decode even the first frame: {e}")
+                raise RuntimeError(f"Failed to decode even the first frame: {e}") from e
         else:
             raise e
```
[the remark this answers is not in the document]
piyifan123: check this.

veomni/data/multimodal/multimodal_chat_template.py
```diff
@@ -32,7 +32,7 @@
 class MultimodalChatTemplate(ChatTemplate):
     @abstractmethod
     def encode_messages(
-        self, messages: Sequence[Dict[str, str]], num_tokens: Dict[str, List[int]] = defaultdict(list), **kwargs
+        self, messages: Sequence[Dict[str, str]], num_tokens: Dict[str, List[int]] = None, **kwargs
```
piyifan123, line 35: so all child classes of this have done `if none: initialize` logic?

piyifan123 (commented):
@FoolPlayer need another pairs of eyes to review 

i've gone through once and seems mostly fine.

veomni/data/multimodal/multimodal_chat_template.py
LiuzcEECS: Yes

tests/data/multimodal/test_video_utils.py
LiuzcEECS: if we want to noqa for all cases, we can just ignore `B905` in the config file

tests/ops/test_comp.py
```diff
@@ -30,7 +30,7 @@ def fast_pos_embed_interpolate_ref(self, grid_thw):
 
     device = self.pos_embed.weight.device
     dtype = self.pos_embed.weight.dtype
-    for t, h, w in zip(grid_ts, grid_hs, grid_ws):
+    for _t, h, w in zip(grid_ts, grid_hs, grid_ws, strict=False):
```
FoolPlayer: Explicitly stating the contents in `zip` file may have mismatched length is indeed necessary. But maybe some cases expect the strict same length, like handling the vedio data here.
@Coach257 Please have a review.

tests/data/multimodal/test_video_utils.py
FoolPlayer: Agree `B905` rule. But some case should require strcit same lenght.

tests/ops/test_comp.py
Coach257: Yes, `strict = True` here.

PR #526, opened 2026-03-02
[model] feat: update Qwen2 to transformers v5

xzzWZY:
Upgrade Qwen 2 to transformers v5, with tests in ci.


Luosuu:
Thank you very much!
Please paste a figure to show loss/grad norm parity with transformers v4.57.3
By the way, is there a way to test it in model tests? cc @Coach257

veomni/models/transformers/qwen2/generated/patched_modeling_qwen2_gpu.diff
```diff
@@ -0,0 +1,293 @@
...
++
++
+ @auto_docstring
+ class Qwen2Model(Qwen2PreTrainedModel):
+     def __init__(self, config: Qwen2Config):
+@@ -385,9 +388,7 @@
+         if position_ids is None:
+             position_ids = cache_position.unsqueeze(0)
+
+-        # It may already have been prepared by e.g. `generate`
```
piyifan123, line 207: are you using the latest main? i think patchgen would not make such diff on functions (__init__ here) not patched?

PR #536, opened 2026-03-05
[ops] feat: add DCP consolidation patch for HDFS FUSE compatibility

hjshi84:
### What does this PR do?

Add a monkey patch for PyTorch DCP (Distributed Checkpoint) safetensors consolidation to support HDFS FUSE compatibility.

  Problem: When saving HuggingFace format checkpoints with enable_consolidation=True, PyTorch internally uses r+b mode for random write access, which is not supported by HDFS via FUSE (append-only file systems).

  Solution: This PR patches the _process_output_file function in torch.distributed.checkpoint._consolidate_hf_safetensors to use ab (append) mode instead.

### Checklist Before Starting

 - Search for similar PRs. Paste at least one query link here: N/A (new feature)
  - Format the PR title as [{modules}] {type}: {description}
    - Title: [ops] feat: add DCP consolidation patch for HDFS FUSE compatibility

### Test

  This patch was tested on HDFS FUSE environment with PyTorch 2.9.1:
  - Successfully saves consolidated HuggingFace format checkpoints
  - No OSError: [Errno 95] Operation not supported errors

### API and Usage Example

No API changes. The patch is applied automatically when using distributed safetensors save:

  from veomni.utils.save_safetensor_utils import _save_hf_safetensor_distributed

  # Patch is applied just-in-time inside this function
  _save_hf_safetensor_distributed(...)

### Design & Code Changes

High-level Design

  1. Just-in-time Patching: The patch is applied automatically before dcp.save() is called, no user action required
  2. Version Guard: Includes PyTorch version check (requires 2.9.x) to ensure compatibility
  3. Duplicate Prevention: Uses _dcp_consolidation_patch_applied flag to prevent duplicate patching

  Specific Changes

  | File                                  | Change                                           |
  |---------------------------------------|--------------------------------------------------|
  | veomni/ops/dcp_consolidation.py       | New file: DCP consolidation patch implementation |
  | veomni/utils/save_safetensor_utils.py | Integrate patch call                             |
  | docs/usage/hdfs_fuse_patch.md         | New file: documentation                          |

  - Read the Contribute Guide.
  - Apply pre-commit checks: pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always
  - Add / Update the documentation.
  - Add unit or end-to-end test(s) to the CI workflow to cover all the code. If not feasible, explain why:
    - Explanation: This patch modifies runtime behavior of PyTorch internal functions and requires HDFS FUSE environment to test properly. Manual testing on HDFS FUSE cluster has been performed.

### DIFF
--- a/veomni/ops/dcp_consolidation.py
+++ b/veomni/ops/dcp_consolidation.py
@@ -68,20 +68,22 @@ def apply_dcp_consolidation_patch():
             f"but got {torch.__version__}. Please update the patch or verify compatibility."
         )
 
-    try:
-        import torch.distributed.checkpoint._consolidate_hf_safetensors as hf_module
-    except ImportError:
-        return
+    import torch.distributed.checkpoint._consolidate_hf_safetensors as hf_module
 
     if not hasattr(hf_module, "_process_output_file"):
-        return
+        raise RuntimeError(
+            f"torch.distributed.checkpoint._consolidate_hf_safetensors does not have "
+            f"_process_output_file attribute. Please verify torch {_REQUIRED_TORCH_VERSION}.x compatibility."
+        )
 
     # Define the replacement function logic
+    # This is a modified version of torch.distributed.checkpoint._consolidate_hf_safetensors._process_output_file
+    # Original: https://github.com/pytorch/pytorch/blob/v2.9.1/torch/distributed/checkpoint/_consolidate_hf_safetensors.py
+    # Key change: Use append mode ("ab") instead of read-write mode ("r+b") for HDFS FUSE compatibility
     def _process_output_file_impl(output_file, output_data, input_files_data):
-        """Process output file using append mode instead of r+b."""
         sorted_tensors = sorted(output_data.fqn_data.items(), key=lambda x: x[1].offset_in_file)
 
-        with open(output_file, "ab") as output_stream:
+        with open(output_file, "ab") as output_stream:  # Changed from "r+b"
             for tensor_fqn, tensor_fqn_data in sorted_tensors:
                 full_tensor_mv = memoryview(
                     bytearray(math.prod(tensor_fqn_data.shape_in_file) * tensor_fqn_data.dtype_size)


veomni/ops/__init__.py
```diff
@@ -49,6 +49,11 @@ def apply_ops_patch():
         apply_veomni_fused_moe_patch()
         logger.info_rank0("✅ VeOmni ops patch applied.")
 
+    # Always apply DCP consolidation patch for HDFS FUSE compatibility
```
piyifan123: monkey patch in ops/__init__.py does not sounds the right place. can we monkey patch inside the hf checkpointing logic? i.e., monkey patch just in time when we need to run the hf checkpoining logic.

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,112 @@
...
+        with open(output_file, "ab") as output_stream:
+            ...
+
+    Note: Append mode requires tensors to be processed in offset order, which is
+    already ensured by sorting tensors before writing.
+
+    The patch uses types.FunctionType to create a new function with __globals__
+    bound to the target module, enabling access to internal functions like
+    _read_tensor_data_mmap and _write_sub_tensor_to_file_optimized.
+    """
```
piyifan123, line 56: can we add a torch version assertion here? i.e., if it's not torch 2.9.1 then throw an exception so that we know we need to update this whenever we update torch versions.

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,112 @@
...
+    # This ensures that internal functions like _read_tensor_data_mmap are resolved correctly
+    patched_func = types.FunctionType(
+        _process_output_file_impl.__code__,
+        hf_module.__dict__,  # Use target module's globals for symbol resolution
+        _process_output_file_impl.__name__,
+        _process_output_file_impl.__defaults__,
+        _process_output_file_impl.__closure__,
+    )
+
+    hf_module._process_output_file = patched_func
```
piyifan123, line 134: should we use a global var to check if monkey patch has already been done?

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,112 @@
...
+
+This module provides a monkey patch to fix compatibility issues with distributed
+file systems (e.g., HDFS via FUSE) that do not support r+b (read-write binary)
+file mode for random write access.
+"""
+
+import types
+
+
+def apply_dcp_consolidation_patch():
```
piyifan123, line 36: nit: add a doc under docs to explain this.

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,112 @@
...
+
+"""Patch for PyTorch DCP (Distributed Checkpoint) safetensors consolidation.
+
+This module provides a monkey patch to fix compatibility issues with distributed
+file systems (e.g., HDFS via FUSE) that do not support r+b (read-write binary)
+file mode for random write access.
+"""
+
+import types
+
```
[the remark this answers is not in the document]
piyifan123: is this a right suggestion?

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,133 @@
...
+    if not torch.__version__.startswith(_REQUIRED_TORCH_VERSION):
+        raise RuntimeError(
+            f"DCP consolidation patch requires torch {_REQUIRED_TORCH_VERSION}.x, "
+            f"but got {torch.__version__}. Please update the patch or verify compatibility."
+        )
+
+    try:
+        import torch.distributed.checkpoint._consolidate_hf_safetensors as hf_module
+    except ImportError:
+        return
```
piyifan123: nit: do not guard here. since we have already verified torch version here. this import should succeed. swallowing errors could hide some real problems.

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,133 @@
...
+
+    if not hasattr(hf_module, "_process_output_file"):
```
piyifan123, line 73: same here.

veomni/ops/dcp_consolidation.py
```diff
@@ -0,0 +1,133 @@
...
+        return
+
+    # Define the replacement function logic
+    def _process_output_file_impl(output_file, output_data, input_files_data):
+        """Process output file using append mode instead of r+b."""
+        sorted_tensors = sorted(output_data.fqn_data.items(), key=lambda x: x[1].offset_in_file)
+
+        with open(output_file, "ab") as output_stream:
```
piyifan123: nit: annotate the place where we changed from original torch code. and include a github link in comment to the exact version tag of the file so that we can easy to compare if the patch is correct.

veomni/ops/dcp_consolidation.py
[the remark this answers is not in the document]
hjshi84: No.
patched_func = types.FunctionType(
      _process_output_file_impl.__code__,
      hf_module.__dict__,  # Use target module's globals for symbol resolution
      ...
  )

  The patched function uses hf_module.__dict__ as its __globals__. Since the original PyTorch module torch.distributed.checkpoint._consolidate_hf_safetensors already imports math, the patched function will resolve math from hf_module's namespace at runtime.

  Adding import math to this file would be unnecessary - it wouldn't be used because the function's globals are explicitly bound to the target module's dict, not this file's namespace.

PR #542, opened 2026-03-06
[parallel, model] feat: add Ulysses sequence parallelism to Qwen3.5 GatedDeltaNet

Luosuu:
### What does this PR do?

Adds Ulysses sequence parallelism (all-to-all head sharding) support to the `Qwen3_5GatedDeltaNet` linear attention layer. Without this, SP only works for the softmax attention layers in the Qwen3.5 hybrid model; linear attention layers would compute on partial sequences without proper communication.

Reference implementation: `chefmodels/qwen3_next/modeling_qwen3_next.py` (lines 720-941).

### Checklist Before Starting

- [x] Search for similar PRs. Paste at least one query link here: https://github.com/ByteDance-Seed/VeOmni/pulls?q=is%3Apr+qwen3_5+ulysses
- [x] Format the PR title as `[{modules}] {type}: {description}`

### Test

- Added `tests/parallel/ulysses/test_qwen3_5_gated_deltanet_ulysses.py` with 4 test categories:
  1. **Conv1d weight slicing** (single-GPU): validates `_get_local_conv1d_weight` produces correct sharded weights
  2. **SP forward/backward equivalence** (multi-GPU): compares SP outputs and gradients against non-SP baseline
  3. **SP forward determinism** (multi-GPU): 100 repeated forward passes produce identical outputs
  4. **Non-SP forward determinism** (single-GPU): regression guard
- Added two training configs (`configs/text/qwen3_5_sp{1,2}.yaml`) for manual end-to-end loss curve comparison (micro batch size=1 & sp=1 vs micro batch size= 2&  sp=2)

### API and Usage Example

No API changes. To enable Ulysses SP for Qwen3.5, set `ulysses_parallel_size > 1` in the training config:

```yaml
train:
  ulysses_parallel_size: 2
```

### Design & Code Changes

**`qwen3_5_gpu_patch_gen_config.py`** (sole code change):

1. **New import**: `gather_seq_scatter_heads`, `gather_heads_scatter_seq` from `veomni.distributed.sequence_parallel.ulysses`
2. **New method `_get_local_conv1d_weight`**: Shards depthwise conv1d weights by extracting Q/K/V channels for local heads based on Ulysses rank
3. **Rewritten `Qwen3_5GatedDeltaNet.forward`** with SP support:
   - After projection: split QKV → reshape to heads → `gather_seq_scatter_heads` (all-to-all)
   - Conv1d with sharded weights on full sequence / local heads
   - Slice `A_log`/`dt_bias` for local V-heads
   - Core attention on local heads with full sequence
   - `gather_heads_scatter_seq` to restore sequence-sharded layout before gated norm
   - Non-SP path unchanged, guarded by `get_parallel_state().ulysses_enabled`

**CI** (`.github/workflows/gpu_unit_tests.yml`):
- Path-based conditional: test only runs when `veomni/distributed/sequence_parallel/` or `veomni/models/transformers/qwen3_5/` files are changed
- Runs under the transformers-v5 section (Qwen3.5 requires transformers >= 5.0)

- [x] Apply [pre-commit checks](https://github.com/ByteDance-Seed/VeOmni/blob/main/CONTRIBUTING.md?plain=1#L22): `pre-commit install && pre-commit run --all-files --show-diff-on-failure --color=always`
- [x] Add unit or end-to-end test(s) to [the CI workflow](https://github.com/ByteDance-Seed/VeOmni/tree/main/.github/workflows) to cover all the code.


configs/text/qwen3_5_sp2.yaml
```diff
@@ -0,0 +1,39 @@
+# Qwen3.5 SP equivalence test: dp=2, sp=2 (Ulysses enabled)
```
FoolPlayer: Maybe just one .yaml file for one model is OK. And we can add qwen3.5 example with SP to doc.

Luosuu: sure

pyproject.toml
```diff
@@ -304,4 +304,5 @@ exclude = ["veomni/models/transformers/**/generated/*.py"]
 [tool.pytest.ini_options]
 markers = [
     "benchmark: marks tests as benchmark tests (deselect with '-m \"not benchmark\"')",
+    "qwen3_5_ulysses: Qwen3.5 GatedDeltaNet Ulysses SP tests (run when ulysses or qwen3_5 code is touched)",
```
FoolPlayer, line 307: Is this just for debug?

Luosuu: To reduce CI time cost, Linear attention SP test takes a lot of time to finish

Luosuu: So only when SP or Qwen3.5 are touched, the CI would be triggered

Issue #540, opened 2026-03-06
[full-determinism] Qwen_2_vl: training loss fails to converge to the same values with image inputs


xzzWZY:
### System Info

Transformers: 4.57.3
torch: 2.9.1+cu129
NVIDIA driver: 535.161.08
nvidia-cudnn-cu12: 9.15.1.9

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

For Qwen 2 vl:

1. training losses aren't identical between two runs, with same setting for dataset with images
2. training losses are identical wit text-only dataset like tulu-3-sft-mixture
3. set  "torch.use_deterministic_algorithms(True, warn_only=False)" wouldn't lead to any errors

The detailed setting is shown here:

```
{
  "model": {
    "config_path": "../hf_cache/Qwen2-VL-7B-Instruct/",
    "model_path": "../hf_cache/Qwen2-VL-7B-Instruct/",
    "tokenizer_path": "../hf_cache/Qwen2-VL-7B-Instruct/",
    "safetensor_idx_path": "../hf_cache/Qwen2-VL-7B-Instruct/model.safetensors.index.json",
    "foundation": {},
    "encoders": {},
    "decoders": {},
    "input_encoder": "encoder",
    "output_encoder": "decoder",
    "encode_target": false,
    "attn_implementation": "veomni_flash_attention_2_with_sp",
    "moe_implementation": null,
    "basic_modules": [],
    "encoder_data_balance": false,
    "encoder_data_balance_sorting_algo": "post_mbs_balancing_greedy_without_pad"
  },
  "data": {
    "train_path": "../hf_cache/ShareGPT4V_dataset/sharegpt4v_instruct_gpt4-vision_cap100k_coco.json",
    "eval_path": null,
    "train_size": 80000000,
    "train_sample": 10000,
    "data_type": "conversation",
    "dataloader_type": "native",
    "datasets_type": "iterable",
    "multisource_datasets_type": "interleave",
    "source_name": "sharegpt4v_sft",
    "dyn_bsz_buffer_size": 200,
    "text_keys": "messages",
    "chat_template": "qwen2vl",
    "max_seq_len": 4096,
    "num_workers": 2,
    "prefetch_factor": 2,
    "drop_last": true,
    "pin_memory": true,
[...]

PR #543, opened 2026-03-09
[model] feat: upgrade Qwen2 vl to transformers v5

xzzWZY:
Upgrade Qwen 2 vl to transformers v5 with tests in ci.

### DIFF
--- a/tests/models/test_models_patch.py
+++ b/tests/models/test_models_patch.py
@@ -261,6 +261,8 @@ _TEST_CASES_TRANSFORMERS_V5 = [
         _DEFAULT_RTOL,
         _DEFAULT_ATOL,
         id="qwen2_vl",
+    ),
+    pytest.param(
         "./tests/toy_config/qwen2_toy/config.json",
         False,
         _DEFAULT_RTOL,


Coach257:
Have you fixed #540

xzzWZY:
> Have you fixed #540

image

The issue resolved if we disable liger kernel. Thanks. 

There's a question remaining that current patchgen for transformers v5 doesn't support conditional patch/replacement. I may discuss with @piyifan123 to check if we need to update logic of patchgen to support it.

piyifan123:
> > Have you fixed #540
> 
> image
> The issue resolved if we disable liger kernel. Thanks.
> 
> There's a question remaining that current patchgen for transformers v5 doesn't support conditional patch/replacement. I may discuss with @piyifan123 to check if we need to update logic of patchgen to support it.

do we know which liger kernel replacement caused the issue?

veomni/models/transformers/qwen2_vl/generated/patched_modeling_qwen2_vl_gpu.diff
```diff
@@ -0,0 +1,564 @@
...
+-        use_cache: bool | None = None,
+-        output_attentions: bool | None = None,
+-        output_hidden_states: bool | None = None,
+-        pixel_values: torch.Tensor | None = None,
+-        pixel_values_videos: torch.FloatTensor | None = None,
+-        image_grid_thw: torch.LongTensor | None = None,
+-        video_grid_thw: torch.LongTensor | None = None,
+-        rope_deltas: torch.LongTensor | None = None,
+-        cache_position: torch.LongTensor | None = None,
++        input_ids: Optional[torch.LongTensor] = None,
```
piyifan123: nit: make the patch code to match the forward signature to remove this unessary diff

PR #550, opened 2026-03-10
[data, model] fix: Fix Qwen3OmniMoE audio dummy input shape

TimYangst:
### What does this PR do?

> Fix two bugs:
> 1. **Qwen3OmniMoE audio encoder**: Correct the `dummy_forward` input shape from `(num_mel_bins, total_len)` to `(total_len, num_mel_bins)` to match what the patched forward expects before it permutes the tensor.

### Checklist Before Starting

- [x] PR title follows `[{modules}] {type}: {description}` format

### Test

> No new tests required. Both are correctness fixes matching existing behavior:
> - Model: `dummy_forward` is only called during FSDP no-audio ranks; shape fix prevents silent shape mismatch.

### API and Usage Example

> No API changes.

### Design & Code Changes
- `veomni/models/transformers/qwen3_omni_moe/modeling_qwen3_omni_moe.py:555`: Fixed `dummy_forward` docstring and dummy tensor shape from `(num_mel_bins, min_len)` to `(min_len, num_mel_bins)` to match the patched forward's expected input layout.

- [ ] Applied pre-commit checks

### DIFF
--- a/veomni/data/data_transform.py
+++ b/veomni/data/data_transform.py
@@ -372,7 +372,9 @@ def process_sample_qwen_omni(
     image_token_id, video_token_id, audio_token_id = get_omni_token_ids(processor)
 
     source = kwargs["source_name"] if "source_name" in kwargs else sample["source_name"]
-    conversations = sample["conversations"] if ("conversations" in sample and sample["conversations"] is not None and len(sample["conversations"]) > 0) else sample
+    conversations = (
+        sample["conversations"] if ("conversations" in sample and len(sample["conversations"]) > 0) else sample
+    )
     conversations = conv_preprocess(source, conversations, **kwargs)
     input_conversations = [
         {


veomni/data/data_transform.py
```diff
@@ -372,7 +372,11 @@ def get_omni_token_ids(processor: "ProcessorMixin") -> tuple[int, int, int]:
     image_token_id, video_token_id, audio_token_id = get_omni_token_ids(processor)
 
     source = kwargs["source_name"] if "source_name" in kwargs else sample["source_name"]
-    conversations = sample["conversations"] if ("conversations" in sample and sample["conversations"]) else sample
+    conversations = (
```
LiuzcEECS, line 375: From modelchef's side, we cannot use the transforming funcs here without updating it.

compare veomni-recipes/src/veomni_recipes/data/vlm_data_process.py in modelchef and veomni/data/data_transform.py in Open-VeOmni, we can see there are some difference in the func process_sample_qwen3_vl
compare _process_sample_omni in submodules/Open-VeOmni/tasks/data/vlm_data_process.py in modelchef and process_sample_qwen_omni in veomni/data/data_transform.py in Open-VeOmni

Those differences have already caused some task failures

TimYangst: Got it, thanks for catching the issue! Let me revert the change in this field, and let's see if we can address the issue in the dataset itself.

TimYangst: As discussed offline, let's add back the check here to increase logic robustic.

PR #557, opened 2026-03-12
[model] feat: support kernel_hubs to patch external kernels (liger kernels)

xzzWZY:
Add support to use huggingface kernel_hubs to patch external kernels. No test currently.


veomni/models/transformers/qwen3/liger_kernels_hub/layers.py
```diff
@@ -0,0 +1,30 @@
...
+``offset``, ``casting_mode``) are lazily patched with sensible defaults so
+the wrapper can be bound to the original model module via ``MethodType``.
+"""
+
+import torch.nn as nn
+from liger_kernel.transformers.rms_norm import LigerRMSNorm as _OrigLigerRMSNorm
+from liger_kernel.transformers.rope import liger_rotary_pos_emb as _orig_liger_rotary_pos_emb
+
+
+class LigerRMSNorm(nn.Module):
```
piyifan123: not quite sure how this works. where is the parameter defined (since we only call the forward of _OrigLigerRMSNorm? if so, should we use the functional interface (not sure if Liger has one) to be more idomatic?

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
```diff
@@ -48,44 +46,44 @@
...
-
-    return liger_rotary_pos_emb(
-        q,
-        k,
-        cos,
-        sin,
-        position_ids=position_ids,
-        unsqueeze_dim=unsqueeze_dim,
-    )
+# config.patches.append(
```
TimYangst, line 49: Just cleanup these config if we don't need them any more?

PR #558, opened 2026-03-12
[task] feat: Add DPOTrainer

FoolPlayer:
### What does this PR do?

1. Add DPOTrainer
2. Use Qwen3-0.6B and trl-lib/ultrafeedback_binarized as DPO example


veomni/trainer/base_dpo_trainer.py
```diff
@@ -0,0 +1,214 @@
...
+from ..ops.batch_invariant_ops import set_batch_invariant_mode
+from ..utils import helper, logging
+from ..utils.constants import IGNORE_INDEX
+from .base import BaseTrainer
+
+
+logger = logging.get_logger(__name__)
+
+
+class BaseDPOTrainer(BaseTrainer):
```
piyifan123: OOC, why do we inherit from BaseTrainer? can we reference to it?

veomni/trainer/dpo_trainer.py
```diff
@@ -0,0 +1,197 @@
...
+            tensors = [f[key] for f in features]  # each [2, L_i]
+            pad_value = _DPO_PAD_VALUES.get(key, 0)
+            max_len = max(t.shape[-1] for t in tensors)
+            padded = [F.pad(t, (0, max_len - t.shape[-1]), value=pad_value) for t in tensors]
+            # cat along dim=0: [2, L], [2, L], ... → [2*B, max_L]
+            batch[key] = torch.cat(padded, dim=0)
+        return batch
+
+
+class TextDPOTrainer:
```
piyifan123: should we name the file as text_dpo_trainer.py?

veomni/trainer/dpo_trainer.py
```diff
@@ -0,0 +1,197 @@
...
+
+_DPO_PAD_VALUES = {
+    "input_ids": 0,
+    "attention_mask": 0,
+    "labels": IGNORE_INDEX,
+}
+
+
+@dataclass
+class DPOCollator(DataCollator):
```
piyifan123: add unit tests for this?

veomni/trainer/dpo_trainer.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __call__(self, features: List[Dict[str, torch.Tensor]]) -> Dict[str, torch.Tensor]:
+        batch = {}
+        for key in features[0].keys():
+            tensors = [f[key] for f in features]  # each [2, L_i]
+            pad_value = _DPO_PAD_VALUES.get(key, 0)
+            max_len = max(t.shape[-1] for t in tensors)
+            padded = [F.pad(t, (0, max_len - t.shape[-1]), value=pad_value) for t in tensors]
+            # cat along dim=0: [2, L], [2, L], ... → [2*B, max_L]
+            batch[key] = torch.cat(padded, dim=0)
```
piyifan123: would cu_seqlens calculation handle padding here?

veomni/trainer/base_dpo_trainer.py
```diff
@@ -0,0 +1,264 @@
...
+
+        self.reference_model.requires_grad_(False)
+
+        self.reference_model = build_parallelize_model(
+            self.reference_model,
+            init_device=args.train.init_device,
+            weights_path=args.model.model_path,
+            enable_full_shard=args.train.accelerator.fsdp_config.full_shard,
+            enable_reshard_after_forward=args.train.accelerator.fsdp_config.reshard_after_forward,
+            enable_mixed_precision=args.train.enable_mixed_precision,
```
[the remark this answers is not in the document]
piyifan123: would this cause issue when args.dpo_config.refer_model_precision is set to fp32 but mixed_precision enabled? should we always false here (with a comment to explain)?

veomni/data/data_transform.py
```diff
@@ -105,6 +106,76 @@ def process_conversation_example(
     return [tokenized_example]
 
 
+_DPO_PAD_VALUES = {"input_ids": 0, "attention_mask": 0, "labels": IGNORE_INDEX}
+
+
+@DATA_TRANSFORM_REGISTRY.register("dpo")
+def process_dpo_example(
```
piyifan123, line 113: better to have unit tests for data process functions

veomni/trainer/base_dpo_trainer.py
FoolPlayer: I wanna use this BaseDPOTrainer as the base Trainer for DPO, and then the sub-task like DPOtext reference  BaseDPOTrainer

veomni/trainer/dpo_trainer.py
FoolPlayer: OK, make sense.

veomni/trainer/dpo_trainer.py
FoolPlayer: OK

veomni/trainer/dpo_trainer.py
FoolPlayer: 🤔 I'll fix

veomni/trainer/base_dpo_trainer.py
Luosuu: I think if in this case we do not need DPOBase trainer? the base trainer is DPOText trainer then. My point is that we should avoid inheritance as much as possible

veomni/arguments/arguments_types.py
```diff
@@ -845,6 +847,46 @@ def train_steps(self) -> int:
         return self._train_steps
 
 
+# ================================ DPO Arguments ======================================
+
+
+@dataclass
+class DPOConfig:
```
Coach257, line 872: Maybe mv dpoconfig to the head of dpo_trainer?

tests/data/test_classification_data_processor.py
```diff
@@ -2,7 +2,7 @@
 import torch
 from transformers import AutoTokenizer
 
-from veomni.data.data_transform import process_classification_example
+from veomni.data.data_transform import process_classification_example, process_dpo_example
```
Coach257: Create a new testscript named test_dpo_data_processor

veomni/trainer/text_dpo_trainer.py
```diff
@@ -0,0 +1,367 @@
...
+
+_DPO_PAD_VALUES = {
+    "input_ids": 0,
+    "attention_mask": 0,
+    "labels": IGNORE_INDEX,
+}
+
+
+@dataclass
+class DPOCollator(DataCollator):
```
Coach257: Flatten 2\*B to B, 2\*seqlen; use cuseqlens，so we can use MainCollator here and directly support data sp slice.

tests/data/test_classification_data_processor.py
piyifan123: +1

PR #568, opened 2026-03-16
[ci] fix: Fix transformers-v5 CI by correctly select the v5 dep group.

piyifan123:
Without the `--no-group transformers-stable --extra transformers5-exp` declaration, uv still falls back to v4 since it's the default group.


piyifan123:
Closed with https://github.com/ByteDance-Seed/VeOmni/pull/574. Re-enabling everything would cause OOM issue so will deal with it later.

PR #670, opened 2026-04-17
[model, ci] test: bitwise-equal HF vs veomni logits; fix bf16 MoE dtype divergences

Luosuu:
### What does this PR do?

Adds a pytest suite that asserts veomni's forward logits are **bitwise identical** (via `torch.equal`) to native HuggingFace across qwen3 / qwen3_moe / deepseek_v3 toy configs, in both `eager`+fp32 and `flash_attention_2`+bf16 configurations. Fixes three production bugs that the test surfaced in veomni's MoE bf16 dtype handling and deepseek_v3's unconditional GPU kernel patching.

### Checklist Before Starting

- Related PRs/issues: none
- PR title follows `[{modules}] {type}: {description}` format — modules `model, ci`, type `test`.

### Test

All 6 cases pass bitwise:

| case | attn_implementation | dtype | result |
|---|---|---|---|
| qwen3-toy | eager | fp32 | PASS |
| qwen3_moe-toy | eager | fp32 | PASS |
| deepseek_v3-toy | eager | fp32 | PASS |
| qwen3-toy | flash_attention_2 | bf16 | PASS |
| qwen3_moe-toy | flash_attention_2 | bf16 | PASS |
| deepseek_v3-toy | flash_attention_2 | bf16 | PASS |

```
============================== 6 passed in 32.92s ==============================
```

Test command:

```bash
VEOMNI_USE_LIGER_KERNEL=0 VEOMNI_USE_FUSED_KERNELS=0 \
  pytest -v tests/models/test_models_logits_equal.py
```

### API and Usage Example

New env var **`VEOMNI_USE_FUSED_KERNELS`** (default `"1"`, preserves current production behavior). Setting `"0"` causes `apply_veomni_deepseek_v3_gpu_patch` to early-return, leaving HF's native RoPE + RMSNorm in place. Previously the Triton RoPE bmm + batch-invariant RMSNorm substitution was unconditional in the `else` branch of the Liger gate, with no way to opt out — this made bitwise parity with HF impossible to achieve for deepseek_v3.

### Design & Code Changes

**Test (`tests/models/test_models_logits_equal.py`)**
- Single parametrized test `test_logits_bitwise_equal(case)` over 6 `Case`s (3 models × 2 attn/dtype combinations). HF is random-initialised from the toy config; its state dict is copied to veomni via the existing `weight_sync_adapters` for MoE layout conversion.
- Execution order within each test is mandatory: HF forward before any veomni model build, because `build_foundation_model` triggers `apply_veomni_*_patch()` which monkey-patches HF module classes process-wide.
- HF model is built under `torch.device(""cuda"")` so init-time buffers (rotary `inv_freq` under YaRN) use the same arithmetic path as veomni's `CustomizedModelingLoader`.
- Determinism: TF32 off (matmul + cudnn), cudnn deterministic, `use_deterministic_algorithms(warn_only=True)`, `CUBLAS_WORKSPACE_CONFIG=:4096:8`.

**Test isolation helper (`tests/models/utils.py`)**
- `apply_veomni_hf_unpatch()` restores pristine HF class attributes. Snapshots are captured at `utils.py` import time, before veomni's lazy per-model patches can fire. Called before every HF build so patches leaked by an earlier veomni build in the same pytest session do not poison subsequent HF models.
- Made `from veomni.data.dummy_dataset import build_dummy_dataset` lazy (moved into `prepare_data`) so importers that only need the unpatch helpers don't pay the torchcodec/ffmpeg import cost.

**Production fix 1: `veomni/utils/env.py` + `veomni/models/transformers/deepseek_v3/gpu_patch.py`**
- Register `VEOMNI_USE_FUSED_KERNELS` in `ENV_DEFAULTS` (default `""1""`).
- `apply_veomni_deepseek_v3_gpu_patch()` now early-returns when the env var is `""0""`, skipping both the Liger branch and the Triton RoPE + batch-invariant RMSNorm branch. Current production behavior is preserved by the default.

**Production fix 2: `veomni/models/transformers/qwen3_moe/modeling_qwen3_moe.py`**
- `PatchQwen3MoeTopKRouter.forward` now casts `router_top_value` to the **input** dtype rather than `router_logits.dtype`. After `softmax(dtype=torch.float)`, `router_logits` is fp32, so the existing cast was a no-op; the intent (matching HF's `routing_weights.to(hidden_states.dtype)`) was lost.
- In fp32 both casts are no-ops so the bug was invisible. In bf16, keeping router weights in fp32 promoted the expert multiplication `expert_output(bf16) * routing_weights(fp32)` to fp32 → cast back to bf16, which rounded differently from HF's bf16×bf16=bf16.

**Production fix 3: `veomni/models/transformers/deepseek_v3/modeling_deepseek_v3.py`**
- `PatchDeepseekV3NaiveMoe.forward` eager branch now allocates `final_hidden_states` in `top_k_weights.dtype` (fp32) and casts to input dtype once at return, matching HF's `DeepseekV3MoE.moe` which uses `dtype=topk_weights.dtype`.
- Previously it allocated in `hidden_states.dtype` (bf16) and cast each expert contribution to bf16 before `index_add_`, rounding per-expert rather than once at the end. Again invisible in fp32; visible in bf16.

- [x] Applied pre-commit checks (`make style` + `make quality` clean)
- [ ] Added/updated documentation — no docs impact; the new env var's default preserves prior behavior
- [ ] If `tasks/` training scripts were moved or renamed — N/A
- [ ] Added tests to CI workflow — the new test file is standalone and self-skipping (skips on non-CUDA, on transformers >= 5); wiring into nightly/presubmit GPU CI can be done in a follow-up

### DIFF
--- a/tests/models/test_models_logits_equal.py
+++ b/tests/models/test_models_logits_equal.py
@@ -8,6 +8,8 @@ from typing import Optional
 import pytest
 import torch
 
+from veomni.utils.device import IS_CUDA_AVAILABLE, empty_cache, get_device_type, get_torch_device
+
 # Importing `utils` now captures pristine HF class attributes (in its
 # `_PRISTINE_HF_CLASSES` snapshot) before any veomni import has a chance to
 # monkey-patch them. `apply_veomni_hf_unpatch()` restores them; we call it
@@ -93,7 +95,6 @@ CASES = [
 
 
 def _apply_determinism():
-    torch.backends.cuda.matmul.allow_tf32 = False
     torch.backends.cudnn.allow_tf32 = False
     torch.backends.cudnn.deterministic = True
     torch.backends.cudnn.benchmark = False
@@ -102,22 +103,22 @@ def _apply_determinism():
 
 def _release():
     gc.collect()
-    if torch.cuda.is_available():
-        torch.cuda.empty_cache()
+    if IS_CUDA_AVAILABLE:
+        empty_cache()
 
 
 def _build_hf_model(case: Case):
-    """Return a CUDA, eval-mode HF model randomly initialised from config."""
+    """Return a device-resident, eval-mode HF model randomly initialised from config."""
     from transformers import AutoConfig, AutoModelForCausalLM
 
     apply_veomni_hf_unpatch()
     config = AutoConfig.from_pretrained(case.path)
     torch.manual_seed(0)
-    torch.cuda.manual_seed_all(0)
-    # Init directly on CUDA so init-time buffers (e.g. rotary `inv_freq`) use
-    # the same arithmetic path as the veomni build, which allocates under
-    # `torch.device("cuda")` via its CustomizedModelingLoader.
-    with torch.device("cuda"):
+    get_torch_device().manual_seed_all(0)
+    # Init directly on device so init-time buffers (e.g. rotary `inv_freq`)
+    # use the same arithmetic path as the veomni build, which allocates under
+    # `torch.device(get_device_type())` via its CustomizedModelingLoader.
+    with torch.device(get_device_type()):
         model_hf = AutoModelForCausalLM.from_config(
             config,
             torch_dtype=_DTYPE_MAP[case.dtype],
@@ -127,7 +128,7 @@ def _build_hf_model(case: Case):
 
 
 def _build_veomni_model(case: Case, hf_state_dict):
-    """Return a CUDA, eval-mode veomni model with HF weights loaded."""
+    """Return a device-resident, eval-mode veomni model with HF weights loaded."""
     from veomni.models.auto import build_foundation_model
 
     model = build_foundation_model(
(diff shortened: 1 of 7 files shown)


tests/models/test_models_logits_equal.py
```diff
@@ -0,0 +1,217 @@
...
+    with torch.no_grad():
+        logits_ve = model_ve(input_ids=input_ids, use_cache=False).logits.detach().clone()
+    del model_ve, hf_state_dict
+    _release()
+
+    assert logits_hf.shape == logits_ve.shape, (
+        f"[{case.case_id}] shape mismatch: hf={tuple(logits_hf.shape)} ve={tuple(logits_ve.shape)}"
+    )
+
+    if not torch.equal(logits_hf, logits_ve):
```
FoolPlayer, line 206: Do we also need bitwise-check for grad after backward?

Luosuu: maybe let me do that in the following PRs to keep this PR minimal

TimYangst (approved):

PR #651, opened 2026-04-14
[ops] refactor: replace kernel env vars with args

FoolPlayer:
### What does this PR do?

Replace the all-or-nothing `VEOMNI_USE_LIGER_KERNEL` / `USE_GROUP_GEMM` env vars with per-op fields on `OpsImplementationConfig`, and reorganize `veomni/ops/` around a unified kernel registry so adding a new kernel/backend is a one-file change.

This started as the env-var → args refactor and grew into a full `veomni/ops/` reorganization once the patch surface was visible across every model.

### Motivation

- Env-var control was binary and global — you couldn't, e.g., turn on Liger RMSNorm but keep eager cross-entropy.
- Each model's `gpu_patch.py` / `npu_patch.py` had duplicated `if ligerkernel: setattr(...)` blocks that were drifting out of sync.
- New ops/backends (DeepSeek V3 deterministic RoPE, Wan Triton RMSNorm, NPU chunked loss) had no clean home.

### Key changes

**1. Per-op config fields** (`veomni/arguments/arguments_types.py`)

Five new `str` fields on `OpsImplementationConfig`, all defaulting to `"eager"` (no implicit "auto" — users opt in):

| Field | Backends |
|---|---|
| `cross_entropy_loss_implementation` | `eager`, `liger_kernel`, `npu` (chunked loss) |
| `rms_norm_implementation` | `eager`, `liger_kernel`, `npu`, `triton`* |
| `swiglu_mlp_implementation` | `eager`, `liger_kernel` |
| `rotary_pos_emb_implementation` | `eager`, `liger_kernel`, `npu`, `triton`* |
| `load_balancing_loss_implementation` | `eager`, `triton` |

`*` `triton` is registered per-model via `extra_backends` (DeepSeek V3, Wan).

`__post_init__` validates backend availability (`liger_kernel` / `torch_npu` packages) up-front instead of failing with a cryptic error at first batch.

**2. Removed env vars**
- `VEOMNI_USE_LIGER_KERNEL` → split into 4 per-op fields
- `USE_GROUP_GEMM` → `moe_implementation`
- `VEOMNI_ENABLE_CHUNK_LOSS` → `cross_entropy_loss_implementation="npu"`
- `MODELING_BACKEND` is **kept** (still controls the import-time attention patch).

**3. New unified registry** (`veomni/ops/config/`)
- `registry.py`: `OpSpec` / `BackendSpec` / `OpScope` + `register_op` / `apply_global_ops` / `apply_per_model_patches`.
[...]

PR #678, opened 2026-04-20
[BREAKING][ops, model] feat: registration-based kernel replacement framework

TimYangst:
[BREAKING][ops, model] feat: registration-based kernel replacement framework

### What does this PR do?

> Introduces a registration-based framework for kernel replacement. Kernels (cross-entropy loss, MoE experts, load-balancing loss, Liger fused ops) are registered into `KERNEL_REGISTRY` as `KernelSpec` entries and bound to consumers through `OpSlot` indirection, replacing the previous ad-hoc monkey-patching. Built on top of @piyifan's #569 (squashed into the first commit); this branch extends the framework with numerical alignment tests, documentation, and fixes surfaced while porting existing ops to the new dispatch path.

### ⚠️ Breaking Changes

> **`moe_implementation` accepted values and default changed.** The field was already single-valued on main (`Optional[Literal["eager", "fused", "fused_quack"]]`, default `None`). This PR:
>
> 1. Renames `"fused"` → `"fused_triton"` so the Triton backend is explicit in the value itself.
> 2. Adds `"fused_npu"` as an explicit NPU value, removing the previous silent auto-pick where `moe_implementation: fused` on a machine with `torch_npu` installed would transparently switch to the NPU group-gemm kernel.
> 3. Drops `Optional` and changes the default from `None` to `"eager"`.
>
> Mismatched selections (e.g. `fused_triton` on NPU, `fused_quack` on sub-SM90) now raise at patch / bind time instead of silently falling back.
>
> Migrate any YAML / CLI that used the old values:
>
> ```diff
> model:
>   ops_implementation:
> -   moe_implementation: fused
> +   moe_implementation: fused_triton
> ```
>
> On NPU, the hardware-specific value is now required — the previous silent selection is gone:
>
> ```diff
> model:
>   ops_implementation:
> -   moe_implementation: fused        # (on NPU, silently used the NPU group-gemm)
> +   moe_implementation: fused_npu
> ```
>
> No other v4 model or config surface is affected — see the v4/v5 impact table below for the full list of transparent vs. user-visible changes.

### Checklist Before Starting

- Related PR: #569 (piyifan/kernel-patch-exp, squashed into commit `5848726c`; origin branch retained for attribution)
- PR title follows `[BREAKING][{modules}] {type}: {description}` format (breaking prefix added per `.github/workflows/check_pr_title.yml`)

### Test

> - `tests/ops/test_kernel_registry_numerical.py` — new numerical alignment tests for every registered kernel against the eager reference.
> - `tests/ops/test_moe_hw_gate.py` — new hardware-gate tests for MoE kernel selection. Mocks `torch_npu` / `IS_CUDA_AVAILABLE` / `get_gpu_compute_capability` so every (backend × hardware) mismatch (e.g. `fused_quack` on A100, `fused_triton` on NPU, `fused_npu` on GPU) raises with a clear message at bind time instead of silently falling back. Covers both the legacy `apply_veomni_fused_moe_patch` path and the OpSlot + `KERNEL_REGISTRY.resolve` path, plus an end-to-end reproducer through `_bind_veomni_ops`.
> - `tests/models/test_models_patch.py` — updated to exercise the new dispatch path.
> - `tests/e2e/utils.py` — e2e helper updates for the new config plumbing.
> - `.github/workflows/npu_unit_tests.yml` — added `test_moe_hw_gate.py` to the NPU ops job (the test is hardware-portable via mocks; GPU CI already runs the whole `tests/ops/` directory).
> - `make quality` passes locally.

### API and Usage Example

> New module: `veomni/ops/kernel_registry.py` — exposes `KERNEL_REGISTRY`, `KernelSpec`, `OpSlot`.
> New module: `veomni/ops/dispatch.py` — dispatch helpers used by consumers.
> `OpsImplementationConfig.moe_implementation` now accepts `"eager" | "fused_triton" | "fused_quack" | "fused_npu"` (was `"eager" | "fused" | "fused_quack"` on main). `"fused"` is renamed to `"fused_triton"` and `"fused_npu"` is added so the NPU backend must be opted into explicitly instead of being auto-picked when `torch_npu` is importable. Example YAML:
>
> ```yaml
> model:
>   ops_implementation:
>     moe_implementation: fused_triton  # eager | fused_triton | fused_quack | fused_npu
>     cross_entropy_loss_implementation: liger_kernel
>     load_balancing_loss_implementation: triton
> ```

### Design & Code Changes

> - **New framework**
>   - `veomni/ops/kernel_registry.py` — central registry (`register_op`, `KernelSpec`, `OpSlot`) with hardware-requirement checks, warnings on rebind, and documented field semantics.
>   - `veomni/ops/dispatch.py` — `OpSlot` dispatcher resolving against config at `build_foundation_model` time.
> - **Kernels migrated to registry**
>   - `cross_entropy`: OpSlot split into `causal` vs. `seq_cls` variants; single install point via `install_loss_mapping(impl)` inside `apply_ops_config`, which pre-binds `partial(ForCausalLMLoss, cross_entropy_fn=<impl>)` into `LOSS_MAPPING`. No runtime 3-way resolution and no silent fallback — missing kernel packages raise `RuntimeError` with a fix-it hint.
>   - `load_balancing_loss`: registered as a `KernelSpec` with `OpScope.GLOBAL`; legacy MoE dispatch decoupled from the loss path.
>   - `moe`, `liger` (RMSNorm / RoPE / SwiGLU): wired through the registry.
> - **Args**: `moe_implementation` accepted values changed from `"eager" | "fused" | "fused_quack"` (default `None`) to `"eager" | "fused_triton" | "fused_quack" | "fused_npu"` (default `"eager"`). `"fused"` is renamed to `"fused_triton"`, and NPU now requires the explicit `"fused_npu"` value instead of being silently auto-selected when `torch_npu` is importable.
>   - **Hardware-gate behavior:** Both dispatch paths now raise on hardware mismatch instead of falling through to eager:
>     - Legacy `apply_veomni_fused_moe_patch` — `fused_triton`/`fused_quack` on NPU, `fused_quack` on sub-SM90 GPUs, `fused_npu` without `torch_npu` all raise `RuntimeError` at patch time.
>     - OpSlot path — `KERNEL_REGISTRY.resolve` checks `HardwareRequirement` (device type + min compute capability) at `slot.bind()` time and raises.
>     - `_bind_veomni_ops` translates the user-facing `moe_implementation=fused_*` value to the bare registry name (`triton`/`quack`/`npu`) so the hardware-requirement gate fires correctly for the one op whose config field name doesn't follow the `<op_name>_implementation` convention.
> - **Model integration**: patch configs updated for qwen3, qwen3_moe, qwen3_5_moe (GPU/NPU), qwen3_vl_moe (GPU/NPU), qwen3_omni_moe, deepseek_v3. Generated patch files regenerated via `make patchgen`.
> - **Fixes**
>   - Run legacy MoE patch before model load (pre-init ordering).
>   - Device check correctness in `veomni/utils/device.py`.
>   - `build_foundation_model` emits a warning + installs `OpsImplementationConfig()` defaults when called outside `BaseTrainer`, so standalone scripts don't trip on HF's stock `ForCausalLMLoss` (which doesn't understand `hidden_states=` / `weights=`).
> - **Docs**: `docs/design/unified_kernel_registry.md` (new, full design doc) and `docs/design/kernel_selection.md` (updated to describe the registry-based selection path, including the `install_loss_mapping` single-install-point contract).

### Transformers v4 / v5 Impact

> The dispatch system has two layers; they have very different coupling to the transformers version.
>
> - **Process-global layer** (`LOSS_MAPPING`, GLOBAL scope function pointers, legacy `apply_veomni_fused_moe_patch`): binds shared module-level state from `apply_ops_config`. Works on **both v4 and v5** and is transparent to user code.
> - **Per-model OpSlot layer** (`KERNEL_REGISTRY` + `OpSlot` guards inside `generated/patched_modeling_*.py`): lives exclusively inside patchgen-generated modeling files and only runs for **v5-migrated models**. See `docs/transformers_v5/patchgen.md` for how those files are produced.
>
> | Change | v4 | v5 | How v4 is impacted |
> |---|:-:|:-:|---|
> | `install_loss_mapping` / `LOSS_MAPPING` (cross-entropy: eager / liger / NPU chunk loss) | ✅ | ✅ | Transparent. `self.loss_function(...)` routes through `LOSS_MAPPING["ForCausalLM" / "ForConditionalGeneration" / "ForSequenceClassification"]` on both versions. No user action. |
> | `load_balancing_loss_implementation` (GLOBAL scope) | ✅ | ✅ | Transparent. `apply_global_ops` rebinds the module-level function pointer that both v4 and v5 MoE models import. No user action. |
> | `apply_veomni_fused_moe_patch` (legacy MoE dispatch) | ✅ | ✅ | Transparent. v4-era models (qwen2_moe, pre-migration qwen3_moe) keep reading the same global `_fused_moe_forward`; the patch now runs before model load to fix pre-init ordering. |
> | Liger RMSNorm / RoPE / SwiGLU via `device_patch.py` `apply_per_model_patches` | ✅ | ✅ | Transparent for models whose `device_patch.py` registers the op. v4 models still replace HF classes by name; v5 models do the same before the patched modeling file is imported. |
> | `moe_implementation` value rename + `fused_npu` added | ✅ | ✅ | User-visible **breaking** config change. `"fused"` is renamed to `"fused_triton"`, `"fused_npu"` is added to replace the silent NPU auto-pick, and the default changes from `None` to `"eager"`. Configs using `moe_implementation: fused` must migrate to `fused_triton` (GPU) or `fused_npu` (NPU); `fused_quack` is unchanged. |
> | `KERNEL_REGISTRY` + `OpSlot` dispatch (rotary, MoE experts, CE OpSlots) | ❌ | ✅ | v4 models do not import generated `patched_modeling_*.py` and never see an `OpSlot`. For v4, the same kernels are still reachable via the `LOSS_MAPPING` / GLOBAL / legacy-MoE layers above. No user action needed, but the `cross_entropy_loss_implementation=liger_kernel` path on v4 goes through `LOSS_MAPPING` (wrapper-level), not through the per-model OpSlot guard. |
> | Per-op config fields (`rms_norm_implementation`, `rotary_pos_emb_implementation`, `swiglu_mlp_implementation`, `cross_entropy_loss_implementation`, `load_balancing_loss_implementation`) | ✅ | ✅ | Transparent. The fields replace the removed `VEOMNI_USE_LIGER_KERNEL` / `USE_GROUP_GEMM` env vars; users of those env vars need to migrate. |
>
> **Summary for reviewers:** no v4 model requires a code change to keep working. The one v4-visible change is the `moe_implementation` value rename and the NPU explicit-opt-in — a breaking config change for any YAML using `moe_implementation: fused`, which must migrate to `fused_triton` (GPU) or `fused_npu` (NPU); `fused_quack` is unchanged. All new v5 machinery (generated patchgen files, `OpSlot` guards, per-model OpSlot registrations) is orthogonal to the v4 dispatch path.

- [ ] Read the [Contribute Guide](https://github.com/ByteDance-Seed/VeOmni/blob/main/CONTRIBUTING.md)
- [x] Applied pre-commit checks (`make quality` passes)
- [x] Added/updated documentation (`docs/design/unified_kernel_registry.md`, `docs/design/kernel_selection.md`)
- [x] Added tests to CI workflow (`tests/ops/test_kernel_registry_numerical.py`, `tests/ops/test_moe_hw_gate.py`, `tests/models/test_models_patch.py`; NPU workflow updated for `test_moe_hw_gate.py`)


TimYangst:
@gemini-code-assist please help to review the pr again

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -238,3 +239,33 @@ def install_loss_mapping() -> None:
...
+KERNEL_REGISTRY.register(
+    KernelSpec(
+        name="liger_kernel",
+        op_name="cross_entropy_loss",
+        variant="standard",
+        factory=_liger_fused_ce_factory,
+        hardware=HardwareRequirement(device_type="gpu"),
+        description="Liger fused linear cross-entropy loss (with label shifting and SP reduction)",
+    )
+)
```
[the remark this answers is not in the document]
TimYangst: Same above

veomni/ops/kernel_registry.py
```diff
@@ -0,0 +1,147 @@
...
+                if get_gpu_compute_capability() < self.min_compute_capability:
+                    return False
+            return True
+        elif self.device_type == "npu":
+            try:
+                import torch_npu  # noqa: F401
+
+                return True
+            except ImportError:
+                return False
```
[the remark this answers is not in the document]
TimYangst: Replaced the bare `try: import torch_npu` with `IS_NPU_AVAILABLE` (from `veomni.utils.device`, which calls `is_torch_npu_available()`). This also catches    
   dev boxes that have the package installed but no actual NPU device.

veomni/models/transformers/qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_gpu.py
```diff
@@ -2530,25 +2545,41 @@ def forward(
...
-                logits=logits,
-                labels=labels,
-                vocab_size=self.config.text_config.vocab_size,
-                hidden_states=hidden_states,
-                weights=self.lm_head.weight,
-                **kwargs,
-            )
+            # Modification: OpSlot guard for cross-entropy loss.
+            if veomni_cross_entropy_loss.has_kernel:
+                loss, logits = veomni_cross_entropy_loss(
```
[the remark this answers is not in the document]
TimYangst: Same above

veomni/models/transformers/qwen3/generated/patched_modeling_qwen3_gpu.py
```diff
@@ -555,14 +616,19 @@ def forward(
...
-            )
+            # Modification: OpSlot guard for cross-entropy loss.
+            if veomni_cross_entropy_loss.has_kernel:
+                loss, logits = veomni_cross_entropy_loss(
+                    logits=logits,
+                    labels=labels,
+                    num_labels=self.num_labels,
+                    hidden_states=hidden_states,
+                    weights=self.score.weight,
+                    **kwargs,
```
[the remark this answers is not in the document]
TimYangst: The OpSlot `("cross_entropy_loss", "standard")` was bound to `ForCausalLMLoss` and also used from `Qwen3ForSequenceClassification.forward`, which silently shifted classification labels and ignored `num_labels`. Fixed by splitting into two variants:                                                               
   - `variant="causal"` → `ForCausalLMLoss` (shifts labels, used by ForCausalLM / ForConditionalGeneration)                                                     
   - `variant="seq_cls"` → `ForSequenceClassificationLoss` (no shift, token-level labels)   

Added `cross_entropy_fn` parameter to `ForSequenceClassificationLoss` so the Liger fused kernel can be injected via `partial()`, paralleling the `ForCausalLMLoss` convention. Patchgen configs de

veomni/models/transformers/qwen3/generated/patched_modeling_qwen3_gpu.py
```diff
@@ -501,14 +556,19 @@ def forward(
...
-                logits=logits,
-                labels=labels,
-                vocab_size=self.config.vocab_size,
-                hidden_states=hidden_states,
-                weights=self.lm_head.weight,
-                **kwargs,
-            )
+            # Modification: OpSlot guard for cross-entropy loss.
+            if veomni_cross_entropy_loss.has_kernel:
+                loss, logits = veomni_cross_entropy_loss(
```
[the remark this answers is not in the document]
TimYangst: The OpSlot is bound to **VeOmni's** `ForCausalLMLoss` (in `veomni/ops/kernels/cross_entropy/__init__.py`, line 102: `return loss, logits`), not HuggingFace's. VeOmni's version returns a `(loss, logits)` tuple, so the unpacking is correct.

TimYangst:
@gemini-code-assist I've address the previous comments, please take another code review

veomni/models/transformers/qwen3_5_moe/qwen3_5_moe_gpu_patch_gen_config.py
```diff
@@ -748,25 +752,41 @@ def qwen3_5_moe_forconditional_generation_forward_patched(
...
+        # Modification: OpSlot guard for cross-entropy loss.
+        if veomni_causal_lm_loss.has_kernel:
+            loss, logits = veomni_causal_lm_loss(
+                logits=logits,
+                labels=labels,
+                vocab_size=self.config.text_config.vocab_size,
+                hidden_states=hidden_states,
+                weights=self.lm_head.weight,
+                **kwargs,
+            )
```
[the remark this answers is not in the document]
TimYangst: Same as above

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
```diff
@@ -194,14 +232,19 @@ def qwen3_forcausallm_forward_patched(
...
+        # Modification: OpSlot guard for cross-entropy loss.
+        if veomni_causal_lm_loss.has_kernel:
+            loss, logits = veomni_causal_lm_loss(
+                logits=logits,
+                labels=labels,
+                vocab_size=self.config.vocab_size,
+                hidden_states=hidden_states,
+                weights=self.lm_head.weight,
+                **kwargs,
+            )
```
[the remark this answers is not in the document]
TimYangst: VeOmni's ForCausalLMLoss signature takes num_items_in_batch: Optional[int] = None directly — there is no config parameter.

veomni/arguments/arguments_types.py
```diff
@@ -719,8 +727,43 @@ def __post_init__(self):
                 logger.info_rank0(f"Replacing attn_implementation from '{self.attn_implementation}' to '{new_impl}'")
                 self.attn_implementation = new_impl
 
+        self._bridge_legacy_moe_implementation()
         self._validate_implementations()
 
+    def _bridge_legacy_moe_implementation(self):
```
Luosuu: why we need this? will this apply to both v4 and v5?

Luosuu: moe_implementation: "fused" / "eager"
fused_moe_kernel: triton / quack

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -73,15 +74,15 @@ def ForCausalLMLoss(
     # Enable model parallelism
     shift_labels = shift_labels.to(device)
 
+    # Resolve the cross-entropy implementation: explicit arg > module global > eager fallback.
+    loss_func = cross_entropy_fn or _cross_entropy
```
Luosuu: this is hard to read. make this deterministic

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -171,15 +173,15 @@ def ForSequenceClassificationLoss(
     # Enable model parallelism
     target = target.to(device)
 
+    # Resolve the cross-entropy implementation: explicit arg > module global > eager fallback.
+    loss_func = cross_entropy_fn or _cross_entropy
```
Luosuu: same

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -238,3 +240,55 @@ def install_loss_mapping() -> None:
     LOSS_MAPPING["ForCausalLM"] = ForCausalLMLoss
     LOSS_MAPPING["ForConditionalGeneration"] = ForCausalLMLoss
     LOSS_MAPPING["ForSequenceClassification"] = ForSequenceClassificationLoss
+
+
+# ── OpSlot kernel registration ───────────────────────────────────────────────
+
+from ...kernel_registry import KERNEL_REGISTRY, HardwareRequirement, KernelSpec
```
Luosuu, line 317: please give some explanation here

veomni/ops/kernels/moe/__init__.py
```diff
@@ -93,3 +93,66 @@ def apply_veomni_fused_moe_patch(
...
     else:
         _fused_moe_forward = None
+
+
+# ── OpSlot kernel registrations ──────────────────────────────────────────────
+
+from ...kernel_registry import KERNEL_REGISTRY, HardwareRequirement, KernelSpec
+
+
+def _make_moe_experts_adapter(raw_forward):
```
Luosuu, line 114: please give more explaination

veomni/ops/dispatch.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class OpSlot:
+    """A named dispatch slot that can be bound to a kernel implementation."""
+
+    def __init__(self, op_name: str, variant: str):
+        self.op_name = op_name
+        self.variant = variant
+        self._kernel: Callable | None = None
+        self._bound = False
```
Luosuu: what this means

veomni/arguments/arguments_types.py
TimYangst: Updated.

veomni/ops/kernels/cross_entropy/__init__.py
TimYangst: Added a module-level docstring (lines 15–41) describing the two dispatch paths (LOSS_MAPPING vs KERNEL_REGISTRY) and the apply_ops_config contract, plus an install_loss_mapping docstring covering when it runs and why it's required before build_foundation_model. Let me know if a specific aspect is still unclear and I'll expand it.

veomni/ops/kernels/moe/__init__.py
TimYangst: Expanded the docstring to spell out both signatures side-by-side — the OpSlot shape is the HF experts forward(self, hidden_states, top_k_index, top_k_weights), the raw kernel shape is the flat tensor one, and the adapter pulls num_experts / gate_up_proj / down_proj off self so the OpSlot stays a drop-in for HF's forward.

veomni/ops/dispatch.py
TimYangst: The bind/_bound path got a docstring + rebind-collision warning in 5348d677. _bound=True now distinguishes "this slot has been resolved (possibly to eager/None)" from "never resolved" — which matters because two models sharing a generated modeling module share the same slot, and rebinding silently affects both.

veomni/ops/kernels/cross_entropy/__init__.py
TimYangst: Addressed in c0d38b0c. The 3-way runtime resolution (explicit arg > module global > eager fallback) is gone. cross_entropy_fn is now a keyword-only argument with no runtime decision inside the wrapper — the kernel is pre-bound once at install_loss_mapping / KERNEL_REGISTRY.resolve time via functools.partial. Call-site is deterministic: partial(ForCausalLMLoss, cross_entropy_fn=<impl>) is stored in LOSS_MAPPING["ForCausalLM"], and the wrapper just calls cross_entropy_fn(...) unconditionally.

veomni/ops/kernels/cross_entropy/__init__.py
TimYangst: Addressed in c0d38b0c — the ForSequenceClassificationLoss path received the same refactor. No runtime resolution inside the wrapper; the inner CE kernel is pre-bound via partial at install_loss_mapping / KERNEL_REGISTRY time.

veomni/models/auto.py
```diff
@@ -133,7 +133,29 @@ def build_foundation_model(
     Builds the foundation model.
 
     If weights_path is provided, it loads the pre-trained weights, otherwise it initializes weights.
+
+    Contract: ``apply_ops_config(ops_implementation)`` must run before this
```
Luosuu, line 160: what if we are building a model without any kernel patch?

docs/design/unified_kernel_registry.md
```diff
@@ -255,7 +255,7 @@ class OpsImplementationConfig:
     moe_implementation: Literal["eager", "fused"] = "eager"
     fused_moe_kernel: Literal["triton", "quack"] = "triton"
     cross_entropy_loss_implementation: str = "eager"
-    moe_load_balancing_loss_implementation: str = "eager"
```
Luosuu: do we have to change the name here?

veomni/arguments/arguments_types.py
```diff
@@ -719,8 +729,45 @@ def __post_init__(self):
                 logger.info_rank0(f"Replacing attn_implementation from '{self.attn_implementation}' to '{new_impl}'")
                 self.attn_implementation = new_impl
 
+        self._bridge_legacy_fused_quack()
         self._validate_implementations()
 
+    def _bridge_legacy_fused_quack(self):
```
Luosuu: it is fine to enforce the new style, not having to bridge the legacy ones

veomni/ops/kernels/moe/__init__.py
```diff
@@ -78,18 +78,91 @@ def apply_veomni_fused_moe_patch(
...
+    the OpSlot stays a drop-in replacement for the HF ``forward``.
+    """
+
+    def adapter(self, hidden_states, top_k_index, top_k_weights):
+        return raw_forward(
+            num_experts=self.num_experts,
+            routing_weights=top_k_weights.to(hidden_states.dtype),
+            selected_experts=top_k_index,
+            hidden_states=hidden_states,
+            fc1_1_weight=None,
```
Luosuu, line 137: does this mean we do not support split fc1_1 and fc1_2 anymore?

docs/design/unified_kernel_registry.md
TimYangst: This field was introduced in #651 and is already live in main, the pr didn't change it. 

I made an update to add all the flags that can be config there for users references.

veomni/models/auto.py
TimYangst: The code already handles the "user didn't call it" case at auto.py:148-157 — we detect get_ops_config() is None, log a warning, and install OpsImplementationConfig() defaults. 

The reason it's still needed even for "no kernel patch" is that install_loss_mapping has to re-install  LOSS_MAPPING["ForCausalLM"] with a partial(ForCausalLMLoss, cross_entropy_fn=eager_cross_entropy) wrapper — VeOmni's patched modeling passes   hidden_states= / weights= to self.loss_function(...), which HF's stock ForCausalLMLoss rejects (cross_entropy/__init__.py:247). So even the all-eager path needs the wrapper.

veomni/arguments/arguments_types.py
TimYangst: Done, cleaned the flag.

veomni/ops/kernels/moe/__init__.py
TimYangst: No, split is still supported everywhere it was before. This adapter only fronts the new v5 OpSlot path, and every v5 HF MoE module we migrate uses merged gate_up_proj. The underlying kernel (fused_moe_forward at moe/__init__.py:32, plus the Triton / Quack / NPU backends) still accepts bot split (fc1_1_weight, fc1_2_weight) and merged (fc1_1_2_weight) — see the split-vs-merged parity test at tests/ops/test_fused_moe_split_vs_merged.py.   
The v4 legacy caller veomni/models/transformers/qwen3_moe/modeling_qwen3_moe.py:108 still passes fc1_1_weight=self.gate_proj, fc1_2_weight=self.up_pro  and goes through the same kernel via the global _fused_moe_forward pointer.

veomni/trainer/vlm_trainer.py
```diff
@@ -141,9 +141,11 @@ def _build_model(self):
             torch_dtype="float32" if args.train.accelerator.fsdp_config.mixed_precision.enable else "bfloat16",
             attn_implementation=args.model.ops_implementation.attn_implementation,
             moe_implementation=args.model.ops_implementation.moe_implementation,
+            fused_moe_kernel=args.model.ops_implementation.fused_moe_kernel,
             init_device=args.train.init_device,
             encoder_data_balance=args.model.encoder_data_balance,
             encoder_data_balance_sorting_algo=args.model.encoder_data_balance_sorting_algo,
+            ops_implementation=args.model.ops_implementation,
```
Luosuu: remind me why we need one more here?

veomni/trainer/base.py
```diff
@@ -245,7 +245,9 @@ def _build_model(self):
             torch_dtype="float32" if self.args.train.accelerator.fsdp_config.mixed_precision.enable else "bfloat16",
             attn_implementation=self.args.model.ops_implementation.attn_implementation,
             moe_implementation=self.args.model.ops_implementation.moe_implementation,
+            fused_moe_kernel=self.args.model.ops_implementation.fused_moe_kernel,
             init_device=self.args.train.init_device,
+            ops_implementation=self.args.model.ops_implementation,
```
Luosuu: why we need this here?

veomni/ops/dispatch.py
```diff
@@ -0,0 +1,90 @@
...
+                "instance sharing this module will pick up the new binding."
+            )
+        self._kernel = KERNEL_REGISTRY.resolve(self.op_name, self.variant, impl_name)
+        self._bound = True
+        self._impl_name = impl_name
+
+    @property
+    def has_kernel(self) -> bool:
+        """``True`` when a non-eager kernel is bound."""
+        # TODO(compile): `has_kernel` + `__call__` go through Python attribute
```
Luosuu: this issue looks unacceptable

veomni/models/auto.py
```diff
@@ -58,6 +61,49 @@ def build_config(config_path: str, **config_kwargs) -> "PretrainedConfig":
     return get_model_config(config_path, trust_remote_code=trust_remote_code, **config_kwargs)
 
 
+def _apply_legacy_moe_patch(config, moe_implementation: str, fused_moe_kernel: str):
```
Luosuu: apply_moe_patch_transformers_v4?

docs/usage/arguments.md
```diff
@@ -136,7 +136,8 @@ without modifying the config class.
 | Field | Type | Default | Description |
 | --- | --- | --- | --- |
 | attn_implementation | `Optional[Literal[...]]` | `"flash_attention_2"` | Attention implementation to use. |
-| moe_implementation | `Optional[Literal["eager", "fused", "fused_quack"]]` | `None` | MoE implementation: `eager` (reference loop), `fused` (Triton), `fused_quack` (Quack CUTLASS, SM90+). |
+| moe_implementation | `Literal["eager", "fused"]` | `"eager"` | MoE experts forward mode. `fused` binds a fused kernel (selected via `fused_moe_kernel`); `eager` runs the reference loop. |
```
piyifan123: can we use just one moe_implementation arg like: eager, fused_triton, fused_quack to be aligned with other ops?

veomni/arguments/arguments_types.py
```diff
@@ -721,6 +730,20 @@ def __post_init__(self):
 
         self._validate_implementations()
 
+    def resolve_impl_name(self, op_name: str) -> str:
```
piyifan123: if we merge moe into one arg (like the other comment suggests), we can remove this?

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
```diff
@@ -37,55 +37,90 @@
...
     k: torch.Tensor,
     cos: torch.Tensor,
     sin: torch.Tensor,
     position_ids: Optional[torch.Tensor] = None,
     unsqueeze_dim: int = 1,
 ) -> tuple[torch.Tensor, torch.Tensor]:
-    from liger_kernel.transformers.rope import liger_rotary_pos_emb
+    # Modification: OpSlot guard — use fused RoPE kernel when bound.
+    if veomni_apply_rotary_pos_emb.has_kernel:
+        return veomni_apply_rotary_pos_emb(q, k, cos, sin, position_ids=position_ids, unsqueeze_dim=unsqueeze_dim)
```
piyifan123: why do we need to pass down position_ids down?

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
```diff
@@ -37,55 +37,90 @@
...
-        cos,
-        sin,
-        position_ids=position_ids,
-        unsqueeze_dim=unsqueeze_dim,
-    )
+
+# ── Qwen3Model.forward (SP support) ─────────────────────────────────────────
 
 
 @config.override_method("Qwen3Model.forward", description="Support SP in Qwen3Model.forward")
```
piyifan123: can we remove this patch? this seems redundant now?

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
```diff
@@ -37,55 +37,90 @@
...
+# ── RMSNorm (OpSlot guard, functional Liger kernel) ──────────────────────────
+
+
+@config.override_method(
+    "Qwen3RMSNorm.forward",
+    description="OpSlot guard for Liger fused RMSNorm (standard formulation)",
 )
+def qwen3_rmsnorm_forward_patched(self, hidden_states: torch.Tensor) -> torch.Tensor:
+    # Modification: OpSlot guard — use fused RMSNorm kernel when bound.
+    if veomni_rms_norm.has_kernel:
```
piyifan123: nit: can we name has_kernel to use_non_eager_impl to be more exact?

veomni/trainer/dit_trainer.py
```diff
@@ -272,6 +272,7 @@ def _build_model(self):
                 torch_dtype="float32" if args.train.accelerator.fsdp_config.mixed_precision.enable else "bfloat16",
                 attn_implementation=args.model.ops_implementation.attn_implementation,
                 moe_implementation=args.model.ops_implementation.moe_implementation,
+                fused_moe_kernel=args.model.ops_implementation.fused_moe_kernel,
```
piyifan123: once we merge the args into one, we can revert those?

veomni/models/auto.py
piyifan123: +1 simplify this.

veomni/models/auto.py
```diff
@@ -58,6 +61,49 @@ def build_config(config_path: str, **config_kwargs) -> "PretrainedConfig":
...
+        logger.warning_rank0("You are using eager moe implementation, expect this to be VERY SLOW!")
+        config._moe_implementation = "eager"
+    else:
+        config._moe_implementation = "fused"
+        from ..ops.kernels.moe import apply_veomni_fused_moe_patch
+
+        apply_veomni_fused_moe_patch(fused_moe_kernel=fused_moe_kernel)
+
+
+def _bind_veomni_ops(modeling_module, ops_config: OpsImplementationConfig) -> bool:
```
piyifan123, line 93: where do we check the kernel availability? for example, if i set moe_impl to fused_quack and run on NPU or A100 (fused_quack requires SM>=90 CUDA), where will I get an explicit failure to notify me the selection is not availlable in my current env?

veomni/trainer/vlm_trainer.py
TimYangst: Cleaned up the params.

veomni/trainer/base.py
TimYangst: Cleaned up.

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
TimYangst: Done

veomni/models/auto.py
TimYangst: Done

docs/usage/arguments.md
TimYangst: Done

veomni/arguments/arguments_types.py
TimYangst: Cleaned up.

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
TimYangst: position_ids is vestigial — a holdover from an older HF API. Stock HF in transformers v5 already dropped it from apply_rotary_pos_emb 
(new signature: (q, k, cos, sin, unsqueeze_dim=1)), and Liger's liger_rotary_pos_emb still accepts it only for signature compat but never reads it inside LigerRopeFunction.forward. The one call site in the generated Qwen3Attention.forward passes only (q, k, cos, sin), so nothing actually flows through this parameter. Removed.

veomni/models/transformers/qwen3/qwen3_gpu_patch_gen_config.py
TimYangst: Done

veomni/trainer/dit_trainer.py
TimYangst: Reverted.

veomni/models/auto.py
TimYangst: Both paths have explicit hardware gates; the check fires at build_foundation_model time, not at config construction.                    
                                                                                                                                        
- Legacy MoE models (qwen3_moe, deepseek_v3, qwen3_vl_moe, qwen3_omni_moe): apply_veomni_fused_moe_patch in veomni/ops/kernels/moe/__init__.py:64 — device type + is_quack_gemm_available() (which includes SM90 via is_sm90_or_above()).
- OpSlot MoE (qwen3_5_moe): KERNEL_REGISTRY.resolve → HardwareRequirement.is_satisfied() in kernel_registry.py:41-54 — device_type + min_compute_capability (70 for triton, 90 for quack).                                                                                   

On A100 (SM80) with moe_implementation=fused_quack: clean RuntimeError "Kernel 'quack' for op='moe_experts' requires device_type='gpu', 
compute_capability>=90, ...".
On NPU with fused_quack: RuntimeError "...device_type='gpu'...".                                                                        
                                                                                                                                        
Coverage: tests/ops/test_moe_hw_gate.py mocks the hardware-detection helpers so every (kernel × device × SM) combination is exercised on any CI host.                                                                                                                           
                                                                                                                                        
Note: there's no HW check at OpsImplementationConfig._validate_implementations time — only package-requires (liger_kernel, torch_npu)  gets validated early.

veomni/ops/dispatch.py
TimYangst: Measured on a tiny Qwen3 (2 layers, hidden=64) with torch._dynamo.explain under torch 2.9:                                                
 ```
 Mode                       graphs  breaks                                                                                        
 HF baseline                    1        0                                                                                    
VeOmni, OpSlot → eager         1        0                                                                                                                                                                                        
VeOmni, OpSlot → liger         5       4                                                                                              
```
                                   
  - With OpSlot bound to eager, Dynamo constant-folds use_non_eager_impl and DCE's the guard branch — matches the HF baseline exactly.      
  - With a real kernel bound, the 4 breaks all originate inside kernel bodies (Tensor.item() in Liger, a logger.warning_once in our CE wrapper), not from OpSlot's property or __call__.                                                                                         
                            
So the concern didn't materialize — dropped the TODO. 

If it regresses with a future Dynamo version we can wrap the bound kernel in torch.compiler.assume_constant_result inside bind().

File: veomni/ops/dispatch.py (hunk 13 above)
Comment (TimYangst) [reply veomni/ops/dispatch.py]: Cleaned up the parameter.

**Comment by TimYangst** [comment]
@gemini-code-assist please help to review the PR

File: veomni/models/transformers/qwen3/generated/patched_modeling_qwen3_gpu.py [hunk 29]
```diff
@@ -501,14 +546,19 @@ def forward(
...
+                    labels=labels,
+                    vocab_size=self.config.vocab_size,
+                    hidden_states=hidden_states,
+                    weights=self.lm_head.weight,
+                    **kwargs,
+                )
+            else:
+                logits = self.lm_head(hidden_states)
+                loss = self.loss_function(logits=logits, labels=labels, vocab_size=self.config.vocab_size, **kwargs)
         else:
```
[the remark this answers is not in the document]
TimYangst: Good catch — fixed. The eager fallback now unpacks: loss, _ = self.loss_function(...). Applied in qwen3_gpu_patch_gen_config.py and regenerated via make patchgen. The qwen3_5_moe ForConditionalGeneration path already unpacked correctly (see line 769).

veomni/models/transformers/qwen3/generated/patched_modeling_qwen3_gpu.py
```diff
@@ -555,14 +606,19 @@ def forward(
...
+                    labels=labels,
+                    num_labels=self.num_labels,
+                    hidden_states=hidden_states,
+                    weights=self.score.weight,
+                    **kwargs,
+                )
+            else:
+                logits = self.score(hidden_states)
+                loss = self.loss_function(logits=logits, labels=labels, num_labels=self.num_labels, **kwargs)
         else:
```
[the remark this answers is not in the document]
TimYangst: Fixed — same unpacking applied to the Qwen3ForSequenceClassification.forward eager fallback.

veomni/models/auto.py
```diff
@@ -58,6 +61,74 @@ def build_config(config_path: str, **config_kwargs) -> "PretrainedConfig":
...
+    """
+    if moe_implementation == "eager":
+        logger.warning_rank0("You are using eager moe implementation, expect this to be VERY SLOW!")
+        config._moe_implementation = "eager"
+        return
+
+    if not moe_implementation.startswith("fused_"):
+        raise ValueError(
+            f"Invalid moe_implementation: {moe_implementation!r}. "
+            "Expected one of: 'eager', 'fused_triton', 'fused_quack'."
```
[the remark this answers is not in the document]
TimYangst: Fixed. Also added "fused_npu" to the Literal[...] type hint on build_foundation_model's moe_implementation parameter (auto.py:149) for consistency.

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -12,35 +12,67 @@
...
+    shift_labels: torch.Tensor | None = None,
+    # `*,` marks everything below as keyword-only. HF calls this wrapper with
+    # positional args (logits, labels, vocab_size, ...); keeping `cross_entropy_fn`
+    # keyword-only guarantees the pre-bound kernel from `install_loss_mapping` /
+    # `KERNEL_REGISTRY` (via `functools.partial`) cannot be silently overwritten
+    # by a positional arg overflowing into this slot.
+    *,
+    cross_entropy_fn: Callable = eager_cross_entropy,
     **kwargs,
 ) -> torch.Tensor:
```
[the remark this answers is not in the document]
TimYangst: Fixed — return annotation updated to tuple[torch.Tensor, torch.Tensor | None].

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -105,8 +128,13 @@ def ForSequenceClassificationLoss(
...
+    num_items_in_batch: int | None = None,
     ignore_index: int = -100,
+    # `*,` marks `cross_entropy_fn` keyword-only — same reason as in
+    # `ForCausalLMLoss`: the inner kernel is bound once at install time via
+    # `partial(..., cross_entropy_fn=...)` and must not be reachable via positional args.
+    *,
+    cross_entropy_fn: Callable = eager_cross_entropy,
     **kwargs,
 ) -> torch.Tensor:
     r"""
```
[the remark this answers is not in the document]
TimYangst: Fixed — annotation is now tuple[torch.Tensor, torch.Tensor | None] (the second element can be None when the fused path is taken without materialised logits; otherwise it's a tensor).

veomni/ops/kernels/cross_entropy/liger.py
```diff
@@ -30,6 +30,25 @@ def fused_liger_kernel_cross_entropy(
     shift_labels: Optional[torch.Tensor] = None,
     **kwargs,
 ) -> torch.Tensor:
```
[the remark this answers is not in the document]
TimYangst: Fixed — annotation updated to tuple[torch.Tensor, Optional[torch.Tensor]].

Luosuu (approved):

veomni/ops/kernels/cross_entropy/chunk_loss.py
```diff
@@ -26,6 +28,7 @@
 import torch.nn.functional as F
 
 from ....distributed.parallel_state import get_parallel_state
+from .eager import eager_cross_entropy
 
 
 class ChunkLoss(torch.autograd.Function):
```
FoolPlayer, line 39: It  seems that chunk_loss for NPU is not register, the config `ops_implementation.cross_entropy_loss_implementation=npu` will faile for NPU

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -73,16 +105,7 @@ def ForCausalLMLoss(
     # Enable model parallelism
     shift_labels = shift_labels.to(device)
```
FoolPlayer, line 106: double `shift_labels` args here

veomni/ops/kernel_registry.py
```diff
@@ -0,0 +1,145 @@
...
+            if self.min_compute_capability is not None:
+                if get_gpu_compute_capability() < self.min_compute_capability:
+                    return False
+            return True
+        elif self.device_type == "npu":
+            # IS_NPU_AVAILABLE == is_torch_npu_available(): requires both the
+            # torch_npu package AND an actual NPU device (unlike a bare import
+            # check, which passes on dev boxes that merely have the library).
+            return IS_NPU_AVAILABLE
+        return False
```
FoolPlayer: Maybe raise error here for unknown device type is better

tests/ops/test_kernel_registry_numerical.py
```diff
@@ -0,0 +1,131 @@
...
+    sin = sin.unsqueeze(unsqueeze_dim)
+    return (q * cos) + (_rotate_half(q) * sin), (k * cos) + (_rotate_half(k) * sin)
+
+
+def test_rms_norm_standard_liger_matches_eager():
+    pytest.importorskip("liger_kernel")
+    slot = _fresh_slot("rms_norm", "standard", "liger_kernel")
+    x = torch.randn(2, 16, 128, device=DEVICE, dtype=torch.bfloat16)
+    w = torch.randn(128, device=DEVICE, dtype=torch.bfloat16)
+    assert torch.allclose(slot(x, w, 1e-6), _eager_rms_norm_standard(x, w, 1e-6), atol=1e-2, rtol=1e-2)
```
FoolPlayer: It `atol=1e-2, rtol=1e-2` a bit high for bitwise check? Maybe we can change it in next PR.

veomni/models/transformers/qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_npu.py
```diff
@@ -111,6 +110,15 @@
 fused_recurrent_gated_delta_rule = None
 causal_conv1d_update = None
 
+# ── OpSlot declarations ──────────────────────────────────────────────────
+# These are bound at model-build time by _bind_veomni_ops() in auto.py.
+from veomni.ops.dispatch import OpSlot
+
+
+veomni_moe_experts_forward = OpSlot("moe_experts", "standard")
+veomni_causal_lm_loss = OpSlot("cross_entropy_loss", "causal")
```
FoolPlayer, line 119: We register OpSlot for causal_lm_loss but we didn't use it in ForCausalLM class.

veomni/models/transformers/qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_npu.py
FoolPlayer: Here, we didn't use `veomni_causal_lm_loss`

docs/get_started/installation/install_ascend_arm.md
```diff
@@ -38,6 +38,13 @@ To enable the NPU chunked cross-entropy loss, set
 `model.ops_implementation.cross_entropy_loss_implementation: npu` in your training YAML
 (replaces the legacy `VEOMNI_ENABLE_CHUNK_LOSS` environment variable).
 
+> **Note:** The NPU chunked cross-entropy applies only to `ForCausalLM`.
+> Models whose loss type is `ForConditionalGeneration` (VLMs) or
```
piyifan123: curious on why ForConditionalGeneration (VLM head class) cannot work?

TimYangst: It wasn't a fundamental block, just two missing pieces. Fixed in this PR instead of documenting a limitation:
1. SP reduction — chunk_loss_function previously returned a scalar without reduce_sequence_parallel_loss, whereas ForCausalLMLoss applies the SP reduction. Added the reduction inside chunk_loss_function itself, gated on get_parallel_state().sp_enabled, using the pre-shift labels for num_valid_tokens.
2. Routing — install_loss_mapping("npu") now also binds LOSS_MAPPING["ForConditionalGeneration"] = chunk_loss_function, so VLMs pick it up automatically.
  
ForSequenceClassification still stays on eager — it uses token-level labels with no causal shift, but chunk_loss_function hard-codes labels[..., 1:]. Docs rewritten to say "chunk_loss hard-codes the causal shift" (accurate) instead of "causal-only and no SP reduction" (stale).

veomni/models/transformers/qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_npu.py
TimYangst: Same as above.

veomni/models/transformers/qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_npu.py
TimYangst: Thanks for catching it, updated all the models.

tests/ops/test_kernel_registry_numerical.py
TimYangst: Tightened in this PR where the math allows it: 
- rms_norm standard (bf16 Liger) ,  1e-2 ->  2e-3       ,  single-pass reduction + elementwise mul
- rms_norm qwen3_5 (fp32 compare),  1e-2 ->  1e-4       ,  both sides up-cast to fp32 before comparison
- swiglu_mlp (bf16 Liger)        ,  1e-2 ->  5e-3       ,  3 matmuls stack bf16 rounding
- rotary_pos_emb (bf16 Liger)    ,  1e-2 ->  1e-2 (kept),  see below

RoPE stays at 1e-2: (q*cos) + (rotate_half(q)*sin) is 2 muls + 1 add in bf16, which can stack to ~3 ULPs — real max abs diff on random seeds is ~3e-2 near values of magnitude 1, and anything tighter than 1e-2 flakes. Proper tightening would switch that test to a fp32 reference comparison; filed as follow-up rather than loosening back again here.

veomni/ops/kernel_registry.py
TimYangst: Done

veomni/ops/kernels/cross_entropy/__init__.py
TimYangst: Confirmed — positional 6 was dead. Both eager_cross_entropy and fused_liger_kernel_cross_entropy only use positional 2 (labels, which is already the shifted tensor at that point) and never reference the positional 6 shift_labels. Dropped the duplicate in both ForCausalLMLoss and ForSequenceClassificationLoss call sites.

Leaving the unused shift_labels kwarg on the inner-kernel signatures themselves as a follow-up cleanup so this PR stays focused.

veomni/ops/kernels/cross_entropy/chunk_loss.py
TimYangst: Good catch — the OpSlot path would KeyError on NPU. Fixed by registering chunk_loss_function into KERNEL_REGISTRY as `(op_name="cross_entropy_loss", variant="causal", name="npu")`  alongside the existing liger_kernel entry, gated by HardwareRequirement(device_type="npu")
see the bottom of file:  veomni/ops/kernels/cross_entropy/__init__.py

No seq_cls variant: chunk_loss_function hard-codes the causal labels[..., 1:] shift and cannot back token-level classification losses — those stay on the eager wrapper via install_loss_mapping. KERNEL_REGISTRY.list_available("cross_entropy_loss", "causal") now returns ["liger_kernel", "npu"].

TimYangst:
@gemini-code-assist please help to take a review on the PR

veomni/ops/kernels/cross_entropy/chunk_loss.py
```diff
@@ -85,19 +93,20 @@ def chunk_loss_function(
...
 ) -> torch.Tensor:
-    if not get_parallel_state().sp_enabled:
+    sp_enabled = get_parallel_state().sp_enabled
+    # Snapshot the pre-shift labels for the SP denominator — the non-SP branch
+    # below rewrites `labels` in place with the shifted view.
+    sp_reduction_labels = labels
+
+    if not sp_enabled:
         labels = labels[..., 1:].contiguous()
         hidden_states = hidden_states[..., :-1, :].contiguous()
```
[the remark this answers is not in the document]
TimYangst: The reference ForCausalLMLoss in veomni/ops/kernels/cross_entropy/__init__.py:86-97 does not always
perform the shift — it skips it in the SP branch for the exact same reason chunk_loss_function does:
```
if not sp_enabled:
    if shift_labels is None:
        labels = nn.functional.pad(labels, (0, 1), value=ignore_index)
        shift_labels = labels[..., 1:].contiguous()
else:
    # "labels have been shifted in dataloader when `sp_enabled=True`, ignore shift_labels."
    shift_labels = labels
```
VeOmni's Ulysses SP dataloader pre-shifts labels before dispatching (see the warning message on line 96),
so both the eager CE path and chunk_loss_function rely on that invariant. The SP branch receiving
unshifted labels would already be incorrect by the wrapper's contract. This is not a divergence from the
reference.

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -12,35 +12,67 @@
...
+    # positional args (logits, labels, vocab_size, ...); keeping `cross_entropy_fn`
+    # keyword-only guarantees the pre-bound kernel from `install_loss_mapping` /
+    # `KERNEL_REGISTRY` (via `functools.partial`) cannot be silently overwritten
+    # by a positional arg overflowing into this slot.
+    *,
+    cross_entropy_fn: Callable = eager_cross_entropy,
     **kwargs,
-) -> torch.Tensor:
-    # pop fused loss kwargs
+) -> tuple[torch.Tensor, torch.Tensor | None]:
```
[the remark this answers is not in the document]
TimYangst: Correct observation, and intentional. The tuple is load-bearing: Liger's fused linear+CE and NPU chunk_loss_function fuse the lm_head projection into the loss, so the kernel — not the caller — is where logits come out. Returning only loss would force the caller to re-project lm_head(hidden_states) to get logits, defeating the purpose of the fusion.
 
Contract (per docs/design/kernel_selection.md and this PR's v4/v5 impact table): 
1. install_loss_mapping is installed once per process from BaseTrainer._build_model → apply_ops_config. Standalone callers get the warning installed in build_foundation_model (veomni/models/auto.py:176). 
2. Every v5 patched modeling file in-tree (qwen2 / qwen2_vl / qwen2_5_vl / qwen3 / qwen3_moe / qwen3_5 / qwen3_5_moe / qwen3_vl / qwen3_vl_moe / qwen3_omni_moe / glm_moe_dsa, GPU+NPU) now unpacks loss, _ = self.loss_function(...) or loss, logits = self.loss_function(...) — grep confirms zero loss = self.loss_function( call sites in our tree. 
3. Mixed mode (VeOmni install_loss_mapping applied + unpatched HF LlamaForCausalLM.forward) is unsupported by design — as the "BaseTrainer contract" section of the docs states, you are expected to run through BaseTrainer so every model in the process is patched coherently.
 
Added a short note to the install_loss_mapping docstring spelling out the tuple contract so this is explicit in the code, not just the doc.

veomni/ops/kernels/cross_entropy/chunk_loss.py
```diff
@@ -118,17 +127,10 @@ def ce_loss_func(hidden_states, weight, bias, labels, num_items_in_batch, ignore
...
-    ``chunk_loss_function``, which splits the sequence and calls the
-    currently-bound ``_cross_entropy`` on each chunk.
-    """
-    from transformers.loss.loss_utils import LOSS_MAPPING
-
-    LOSS_MAPPING["ForCausalLM"] = chunk_loss_function
+    # Match ``ForCausalLMLoss`` SP behavior so chunk_loss can back both
+    # ForCausalLM and ForConditionalGeneration heads when SP is enabled.
+    if sp_enabled:
+        num_valid_tokens = (sp_reduction_labels != ignore_index).sum()
```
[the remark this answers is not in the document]
TimYangst: The reference ForCausalLMLoss in __init__.py:119-122 does the same thing:                                 
```
if sp_enabled:                                                                                            
  num_valid_tokens = (labels != ignore_index).sum()
```                                                                                                           
In the SP branch, labels is the pre-shifted labels from the dataloader (length S) — the same tensor that chunk_loss_function snapshots into sp_reduction_labels. There is no length-S-1 shift_labels in the SP path; the shift happened upstream. The denominator here is consistent with the reference by construction.

FoolPlayer (approved):

PR #681, opened 2026-04-21
[parallel, perf] fix: sp gather && optimize: input embeds fusing

JorgenWan:
### What does this PR do?

1. Fix the default `scale_grad` (`True -> False`) value of `gather_outputs`. 

`gather_outputs` is used in the models: `wan, flux, qwen25omni, qwen3omnimoe, qwen3vl, qwen3vlmoe`. 

When enable `scale_grad`, the `grad_output = grad_output * ctx.seq_world_size`. 

As `gather_outputs` is used with `slice_input_tensor` in most models (except that `wan` model uses `slice_input_tensor_scale_grad`), `slice_input_tensor` doesn't  scale grad (`grad_output = grad_output / ctx.seq_world_size`), we should not scale grad by default.

Add `dist.all_reduce(grad_output, op=dist.ReduceOp.SUM, xxx)` in `backward`, as `gather_outputs` needs the grad of other ranks.

2. When hidden dim of input embeds is not divisible by sp_size, `gather_seq_scatter_heads` and `gather_heads_scatter_seq` can not work. 

One way is to pad the hidden dim. Another way is to substitute the `gather_seq_scatter_heads` and `gather_heads_scatter_seq` with `gather_outputs` and `slice_input_tensor`. 

The second method can optimize the input embeds fusing process by substituting `(n+2) x all2all` (`n` is the number of modality, e.g. image+video -> `n=2`) with `(n+1) x all_gather`, noting that `slice_input_tensor` includes no comm.

Although the second method uses more cuda memory, the estimated maximum usage is around `bsz x seq_len x dim ~ 0.1g` for `128k` sequence with `dim=1024`. This is acceptable.

3. Add grad unit test for gather_outputs. When `dist.all_reduce(grad_output, xxx)` is not used in `backward`, the grad will be zero, which is not expected.

### Checklist Before Starting

  - **Allowed modules:** `agent`, `ci`, `ckpt`, `config`, `data`, `dist`, `docker`, `docs`, `logging`, `misc`, `model`, `omni`, `optim`, `ops`, `parallel`, `perf`, `release`, `task`, `trainer`

### Test

> Validation results (training curves, eval metrics) for changes not covered by CI.

### API and Usage Example

### Design & Code Changes

- Applied pre-commit checks
- Added/updated documentation


Luosuu:
@JorgenWan Please give some context or unit tests to explain?

veomni/distributed/sequence_parallel/data.py
```diff
@@ -105,7 +105,7 @@ def gather_outputs(
     gather_dim: int,
     padding_dim: Optional[int] = None,
     unpad_dim_size: Optional[int] = None,
-    scale_grad=True,
```
Luosuu, line 108: please given a unit test or inline comment to explain why this is an issue

JorgenWan: As gather_outputs is used with slice_input_tensor in most models (except that wan model uses slice_input_tensor_scale_grad), slice_input_tensor doesn't scale grad (grad_output = grad_output / ctx.seq_world_size), we should not scale grad by default.

FoolPlayer:
All V5 patch-gen models are still using `gather_seq_scatter_heads`.

veomni/models/seed_omni/modeling_seed_omni.py
```diff
@@ -150,8 +150,8 @@ def image_forward(self, inputs_embeds: torch.Tensor, decoder_inputs, **kwargs):
                     inputs_embeds
                 )
                 if get_parallel_state().sp_enabled:
-                    input_image_features = gather_seq_scatter_heads(
-                        input_image_features, seq_dim=0, head_dim=1, group=get_parallel_state().sp_group
+                    input_image_features = gather_outputs(
```
FoolPlayer, line 153: `gather_outputs` use `_Gather` in `veomni/distributed/sequence_parallel/ulysses.py` . FWD use all-gather but _Gather.backward is split + take local chunk (no cross-rank reduce). For Image_embed we use `gather_outputs -> masked_scatter`, if the embeds are scattered to different ranks, `_Gather.backward` will ignore the contribute from other ranks.

JorgenWan: Yes. It seems that `_Gather.backward` is wrong. The bwd of `all_gather` should be `all_reduce+split` or  `reduce_scatter`.

JorgenWan: If we need the grad of other ranks

FoolPlayer:
Backward fix looks correct 👍
Some follow-ups before merge:
1. Patchgen for v5 model is still using gather_seq_scatter_heads, need to update `*_patch_gen_config.py`
2. Please add a small sp unit test to verify the case for image across rank and not.
3. All-gather + split is 2x comm workload compare with reduce_scatter, can we use reduce_scatter?

JorgenWan:
> Backward fix looks correct 👍 Some follow-ups before merge:
> 
> 1. Patchgen for v5 model is still using gather_seq_scatter_heads, need to update `*_patch_gen_config.py`
> 2. Please add a small sp unit test to verify the case for image across rank and not.
> 3. All-gather + split is 2x comm workload compare with reduce_scatter, can we use reduce_scatter?

OK. `torch.distributed.reduce_scatter` does not support different size of tensor, so use all_reduce .

veomni/models/transformers/qwen2_vl/qwen2_vl_gpu_patch_gen_config.py
```diff
@@ -317,18 +317,14 @@ def qwen2vl_model_forward_patched(
...
 
     if pixel_values is not None:
         image_embeds = self.get_image_features(pixel_values, image_grid_thw, return_dict=True).pooler_output
 
         # VeOmni SP
         if get_parallel_state().sp_enabled:
-            image_embeds = gather_seq_scatter_heads(
-                image_embeds, seq_dim=0, head_dim=-1, group=get_parallel_state().sp_group
-            )
+            image_embeds = gather_outputs(image_embeds, gather_dim=1, group=get_parallel_state().sp_group)
```
FoolPlayer: Maybe dim=0? The shape for image_embeds is `N, hidden_size`, maybe the split should on the first dimension? BTW, it seems that out CI don't cover that, only have single rank vit model test, maybe we can add ci for this.

JorgenWan: This should be dim=0.

Coach257 (approved):

FoolPlayer (approved):

PR #682, opened 2026-04-22
[misc] feat: support Python 3.12 in addition to 3.11

Luosuu:
### What does this PR do?

Allow VeOmni to be installed on Python 3.12 in addition to 3.11. Defaults stay on 3.11 — `uv sync` with no `--python` flag still picks 3.11 — but downstream projects that need 3.12 (e.g. integrations with packages pinned to `requires-python = ">=3.12, <3.13"`) can now share a single VeOmni source via `uv sync --python 3.12 --extra gpu`.

### Checklist Before Starting

- Search for relative PRs/issues and link here: N/A — request from a downstream consumer that needs to pull VeOmni into a 3.12 environment.
- PR title follows `[{modules}] {type}: {description}` format ✅ (using `misc` module since this is a packaging/build change).

### Test

- `uv lock --python 3.12` produces a universal lockfile that resolves cleanly.
- `uv sync --extra gpu --python 3.11` and `uv sync --extra gpu --python 3.12` both succeed and install fa2 + fa3 + torch 2.9.1+cu129 wheels.
- `python -c "import flash_attn"` succeeds on both interpreters: fa2 2.8.4 on 3.11, fa2 2.8.3 on 3.12 — both built against torch 2.9.1+cu129, no ABI mismatch.

### API and Usage Example

No API changes. The same `uv sync --extra gpu` flow works on either Python version; pass `--python 3.12` explicitly to opt into 3.12.

### Design & Code Changes

`pyproject.toml` only:

- `requires-python`: `>=3.11, <3.12` → `>=3.11, <3.13`.
- `[tool.uv.sources].torch`: added cp312 cu129 wheel URL gated by `marker = "python_version == '3.12'"`; existing cp311 URL gated with the matching 3.11 marker.
- `[tool.uv.sources].flash-attn`: added cp312 wheel built against torch 2.9 (from https://github.com/Luosuu/flash-attention-prebuild-wheels); existing cp311 URL gated.
- `[tool.uv.sources].av`: gated cp311 URL with `python_version == '3.11'` (av 14.4.x has no cp312 binary on pypi). The `audio`, `video`, and `dit` extras' `av>=14.3.0,<15.0` requirement now carries the same marker so 3.12 users opting into those extras get a clean resolution failure rather than an attempted source build.

flash-attn-3 and flash-attn-4 unchanged: fa3 ships as cp39-abi3 (Python stable ABI, works on both); fa4 is a git source.

- Pre-commit checks: not applicable for pyproject-only changes
- Documentation: no API changes; no docs touched
- Doc task paths: no `tasks/` files moved
- CI: no new tests needed — packaging-only change, validated via downstream e2e (vexact RL rollout passed on both 3.11 and 3.12 venvs)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ name = "veomni"
 dynamic = ["version"]
 description = "VeOmni: Scaling any Modality Model Training to any Accelerators with PyTorch native Training Framework"
 readme = "README.md"
-requires-python = ">=3.11, <3.12"
+requires-python = ">=3.11, <3.13"
 license = { file = "LICENSE" }
 authors = [
   { name="ByteDance-Seed", email="doubao-llm@bytedance.com" },
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,9 +89,7 @@ gpu = [
   "nvidia-cusparselt-cu12",
   "nvidia-nccl-cu12",
   "liger-kernel",
-  # flash-attn v2 has no torch-2.9-compatible cp312 wheel; users on Python 3.12 must
-  # either skip fa2 or supply their own source.
-  "flash-attn; python_version == '3.11'",
+  "flash-attn",
   "flash-attn-3",
   "hf_transfer",
   "torchcodec==0.8.1", # 0.8.1 is compatible with torch 2.9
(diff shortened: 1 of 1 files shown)


Luosuu:
Reopening with corrected branch name (upgrade-to-python-3-12) and uv.lock included.

PR #693, opened 2026-04-26
[model] fix: read tie_word_embeddings from inner text/decoder config

TimYangst:
### What does this PR do?

Fixes a silent-correctness bug where `CustomizedModelingLoader` (and `post_process_after_weight_loading`) force-tied `lm_head.weight` onto `embed_tokens.weight` after weight loading even when the checkpoint stored them as distinct tensors -- producing wrong logits and inflated training loss for nested ImageTextToText / multimodal models.

#### Root Cause

`PretrainedConfig.tie_word_embeddings` defaults to `True` if not set. For nested ImageTextToText configs the OUTER config typically does not set this flag explicitly, so it returns `True` from `getattr(config, "tie_word_embeddings", True)` -- even though the actual lm_head/embed tying is governed by the INNER decoder config (e.g. `text_config`)

Two sites in VeOmni read the OUTER flag and force-tie:

- `veomni/models/loader.py:258` (post-load fixup in `CustomizedModelingLoader._build_model`)
- `veomni/models/module_utils.py:508` (`post_process_after_weight_loading`)

HF's own `PreTrainedModel.tie_embeddings_and_encoder_decoder` reads `self.config.get_text_config(decoder=True).tie_word_embeddings`, which returns the inner decoder config for nested layouts and falls back to `self` for plain text configs -- this is exactly the lookup we should mirror.

### Checklist Before Starting

- Search for relative PRs/issues and link here: N/A

### Test

Verified end-to-end on an 8x A100 80GB devbox with `transformers==4.57.3`:

**A/B reproducer** (single GPU, no FSDP, same batch / weights / seed for both HF and VeOmni paths):

```
Before fix:
  HF        lm_head.weight std=0.026251  loss=1.403916
  VeOmni    lm_head.weight std=0.028238  loss=4.194746  (tied to embed)
  ratio = 2.9879x

After fix:
  HF        lm_head.weight std=0.026251  loss=1.403916
  VeOmni    lm_head.weight std=0.026251  loss=1.403916  (untied)
  ratio = 1.0000x  (bit-identical)
```

**End-to-end FSDP2 vs HF equivalence** (8 GPU, the integration test that originally caught this):

### API and Usage Example

> N/A -- internal loader fix, no API surface change.

### Design & Code Changes

**1. `veomni/models/loader.py`** -- replace `getattr(model.config, "tie_word_embeddings", True)` with `getattr(model.config.get_text_config(decoder=True), "tie_word_embeddings", True)`. Comment explains the nested-config rationale and points at HF's matching API.

**2. `veomni/models/module_utils.py`** -- same change, same rationale, in `post_process_after_weight_loading` (the FSDP-context tying path).

**Not changed:** `veomni/models/seed_omni/auto.py:193` already reads `model.foundation.config.tie_word_embeddings` -- the SeedOmni `foundation` is a text-only LLM submodel by construction (see `veomni/models/seed_omni/foundation/base.py`), so its config is the inner decoder config already. No nested-multimodal vulnerability at that site.

**Behavior on plain text models (Qwen3, Llama, ...):** `config.get_text_config(decoder=True)` returns `self` for non-nested configs, so the flag read is identical to before. No regression risk for the dense / MoE LLM paths.

**Audit of behavior change** (across all 28 configs registered in `MODEL_CONFIG_REGISTRY`, plus HF-only fallbacks for VL models without a VeOmni subclass):

| Model | outer | inner | Old behavior | New behavior | Verdict |
|---|---|---|---|---|---|
| `cosmos` (`CosmosConfig`, SeedOmni video decoder) | `True` | `None` | try-tie -> caught + logged (the cosmos module has no LM-style input/output embeddings) | skip tie | strict improvement (drops a known-failing tying attempt) |
| All 26 other registered configs | `outer == inner` | -- | unchanged | unchanged | no-op |
| HF-fallback multimodal (`qwen3_vl`, etc.) | `True == True` | -- | unchanged | unchanged | no-op |
| Plain text (`qwen3 dense`, `qwen3 moe`, `llama`, ...) | `outer == inner` (config returns `self`) | -- | unchanged | unchanged | no-op |

No regression cases were found. In particular, no registered config has the `outer=False, inner=True` shape that would silently start tying after the fix.

- [x] Applied pre-commit checks (`make style` + `make quality`)
- [x] Verified no behavior change for plain-text configs (`get_text_config(decoder=True)` returns `self`)
- [x] Confirmed `seed_omni/auto.py` is not affected (already reads inner foundation config)
- [ ] Add a regression test in `tests/` (e.g. a tiny untied-lm_head model checkpoint loaded under both paths and asserting `model.lm_head.weight is not model.get_input_embeddings().weight`) -- suggested follow-up if reviewers want coverage in the upstream test suite.

### DIFF
--- a/veomni/models/loader.py
+++ b/veomni/models/loader.py
@@ -254,8 +254,25 @@ class CustomizedModelingLoader(BaseModelLoader):
             if not empty_init:
                 load_model_weights(model, weights_path, init_device)
 
-            # we should tie embeddings after loading weights because init_empty_weights() leads to untied weights,
-            if getattr(model.config, "tie_word_embeddings", True):
+            # We should tie embeddings after loading weights because
+            # init_empty_weights() leads to untied weights.
+            #
+            # Read the tie flag from the inner text/decoder config rather
+            # than the outer model.config: nested ImageTextToText / multimodal
+            # configs (e.g. InternVLConfig) often leave the OUTER
+            # ``tie_word_embeddings`` at PretrainedConfig's True default while
+            # the actual lm_head <-> embed_tokens tying is governed by the
+            # INNER decoder config (e.g. ``text_config``). Reading the OUTER
+            # flag would force-tie even when the checkpoint stores
+            # ``lm_head.weight`` as a distinct tensor, silently clobbering it
+            # with the input embedding's value at load time.
+            # ``PretrainedConfig.get_text_config(decoder=True)`` returns the
+            # decoder text config for nested layouts and falls back to ``self``
+            # for plain text configs -- the same lookup HF's
+            # ``PreTrainedModel.tie_embeddings_and_encoder_decoder`` uses (see
+            # transformers ``modeling_utils.py``).
+            text_config = model.config.get_text_config(decoder=True)
+            if getattr(text_config, "tie_word_embeddings", True):
                 try:
                     input_embeddings = model.get_input_embeddings()
                     output_embeddings = model.get_output_embeddings()
--- a/veomni/models/loader.py
+++ b/veomni/models/loader.py
@@ -254,17 +254,21 @@ class CustomizedModelingLoader(BaseModelLoader):
             if not empty_init:
                 load_model_weights(model, weights_path, init_device)
 
-            # init_empty_weights() leaves embeddings untied; re-tie if requested.
-            # Check both outer and inner text configs: nested multimodal layouts
-            # can disable tying on either side (InternVL on inner, Qwen3VLMoe on
-            # outer with the inner missing the attribute entirely).
+            # init_empty_weights() leaves embeddings untied; re-tie only when
+            # the config asks for it. Nested multimodal layouts can disable tying
+            # on either side (InternVL on inner, Qwen3VLMoe on outer with inner
+            # silent), so AND both. Treat unset as True so a silent side does not
+            # override an explicit True, but require at least one side to set the
+            # flag -- if neither does, default to False (matches HF v5).
             text_config = (
                 model.config.get_text_config(decoder=True)
                 if hasattr(model.config, "get_text_config")
                 else model.config
             )
-            if getattr(model.config, "tie_word_embeddings", True) and getattr(
-                text_config, "tie_word_embeddings", True
+            if (
+                (hasattr(model.config, "tie_word_embeddings") or hasattr(text_config, "tie_word_embeddings"))
+                and getattr(model.config, "tie_word_embeddings", True)
+                and getattr(text_config, "tie_word_embeddings", True)
             ):
                 try:
                     input_embeddings = model.get_input_embeddings()
--- a/veomni/models/loader.py
+++ b/veomni/models/loader.py
@@ -276,7 +276,18 @@ class CustomizedModelingLoader(BaseModelLoader):
                 if hasattr(model.config, "get_text_config")
                 else model.config
             )
-            if getattr(text_config, "tie_word_embeddings", True):
+            # Prefer the inner text_config's flag when it is explicitly set
+            # (e.g. InternVL leaves the OUTER default at True while the INNER
+            # decoder config says False). When the inner text_config does not
+            # define the attribute (e.g. Qwen3VLMoeTextConfig under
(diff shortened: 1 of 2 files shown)


veomni/models/loader.py
```diff
@@ -254,8 +254,25 @@ def load_model(self, init_kwargs: dict, **kwargs):
...
+            # flag would force-tie even when the checkpoint stores
+            # ``lm_head.weight`` as a distinct tensor, silently clobbering it
+            # with the input embedding's value at load time.
+            # ``PretrainedConfig.get_text_config(decoder=True)`` returns the
+            # decoder text config for nested layouts and falls back to ``self``
+            # for plain text configs -- the same lookup HF's
+            # ``PreTrainedModel.tie_embeddings_and_encoder_decoder`` uses (see
+            # transformers ``modeling_utils.py``).
+            text_config = model.config.get_text_config(decoder=True)
+            if getattr(text_config, "tie_word_embeddings", True):
```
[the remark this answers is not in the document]
TimYangst: Not a bug: getattr(None, 'tie_word_embeddings', True) -> True

veomni/models/module_utils.py
```diff
@@ -503,9 +503,25 @@ def post_process_after_weight_loading(
...
+    # Reading the OUTER flag would force-tie even when the checkpoint stores
+    # ``lm_head.weight`` as a distinct tensor, silently clobbering it with
+    # the input embedding's value at load time.
+    # ``PretrainedConfig.get_text_config(decoder=True)`` returns the decoder
+    # text config for nested layouts and falls back to ``self`` for plain text
+    # configs -- the same lookup HF's
+    # ``PreTrainedModel.tie_embeddings_and_encoder_decoder`` uses (see
+    # transformers ``modeling_utils.py``).
+    text_config = model.config.get_text_config(decoder=True)
+    if getattr(text_config, "tie_word_embeddings", True):
```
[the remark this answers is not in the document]
TimYangst: Same.

veomni/models/loader.py
```diff
@@ -254,8 +254,25 @@ def load_model(self, init_kwargs: dict, **kwargs):
...
+            # INNER decoder config (e.g. ``text_config``). Reading the OUTER
+            # flag would force-tie even when the checkpoint stores
+            # ``lm_head.weight`` as a distinct tensor, silently clobbering it
+            # with the input embedding's value at load time.
+            # ``PretrainedConfig.get_text_config(decoder=True)`` returns the
+            # decoder text config for nested layouts and falls back to ``self``
+            # for plain text configs -- the same lookup HF's
+            # ``PreTrainedModel.tie_embeddings_and_encoder_decoder`` uses (see
+            # transformers ``modeling_utils.py``).
+            text_config = model.config.get_text_config(decoder=True)
```
Luosuu:
```python
text_config = (
      model.config.get_text_config(decoder=True)
      if hasattr(model.config, "get_text_config")
      else model.config
  )
```

would this be better?

TimYangst: Done

Luosuu:
please fix ci @TimYangst

veomni/models/module_utils.py
```diff
@@ -779,9 +779,14 @@ def post_process_after_weight_loading(
...
-    # except in the FSDP2 (swap tensor) context.
-    if getattr(model.config, "tie_word_embeddings", True):
+    # to_empty() leaves embeddings untied (except under FSDP2 swap-tensor);
+    # re-tie if requested. Check both outer and inner text configs: nested
+    # multimodal layouts can disable tying on either side (InternVL on inner,
+    # Qwen3VLMoe on outer with the inner missing the attribute entirely).
+    text_config = (
+        model.config.get_text_config(decoder=True) if hasattr(model.config, "get_text_config") else model.config
+    )
+    if getattr(model.config, "tie_word_embeddings", True) and getattr(text_config, "tie_word_embeddings", True):
```
Luosuu: maybe we prefer the default behavior is to set tie_word_embeedings as False when both model and text configs do not set it?

TimYangst: Done

Luosuu (approved):

PR #696, opened 2026-04-27
[model, ckpt] feat: runtime checkpoint tensor converter for transformers v4 MoE

Luosuu:
## Summary

- Adds a shared `MoEV4StackingConverter` and per-model v4 factories for `qwen3_moe`, `deepseek_v3`, and `qwen3_omni_moe` (thinker, fused mode) so vanilla HuggingFace MoE checkpoints load directly into VeOmni's three stacked `nn.Parameter`s at runtime — no more offline `scripts/moe_ckpt_merge/moe_merge.py` step before training on transformers 4.57.3.
- Mirrors the v5 converter introduced in #589 but emits three separate stacked tensors (`gate_proj`, `up_proj`, `down_proj`) instead of the v5 fused `gate_up_proj`, matching what `PatchQwen3MoeExperts` / `PatchDeepseekV3NaiveMoe` / `Qwen3OmniMoeThinkerExperts` actually allocate.
- The offline merge script gets a `DeprecationWarning` but is kept for users who want to amortize the stacking cost across many runs on very large checkpoints (e.g. Qwen3-235B).

## Why

Today, transformers v4 users have to run the offline merge script on every fresh HF checkpoint — extra disk I/O, extra storage, easy to forget. v5 users already get this for free via the converter protocol added in #589. The infrastructure (`module_utils.py`'s `get_checkpoint_tensor_converter` / `maybe_convert_checkpoint_tensor` / `converter.finalize` calls) is version-agnostic; only the per-model staticmethod attachment was missing for v4.

A prior attempt at this same problem — closed PR #67 *"qwen3-moe-a3b load"* — predated the converter protocol and shipped a custom load hook inside the model patch. This PR routes through the shared protocol instead.

## Scope

| Model | v4 path | Action |
|-------|---------|--------|
| `qwen3_moe` | `PatchQwen3MoeExperts` always uses stacked params | Always-on converter |
| `deepseek_v3` | `PatchDeepseekV3NaiveMoe` always uses stacked params; reads `n_routed_experts`; first `first_k_dense_replace` layers are dense | Always-on converter (regex misses dense layers naturally) |
| `qwen3_omni_moe` | Thinker uses stacked in fused mode, `nn.ModuleList` in eager; talker always `nn.ModuleList` | Converter only when `_moe_implementation == "fused"`, thinker prefix only |
| `qwen3_vl_moe` | Inherits HF's class with native `gate_up_proj [E, H, 2*I]` and transposes at runtime | **No converter needed** |
| `qwen3_5_moe`, `glm_moe_dsa` | Registered only on transformers ≥ 5.2.0 | N/A on v4 |

## Test plan

- [x] `pytest tests/models/test_moe_v4_runtime_merge.py -v` — 20 tests covering the shared base, per-model factories, multi-prefix dispatch, and an end-to-end `load_model_weights` integration that builds a tiny qwen3_moe model, writes per-expert HF safetensors, and verifies post-load `experts.{gate,up,down}_proj` exactly equals `torch.stack` of the inputs.
- [x] `pytest tests/models/test_checkpoint_tensor_converter.py` — 51 existing v5 converter tests pass unchanged (no protocol changes).
- [x] `pytest tests/models/test_model_registry.py` — registration regression check passes.
- [x] `make quality` — ruff check + format clean.
- [x] Offline `moe_merge.py` still runs and emits a `DeprecationWarning` on entry.
- [ ] Suggested follow-up: round-trip equivalence on a real `/mnt/hdfs/models/Qwen3-30B-A3B` checkpoint comparing direct load vs. pre-merged load.
- [ ] Suggested follow-up: 1-step training smoke run on a tiny MoE fixture.

## Risk notes

- Memory: the converter buffers up to `num_experts` per `(prefix, projection)` before stacking — same shape as the existing v5 converter that has been in production since #589.
- Broadcast load path (`module_utils.py:405-430`) already handles `None`-returns from buffered converters, so no extra work there.
- Safetensors key order across shards is not assumed; the converter is keyed by `(prefix, expert_id, projection)`.

### DIFF
--- a/docs/transformers_v5/transformers_v5_moe_weight_loading.md
+++ b/docs/transformers_v5/transformers_v5_moe_weight_loading.md
@@ -95,11 +95,28 @@ VeOmni keeps split expert tensors in patched modeling:
 This differs from native Transformers v5 `gate_up_proj` layout.
 
 Checkpoint loading behavior:
-- VeOmni does not do runtime remapping from legacy per-expert keys;
-- HuggingFace safetensor checkpoints commonly store expert weights in per-expert form.
+- HuggingFace safetensor checkpoints store expert weights in per-expert form
+  (`experts.{j}.gate_proj.weight`, etc.).
+- A runtime `CheckpointTensorConverter` is now registered on each v4 MoE model
+  class (`qwen3_moe`, `deepseek_v3`, and `qwen3_omni_moe` in fused mode), so
+  HF checkpoints load directly into VeOmni's three stacked nn.Parameters
+  without offline preprocessing. The shared converter lives at
+  `veomni/models/transformers/_moe_v4_converter.py`; per-model factories sit
+  alongside their v5 counterparts as `checkpoint_tensor_converter_v4.py`.
+- Output keys mirror what the offline merge script produces:
+  - `experts.gate_proj` `[E, I, H]`
+  - `experts.up_proj` `[E, I, H]`
+  - `experts.down_proj` `[E, H, I]`
+  Unlike the v5 converter, the v4 converter does **not** cat gate+up — v4
+  modeling holds them as separate parameters.
+- For Qwen3-Omni-MoE the converter only fires for the **thinker** tower and
+  only when `_moe_implementation == "fused"`. The talker tower (and the
+  thinker in eager mode) keep `nn.ModuleList` experts and load HF per-expert
+  keys natively.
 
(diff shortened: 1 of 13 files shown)


TimYangst (approved):
LGTM, however could you note somewhere or leave a TODO to instructs that how should we handle the format of the checkpoint after we done the training job?

PR #714, opened 2026-04-30
[ops, model] feat: dispatch Qwen3.5 linear-attention kernels via OpSlot

TimYangst:
### What does this PR do?

> Refactors the three Qwen3.5 GatedDeltaNet kernels — `FusedRMSNormGated`, `causal_conv1d`, `chunk_gated_delta_rule` — onto the OpSlot pattern so users pick the backend through `OpsImplementationConfig` instead of relying on import-time `try: from fla` silent fallback. Adds an opt-in `flash-qla` extra exposing the QwenLM FlashQLA implementation as an alternative `chunk_gated_delta_rule` backend.

### Checklist Before Starting

- Search for relative PRs/issues and link here: depends on the OpSlot dispatch pattern landed in #705

### Test

> - `make quality` — clean.
> - `pytest tests/models/test_models_patch.py -k "qwen3_5"` — both `qwen3_5` and `qwen3_5_moe` pass after opting the test fixture into the FLA backends (mirrors the prior implicit behaviour).
> - End-to-end smoke: `_bind_veomni_ops` resolves all three new slots from `OpsImplementationConfig` and rebinding from `eager` → `fla` swaps in the real `FusedRMSNormGated` / `causal_conv1d` / `chunk_gated_delta_rule` callables.

### API and Usage Example

> Three new fields on `OpsImplementationConfig` (default `eager`, which raises in the patched forward — varlen training requires opting into a non-eager backend):
>
> | field | values |
> |---|---|
> | `rms_norm_gated_implementation` | `eager` / `fla` |
> | `causal_conv1d_implementation` | `eager` / `fla` |
> | `chunk_gated_delta_rule_implementation` | `eager` / `fla` / `flash_qla` |
>
> Example yaml — opting in to FLA (the documented default for Qwen3.5 GPU training):
>
> ```yaml
> model:
>   ops_implementation:
>     attn_implementation: flash_attention_2
>     rms_norm_gated_implementation: fla
>     causal_conv1d_implementation: fla
>     chunk_gated_delta_rule_implementation: fla
> ```
>
> The provided example configs (`configs/text/qwen3_5_sft.yaml`, `configs/multimodal/qwen3_5{,_moe}/qwen3_5*_vl.yaml`) already set the three fields to `fla`.
>
> To use FlashQLA's chunk_gated_delta_rule (opt-in extra):
>
> ```bash
> uv sync --extra flash-qla --extra gpu --extra audio --no-group transformers-stable --extra transformers5-exp --group dev
> ```
>
> ```yaml
> chunk_gated_delta_rule_implementation: flash_qla
> ```

### Design & Code Changes

> - **`veomni/ops/kernels/linear_attention/__init__.py`** (new): registers `rms_norm_gated`, `causal_conv1d`, `chunk_gated_delta_rule` ops with the kernel registry. `fla` backends use lazy factories so hosts without `flash-linear-attention` can still load the module; `flash_qla` is a second registered backend for `chunk_gated_delta_rule` only.
> - **`veomni/arguments/arguments_types.py`**: three new fields on `OpsImplementationConfig` defaulting to `eager`, matching the no-silent-fallback policy (a misconfigured backend now raises with a clear fix instead of degrading silently).
> - **`veomni/models/auto.py`**: move `_bind_veomni_ops` from post-load to pre-init. GatedDeltaNet picks the impl in `__init__` (chooses `Qwen3_5RMSNormGated` vs `FusedRMSNormGated` based on `use_non_eager_impl`), so the slots must be bound before construction. Existing post-`__init__` consumers are unaffected — slots are module-level singletons.
> - **`veomni/models/transformers/qwen3_5{,_moe}/{gpu,npu}_patch_gen_config.py`** + regenerated `generated/`: replace the import-time `try: from fla import ... except ImportError` block with three OpSlot guards inside `Qwen3_5GatedDeltaNet.__init__` / `forward`. NPU patches register no non-eager backends today; selecting `fla` on NPU raises at bind time with a GPU-only message.
> - **`pyproject.toml`** + **`uv.lock`**: new `flash-qla` extra plus a `tool.uv.sources` entry pinned to the 2026-04-29 main commit (`c9b9827`).
> - **`tests/models/utils.py`**: `_build_ops_config_for_mode` opts qwen3_5/qwen3_5_moe tests into the FLA backends so the patch-equivalence tests still exercise the varlen path. Other models are unaffected — their generated modules don't have these OpSlots.
> - **`configs/` + `docs/examples/qwen3_5.md`**: update the three example yamls and add a "Selecting linear-attention kernels" section to the Qwen3.5 docs.

- [x] Added/updated documentation (`docs/examples/qwen3_5.md`, three example yamls)
- [x] Added tests to CI workflow (or explained why not feasible) — covered by existing `tests/models/test_models_patch.py`; the test fixture now exercises the FLA backends to keep the prior coverage of the GatedDeltaNet varlen path


veomni/ops/kernels/gated_delta_rule/__init__.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
piyifan123, line 1: nit: name the forlder as gated_delta_rule as there are many variations of linear attentions and qwen3.5 uses the particular one.

veomni/ops/kernels/linear_attention/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+    return chunk_gated_delta_rule
+
+
+KERNEL_REGISTRY.register(
+    KernelSpec(
+        name="flash_qla",
+        op_name="chunk_gated_delta_rule",
+        variant="standard",
+        factory=_flash_qla_chunk_gated_delta_rule_factory,
+        hardware=HardwareRequirement(device_type="gpu"),
```
piyifan123: note that the current support is only on SM90. we should add a check for that.

1. https://github.com/QwenLM/FlashQLA/issues/2
2. https://github.com/QwenLM/FlashQLA/blob/main/README.md?plain=1#L21

veomni/arguments/arguments_types.py
```diff
@@ -724,6 +724,36 @@ class OpsImplementationConfig:
             "'eager' (default) uses PyTorch reference."
         },
     )
+    rms_norm_gated_implementation: str = field(
+        default="auto",
```
piyifan123: did we say default should be a GPU kernel instead of the ambiguous "auto"?

TimYangst: Done

veomni/ops/kernels/gated_delta_rule/__init__.py
TimYangst: done

veomni/ops/kernels/linear_attention/__init__.py
TimYangst: Updated the condition.

Luosuu (approved):

PR #735, opened 2026-05-07
[model] fix: dit + lora related fix

Coach257:
## Dit trainer related:
1. support sp offline embedding
2. fix helper when token_len == 0 (offline embedding has no token)
## Lora related: (only dit tested currently)
1. fix fsdp1 meta/gpu init load lora_adapter & init lora
2. add fsdp2 meta init load peftmodel & load adapter & init lora
3. del `training architecture`, only keep `model.lora_config`
4. dcp shard save lora
## Others:
1. fix optimizer None state when some modules have no grad
2. add save utils for video with audios

### DIFF
--- a/tasks/deprecated_task/train_flux.py
+++ b/tasks/deprecated_task/train_flux.py
@@ -14,7 +14,7 @@ from transformers import (
 )
 
 from veomni.arguments import DataArguments, ModelArguments, TrainingArguments, parse_args, save_args
-from veomni.checkpoint import build_checkpointer
+from veomni.checkpoint import build_checkpointer, ckpt_to_state_dict
 from veomni.data.diffusion.data_loader import build_dit_dataloader
 from veomni.data.diffusion.dataset import build_text_image_dataset
 from veomni.distributed.clip_grad_norm import veomni_clip_grad_norm
@@ -127,10 +127,6 @@ class MyTrainingArguments(TrainingArguments):
     vit_lr: float = field(
         default=1e-6,
         metadata={"help": "Learning rate for visual encoder parameters."},
-    )
-    train_architecture: Literal["lora", "full"] = field(
-        default="full",
-        metadata={"help": "Model structure to train. LoRA training or full training."},
     )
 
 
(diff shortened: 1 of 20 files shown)


veomni/utils/lora_utils.py
```diff
@@ -97,3 +101,137 @@ def load_state_dict_from_bin(file_path, torch_dtype=None):
...
+        for piece in pieces:
+            if piece.startswith("lora_"):
+                break
+            lora_layer_fqn_parts.append(piece)
+            lora_layer = getattr(lora_layer, piece)
+        lora_layer_fqn = ".".join(lora_layer_fqn_parts)
+        if lora_layer_fqn not in _initialized and hasattr(lora_layer, "reset_lora_parameters"):
+            _initialized.add(lora_layer_fqn)
+            for adapter in getattr(lora_layer, "lora_A", {}).keys():
+                lora_layer.reset_lora_parameters(adapter, init_lora_weights=True)
```
FoolPlayer, line 160: Init lora per rank will use the same random seed, every rank will get the same initial weight. It's better just init on rank0 and shard to other ranks.

veomni/trainer/callbacks/checkpoint_callback.py
```diff
@@ -179,15 +181,14 @@ def _save_checkpoint(self, state: TrainerState, stage: str = "step_end"):
             self.trainer.checkpointer.save_future.result()
 
         if stage == "train_end":
-            del self.trainer.optimizer
-            del self.trainer.lr_scheduler
+            self.trainer.optimizer = None
```
FoolPlayer, line 184: Why change to `= None`

tasks/deprecated_task/train_flux.py
```diff
@@ -14,7 +14,7 @@
 )
 
 from veomni.arguments import DataArguments, ModelArguments, TrainingArguments, parse_args, save_args
-from veomni.checkpoint import build_checkpointer
+from veomni.checkpoint import build_checkpointer, ckpt_to_state_dict
```
FoolPlayer, line 17: Do we still use the deprecated tasks?

veomni/trainer/callbacks/checkpoint_callback.py
Coach257: https://github.com/ByteDance-Seed/VeOmni/pull/735#discussion_r3199080363 Gemini comments here

tasks/deprecated_task/train_flux.py
Coach257: Agent fixed these automatically 😀

veomni/utils/lora_utils.py
Coach257: DTensor random ops are shard-aware, nn.init_weight generates param based on (seed + shard_offset).

FoolPlayer (approved):

PR #751, opened 2026-05-13
[config, ci, docker] feat: make transformers v5.2.0 the default install

Luosuu:
## Summary

Flip the project default from transformers **4.57.3** to **5.2.0**. v4 stays installable as a sunset legacy track via a renamed extra.

- `pyproject.toml`: the existing `transformers-stable` dependency group (kept in `default-groups`) now pins `transformers==5.2.0`. The old `transformers5-exp` extra is replaced by a new opt-in extra `transformers-v4-legacy` that pins `4.57.3`. `[tool.uv.conflicts]` pair updated to `transformers-stable ↔ transformers-v4-legacy`. `uv.lock` regenerated.
- **CI workflows**: every step that was passing `--no-group transformers-stable --extra transformers5-exp` to force v5 now just runs the default `uv sync` (v5 is default). The parallel v4 jobs (gpu/npu unit + e2e) explicitly opt in via `--no-group transformers-stable --extra transformers-v4-legacy` so they keep covering the legacy path.
- **Ascend Dockerfiles**: previous `pip install -e .[npu_aarch64,transformers-stable]` referenced a uv dep-group as a pip extra (silently a no-op for pip), so transformers was being resolved off PyPI indeterminately. Replaced with explicit `pip install transformers==5.2.0` to pin the new default.
- **Docs** (`AGENTS.md`, `.agents/knowledge/uv.md` + `constraints.md`, `docs/get_started/installation/install_ascend_{arm,x86}.md`, `docs/examples/qwen3_5.md`, `docs/transformers_v5/veomni_flash_attention_kernel_adapter.md`) and **skills** (`veomni-uv-update`, `veomni-migrate-transformers-v5`, `veomni-debug`) reframe v5 as the default and v4 as the legacy escape hatch.

## How to use after this PR

```bash
# Default — installs transformers 5.2.0
uv sync --extra gpu --extra audio --dev

# Legacy v4 escape hatch (sunset path)
uv sync --no-group transformers-stable --extra transformers-v4-legacy --extra gpu --extra audio --dev
```

## Adjacent fixes pulled in to keep CI green

Making v5 the default exposed several pre-existing v5/NPU gaps that were masked while v5 was opt-in and only exercised by a subset of CI steps. Fixing them here instead of leaving CI red on the swap:

- **qwen2 v5 docstrings** (`veomni/utils/model_outputs.py`): transformers v5's `@auto_docstring` validates that any class used as a return annotation has an `Args:`/`Parameters:` section. Added them to the five model-specific `*WithLogProbs` subclasses (Qwen2VL, Qwen2.5-VL, Qwen3-VL, Qwen3-VL-MoE, Qwen3OmniMoeThinker).
- **qwen3_omni_moe v5 weight-sync adapter** (`tests/models/weight_sync_adapters.py`): HF v5 stores thinker experts as fused stacked parameters (`experts.gate_up_proj`, `experts.down_proj`), the same layout VeOmni already uses. Detect the v5 layout and skip the v4-style per-expert stacking.
- **qwen2_5_vl Ulysses SP** (`veomni/models/transformers/qwen2_5vl/...`): added the same `get_{image,video}_features` override qwen2_vl already has — return the raw SP-scattered `pooler_output` instead of calling the upstream `torch.split(...)` whose `split_sizes` are computed from the full grid.
- **qwen3_5 / qwen3_5_moe patchgen imports** (`veomni/models/transformers/qwen3_5*/...`): four patch_gen_configs imported `gather_outputs` / `slice_input_tensor` from the wrong module. Moved to the package root; regenerated `.py` and `.diff` sidecars.
- **NPU per-model ops overrides** (`tests/tools/training_utils.py`): qwen2 / qwen3_moe added with `rms_norm_implementation=eager` and `rotary_pos_emb_implementation=eager` until a `KERNEL_REGISTRY` `KernelSpec(name="npu", ...)` exists for those ops.
- **test_trainer_saveload + deepseek_v3 on v5** (`tests/checkpoints/test_trainer_saveload.py`): deepseek_v3 is still v4-only in VeOmni, so its saveload case is skipped on v5 (qwen3_moe still covers the dcp save/load path).

## CI hygiene

- **Pin v4-legacy env** in `gpu_*_tests_v4` via `UV_NO_SYNC=1`: without this, every `uv run --frozen pytest …` after the explicit v4 sync was silently re-syncing to the project default (v5) and overriding the legacy pin.
- **Drop v4-legacy NPU CI**: NPU now runs only the new default (v5). v4 GPU coverage stays via the dedicated `*_v4` jobs.
- **`Free shared L20 GPU` pre-step** on all GPU jobs: `nvidia-smi --query-compute-apps=pid | xargs kill -9` before sync so leaked processes from prior CI runs on the same physical host don't OOM us.
- **`chunk_loss` for qwen3_5 / qwen3_5_moe CE on GPU**: the fused Liger CE materialises the full `[B, S, V]` logits buffer (~5 GiB on the qwen3_5 toy vocab), which OOMs on this models even on a clean card. Switched to chunk_loss in both `_GPU_PER_MODEL_OVERRIDES` and the inline override in `test_models_patch`.
- **Tear down model/optimizer/lr_scheduler** between modes in `test_models_patch`: the previous `_release_device_memory()` only ran GC + `empty_cache()` after the forward — it never dropped the references on the trainer, so qwen3_5's ~5 GiB per-mode footprint accumulated to ~40 GiB across the 6-8 modes the test sweeps. Drop the instance attrs at the start of each `forward_backward_step`.
- **qwen3_5 / qwen3_5_moe gnorm tolerance** bumped to 5% (`tests/models/test_models_patch.py` parametrize): with chunk_loss in liger mode, VeOmni-backend gnorms cluster around the eager value (e.g. 14.5625), and HF native CE drifts ~3% on the qwen3_5 GatedDeltaNet path. Pre-existing implementation drift that was just inside the 1% default with liger but not with chunk_loss.
- **`UV_HTTP_TIMEOUT=300`** on all GPU + NPU workflows: uv's default 30s timeout fires intermittently on the CN runners while pulling large wheels (cudnn ~617 MiB, torch-npu, virtualenv).
- **Regenerated patchgen sidecars** for the surgical `.py` edits on this branch via `python -m veomni.patchgen.check_patchgen --fix` (the tool CI itself runs, which applies ruff in the same order the check uses).

## Test plan

- [x] `uv lock` resolves cleanly with both forks present in the lockfile.
- [x] `uv sync --frozen --extra gpu --extra audio --group dev` installs `transformers==5.2.0`.
- [x] `uv sync --frozen --no-group transformers-stable --extra transformers-v4-legacy --extra gpu --extra audio --group dev` downgrades to `transformers==4.57.3`.
- [x] `make quality` is clean.
- [x] `grep -rn "transformers5-exp" .` returns nothing (lockfile excluded).
- [x] NPU unit + NPU e2e tests pass on this branch.
- [x] GPU e2e v4 passes.
- [ ] GPU unit v4 + GPU unit v5 + GPU e2e v5 — flaky on shared L20 runner, mitigations land in this PR (kill stale processes pre-step, chunk_loss, teardown between modes). Retry-friendly.
- [x] check_patchgen passes (no drift between committed and freshly-regenerated patchgen output).


veomni/utils/model_outputs.py
```diff
@@ -90,39 +90,74 @@ class MoeCausalLMOutputWithLogProbs(MoeCausalLMOutputWithPast):
 
 @dataclass
 class Qwen2VLCausalLMOutputWithLogProbs(Qwen2VLCausalLMOutputWithPast):
```
Coach257, line 92: Maybe move these specific Output definition to modeling_gen_config.py. For example: https://github.com/ByteDance-Seed/VeOmni/blob/882a724fd64cd6ad91259399a7fa652f5da453df/veomni/models/transformers/qwen3_5/qwen3_5_gpu_patch_gen_config.py#L1217-L1236

Luosuu: will do in the following pr. this pr tries to focus on v5 default env switch only

TimYangst (approved):
LGTM, thanks!

PR #758, opened 2026-05-15
[lora] feat: native PEFT-free LoRA stack with MoE-LoRA + expert-parallel support

Coach257:
### What does this PR do?

Adds a self-contained, **PEFT-free** LoRA stack (`veomni.lora`) as a first-class `BaseTrainer` feature, and builds MoE-LoRA + expert-parallel (EP) support on top of it. `VeOmniLoraModel` / `VeOmniLoraConfig` replace `peft.PeftModel` / `LoraConfig` while staying **format-compatible** with PEFT (reads/writes standard `adapter_config.json` + `adapter_model.safetensors`), so adapters interoperate with stock `peft` in both directions. The stack is FSDP2-native and requires no `peft` at train time.

### Key features

- **Native LoRA core**: `VeOmniLoraModel`, `VeOmniLoraConfig`, `LoraLinear` — dense `nn.Linear` LoRA with PEFT-compatible FQNs (`base_layer` / `lora_A` / `lora_B`), rank/alpha patterns, rsLoRA, dropout, bias modes, adapter resume.
- **MoE-LoRA (three modes)** on fused v5 experts (`gate_up_proj` / `down_proj` 3-D `nn.Parameter`s):
  1. **Independent** — per-expert LoRA (`LoraIndependentExperts`, `[E, r, H]`), `share_expert_lora: false`.
  2. **Experts-shared** — one LoRA pair shared across all experts of a layer (`LoraSharedExperts`, `[r, H]`), `share_expert_lora: true`.
  3. **Shared-expert** — plain `LoraLinear` on the dense `shared_expert` module via `lora_modules`.
- **Semantic target mapping**: on fused-MoE models, listing `gate_proj` / `up_proj` / `down_proj` in `lora_modules` auto-maps to the fused `target_parameters` (`resolve_fused_moe_lora_targets` + per-model `_convert_lora_targets_to_parameters` hook) — no hand-written glob patterns.
- **Fused Triton MoE-LoRA kernels** (`veomni/lora/ops/`): fwd + bwd group-GEMM for both independent and shared modes; transparent eager fallback on Quack/NPU.
- **Expert parallelism (EP + FSDP2)**:
  - Independent: `[E, r, H] → [E/ep, r/fsdp, H]`.
  - Experts-shared: replicated across the EP group, with a cross-EP grad SUM all-reduce hook to keep replicas in lock-step, and EP-aware grad-norm handling so EP=1 and EP=N produce identical norms.
- **PEFT-format checkpointing**: DCP + safetensors save/load, EP-shard reassembly, adapter-config round-trip; MoE metadata embedded in `adapter_config.json` (`veomni_lora.moe_mode` + `target_parameters`) — no separate sidecar file.

### Config example

```yaml
model:
  lora_config:
    rank: 64
    alpha: 32
    lora_modules: [q_proj, k_proj, v_proj, o_proj, gate_proj, up_proj, down_proj]
    share_expert_lora: false   # true → experts-shared MoE-LoRA
```

### Architecture (new module layout)

- `veomni/lora/`: `config.py`, `model.py`, `layers.py`, `state_dict.py`, `weight_loading.py`, `mapping.py`, `target_mapping.py`, `moe_layers.py`, `ops/`.
- Integrations: `trainer/base.py` (`_setup_lora`), `models/module_utils.py` (native weight load), `optim/optimizer.py` (EP-replicated LoRA bucketing), `distributed/fsdp2/clip_grad_norm.py` + `distributed/parallel_plan.py` (EP-aware clipping / runtime plan).
- Supported fused-MoE models: `qwen3_moe`, `qwen3_omni_moe`, `qwen3_vl_moe` (+ example configs for `deepseek_v3`, `qwen3_5_moe_vl`).

### Testing

CPU unit + multi-GPU integration coverage:

- Native LoRA: `test_veomni_lora_native.py`, `test_veomni_lora_moe_native.py` (incl. PEFT bidirectional interop, injection structure, save/reload round-trip, post-load init no-clobber guard).
- MoE-LoRA numerics: `test_moe_lora_eager.py`, `test_moe_lora_fused.py` (eager↔fused parity), `test_moe_lora_ep2.py` (EP=2 save/load + align), `test_moe_lora_trainer.py` (trainer round-trip).
- Target mapping: `test_target_mapping.py`.
- Grad clipping: `test_extra_parallel_clip_grad_norm.py` (8-GPU L2 integration + single-process inf-norm unit tests covering the EP multi-bucket path).
- Checkpoint key normalization: `test_dcp_checkpointer.py` (incl. PEFT `base_model.*` case).
- CI (`gpu_unit_tests.yml`) runs the LoRA e2e suite.

### Compatibility / notes

- No `peft` runtime dependency; `peft==0.18.1` ships in the hardware extras only for the one interop test (`pytest.importorskip`).
- Dropped obsolete `lora` / `megatron` / `trl` extras.
- Adapters remain interchangeable with stock PEFT (standard config + safetensors layout).
- Deferred as follow-up work (not in this PR): bucketing the per-param EP grad all-reduce into one collective, and deduplicating the shared-LoRA forward to once-per-token.

### Docs

`docs/key_features/lora.md` fully updated (install, config table, three MoE modes, EP behavior, checkpoint format, semantic target mapping).


Luosuu (commented):
## Code Review — [lora] feat: add moe-lora

Thorough review of the full ~8,500-line diff. Several correctness bugs need fixing before merge.

---

### BUG-1 (CRITICAL): Double `load_lora_model_weights` call in `load_model_weights`
**File**: `veomni/models/module_utils.py`

The diff adds a new `if is_peft_model and adapter_path:` block with `parallel_plan` support, but the pre-existing `if is_peft_model and adapter_path:` call is NOT removed. `load_lora_model_weights()` is called **twice** for the same adapter path — the second call silently double-applies the LoRA initialization.

**Fix**: Remove the original call and keep only the new block that passes `parallel_plan`.

---

### BUG-2 (CRITICAL): `torch.maximum` called with >2 positional arguments
**File**: `veomni/distributed/fsdp2/clip_grad_norm.py`

```python
total_norm = torch.maximum(
    non_extra_parallel_total,
    *extra_parallel_total.values(),
    *extra_parallel_replicated_total.values(),  # new bucket added by this PR
)
```

`torch.maximum` is binary — exactly 2 tensor arguments. With any active EP group this raises `TypeError` at runtime.

**Fix**:
```python
total_norm = non_extra_parallel_total
for t in itertools.chain(extra_parallel_total.values(), extra_parallel_replicated_total.values()):
    total_norm = torch.maximum(total_norm, t)
```

---

### BUG-3 (HIGH): `_init_lora_parameter` broad reset can clobber loaded LoRA weights
**File**: `veomni/utils/lora_utils.py`, `_init_lora_parameter()`

The function calls `reset_lora_parameters(init_lora_weights=True)` on the **entire wrapper** the first time any parameter name under a `LoraSharedExperts`/`LoraIndependentExperts` wrapper is seen, then returns early for subsequent keys. If any key escapes PEFT-prefix normalization and hits `_init_lora_parameter` after `load_lora_model_weights` has already populated the wrapper, the reset zeros `lora_B` and re-randomizes `lora_A` for all specs.

**Fix**: Guard the reset to only fire when all parameters are still on meta device:
```python
if is_lora_moe_experts(module):
    if all(p.is_meta for _, p in module.named_parameters()):
        module.reset_lora_parameters(init_lora_weights=True)
    return
```

---

### BUG-4 (MEDIUM): EP grad sync hook — `to_local()` may return a copy, not a view
**File**: `veomni/utils/moe_lora.py`, `LoraSharedExperts._ensure_ep_grad_sync_hooks()`

```python
local = grad.to_local() if hasattr(grad, "to_local") else grad
if local.is_contiguous():
    dist.all_reduce(local, ...)
else:
    buf = local.contiguous()
    dist.all_reduce(buf, ...)
    local.copy_(buf)   # ← writes into a dead temp if to_local() returned a copy
```

`to_local()` on a DTensor may return a copy for non-contiguous storage layouts. In that case `local.copy_(buf)` writes into a dead temporary and the all-reduce result is silently lost.

**Fix**: Access `_local_tensor` directly:
```python
local_storage = grad._local_tensor if isinstance(grad, DTensor) else grad
if not local_storage.is_contiguous():
    buf = local_storage.contiguous()
    dist.all_reduce(buf, op=dist.ReduceOp.SUM, group=ep_group)
    local_storage.copy_(buf)
else:
    dist.all_reduce(local_storage, op=dist.ReduceOp.SUM, group=ep_group)
```

---

### BUG-5 (MEDIUM): `id(p)` stability across FSDP2 parallelization is undocumented
**File**: `veomni/optim/optimizer.py`, `_collect_ep_replicated_lora_param_ids()`

`id(p)` is stored before FSDP2 replaces `nn.Parameter` objects with DTensor-backed versions. If `build_optimizer` is called before `build_parallelize_model`, the collected IDs refer to GC-collected objects and Python may reuse those addresses for unrelated new objects. Add an assertion or comment that `build_optimizer` must be called after `build_parallelize_model` (fully_shard), or switch to FQN-based keys.

---

### DESIGN-1: `target_parameters`-only LoRA unnecessarily blocked
**File**: `veomni/utils/moe_lora.py`, `veomni_get_peft_model()`

```python
if wrap_moe and not target_modules:
    raise NotImplementedError("target_parameters set without target_modules is not supported yet ...")
```

Forces every user to also specify `lora_modules` even for expert-only LoRA. At minimum, document this constraint prominently in `docs/key_features/lora.md`.

---

### DESIGN-2: Hard sidecar version rejection with no migration path
**File**: `veomni/utils/moe_lora.py`, `read_moe_lora_sidecar()`

Version mismatch raises immediately with no migration tool or guide. If v1–v3 checkpoints exist internally, provide a conversion utility or document what changed between versions.

---

### DESIGN-3: `_veomni_moe_lora_patterns` model attribute is fragile
**File**: `veomni/utils/moe_lora.py`, `_apply_moe_lora_with()`

```python
model._veomni_moe_lora_patterns = list(target_parameter_patterns)
```

Not in `state_dict`, not in `adapter_config.json` — lost on deep-copy or pickle. The patterns are already written to the sidecar JSON; consider removing this attribute and reading from the sidecar exclusively.

---

### TEST GAPS

- **GAP-1 (HIGH)**: No test verifies that loaded LoRA weights survive `post_process_after_weight_loading` (i.e., `_init_lora_parameter` does NOT reset them). The trainer round-trip tests only catch this indirectly via wrong numerics.
- **GAP-2 (MEDIUM)**: No unit test for multi-EP-group `clip_grad_norm` path — would have caught BUG-2.
- **GAP-5 (LOW)**: `_normalize_key` `model.base_model.*` stripping in `dcp_checkpointer.py` is untested.

---

### MINOR / STYLE

- **STYLE-1**: Inconsistent naming `lora_a_down` (dispatcher) vs `lora_a_dn` (autograd Function) — standardize.
- **STYLE-2**: `first_pname = next(iter(self._lora_specs))` relies on dict insertion order — add a comment.
- **STYLE-3**: `Dict` from `typing` used in Python 3.11+ target code — use builtin `dict[str, Shard]`.
- **STYLE-4**: Fragile adapter name injection in `save_safetensor_utils.py` EP shard restoration (`name[:-len('.weight')] + f".{adapter_name}.weight"`) — should assert invariant or iterate known adapter names.
- **PERF-2**: `with torch.no_grad()` inside `_eager_forward` is correct for routing tensors but needs a comment to clarify it's safe under gradient checkpointing.
- **DESIGN-4**: PR description says "three modes" but there are two wrapper classes — the third "mode" is plain PEFT `lora_modules` on `nn.Linear`. Clarify in the description to avoid reviewer confusion.

---

### Summary

| ID | Severity | File | Issue |
|----|----------|------|-------|
| BUG-1 | **CRITICAL** | `module_utils.py` | `load_lora_model_weights` called twice |
| BUG-2 | **CRITICAL** | `clip_grad_norm.py` | `torch.maximum` >2 args → `TypeError` at runtime |
| BUG-3 | **HIGH** | `lora_utils.py` | Init reset can clobber loaded LoRA weights |
| BUG-4 | MEDIUM | `moe_lora.py` | EP grad hook: `to_local()` copy silently loses all-reduce |
| BUG-5 | MEDIUM | `optimizer.py` | `id(p)` invalid after FSDP2 wrapping if order wrong |
| GAP-1 | HIGH | `lora_utils.py` | No test: loaded weights survive `post_process_after_weight_loading` |
| GAP-2 | MEDIUM | `clip_grad_norm.py` | No unit test for multi-EP norm path |

**Must fix before merge**: BUG-1, BUG-2, BUG-3. BUG-4 and GAP-1 are strongly recommended.

Luosuu (commented):
Re-posting the earlier consolidated review as inline comments anchored to the offending lines. Three blockers (BUG-1, BUG-2, BUG-3) plus two follow-ups. The standalone summary comment can be ignored — the line comments below are authoritative.

veomni/models/module_utils.py
```diff
@@ -340,14 +370,31 @@ def _dispatch_kv(name: str, tensor: "torch.Tensor") -> None:
...
+            dtensor_factory,
+            parameter_names_to_load=parameter_names_to_load,
+            # Same EP-aware slicing as the rank0-broadcast variant below;
+            # ``LoraIndependentExperts`` LoRA tensors need to be shrunk
+            # from ``[E, ...]`` to ``[E_local, ...]`` before the DTensor
+            # copy or the propagation asserts on shape mismatch.
+            parallel_plan=parallel_plan,
+        )
 
     if is_peft_model and adapter_path:
```
Luosuu: **BUG (CRITICAL): `load_lora_model_weights` is called twice.**

The new block at lines 382–397 (with `parallel_plan=parallel_plan`) was added, but the pre-existing block starting here at line 399 was not removed. Both blocks run for every PEFT load — `load_lora_model_weights` executes back-to-back and the second call silently re-dispatches the adapter weights into already-populated DTensor params.

In the best case this is wasted work; in the worst case (if the second call's loading path differs in EP slicing because it lacks `parallel_plan`) it overwrites correctly-sliced LoRA tensors with full-shape ones, breaking EP runs.

**Fix**: delete lines 399–409 (the second `if is_peft_model and adapter_path:` block). The new block at 382–397 supersedes it.

veomni/distributed/fsdp2/clip_grad_norm.py
```diff
@@ -110,14 +134,45 @@ def extra_parallel_fsdp2_clip_grad_norm(
...
+                reduce_groups=[
+                    (f"{para}_fsdp", extra_parallel_fsdp_group[para]),
+                ],
+            )
+            extra_parallel_replicated_total[para] = para_total
+            logger.debug_rank0(f"{para} replicated total grad norm: {para_total}")
+
     if math.isinf(norm_type):
-        total_norm = torch.maximum(non_extra_parallel_total, *extra_parallel_total.values())
+        total_norm = torch.maximum(
```
Luosuu: **BUG (CRITICAL): `torch.maximum` accepts exactly two tensor arguments.**

```python
total_norm = torch.maximum(
    non_extra_parallel_total,
    *extra_parallel_total.values(),
    *extra_parallel_replicated_total.values(),
)
```

With any active EP group this expands to ≥3 tensors and raises `TypeError: maximum() takes 2 positional arguments but N were given` at runtime. The pre-existing two-bucket call (`torch.maximum(non_extra_parallel_total, *extra_parallel_total.values())`) is also latently broken if there are multiple EP names; this PR makes it worse by adding `extra_parallel_replicated_total`.

**Fix**:
```python
import itertools
total_norm = non_extra_parallel_total
for t in itertools.chain(extra_parallel_total.values(), extra_parallel_replicated_total.values()):
    total_norm = torch.maximum(total_norm, t)
```

Also worth adding a unit test that exercises `norm_type == inf` with two `ps.extra_parallel_names` entries — none of the trainer tests would catch this regression directly.

veomni/utils/lora_utils.py
```diff
@@ -111,30 +143,58 @@ def load_lora_model_weights(
...
+    * **Standard PEFT LoRA layer**: when the immediate ``lora_A`` /
+      ``lora_B`` parent has ``reset_lora_parameters``, loop over its
+      adapters and reset each. ``lora_B`` is reset implicitly by the
+      ``lora_A`` reset, so we only need to dispatch on ``lora_A`` keys.
+    """
+    from .moe_lora import is_lora_moe_experts
+
+    for piece in name.split("."):
+        if is_lora_moe_experts(module):
+            module.reset_lora_parameters(init_lora_weights=True)
```
Luosuu: **BUG (HIGH): broad reset can clobber loaded LoRA weights.**

`reset_lora_parameters(init_lora_weights=True)` is called on the entire `LoraSharedExperts` / `LoraIndependentExperts` wrapper as soon as any key whose ancestor is a wrapper hits `_init_lora_parameter`. That resets **all** specs and **all** adapters in the wrapper (zeros `lora_B`, re-randomizes `lora_A`).

If an upstream PEFT-prefix normalization edge case leaves any LoRA key in `parameter_names_to_load` after `load_lora_model_weights` populated the wrapper, `post_process_after_weight_loading` will hit this branch and silently overwrite freshly-loaded values with random init.

**Fix**: only reset when no LoRA param has been loaded yet (everything still on meta device):
```python
if is_lora_moe_experts(module):
    if all(p.is_meta for _, p in module.named_parameters(recurse=True)):
        module.reset_lora_parameters(init_lora_weights=True)
    return
```

Applies to the duplicate reset at line 192 too.

**Test gap**: there is no unit test asserting that loaded LoRA weights survive `post_process_after_weight_loading`. The trainer round-trip tests in `tests/lora/test_moe_lora_trainer.py` would catch this only if numerics drift — please add a focused assertion.

veomni/utils/moe_lora.py
```diff
@@ -0,0 +1,1629 @@
...
+                # the collective on the original non-contiguous view, which
+                # either errors or, on newer NCCL, runs on an internal temp
+                # whose result was never written back to ``p.grad`` -- shared
+                # LoRA grads would silently drift across EP ranks.
+                if local.is_contiguous():
+                    dist.all_reduce(local, op=dist.ReduceOp.SUM, group=ep_group)
+                else:
+                    buf = local.contiguous()
+                    dist.all_reduce(buf, op=dist.ReduceOp.SUM, group=ep_group)
+                    local.copy_(buf)
```
Luosuu: **Question (MEDIUM): `to_local()` returning a copy vs. view.**

The inline comment at line 616 asserts that `to_local()` returns a view of `DTensor._local_tensor`, so `local.copy_(buf)` propagates back. For `Shard` placements with row-major storage this holds today, but PyTorch's API contract for `to_local()` is "returns the local component" — there are placement combinations (replicate after redistribute, partial after reduce) where the returned tensor is not guaranteed to be the same storage object.

Would you mind switching to `grad._local_tensor` directly (or a placement-restricted assertion) to make the writability invariant explicit?

```python
local_storage = grad._local_tensor if isinstance(grad, DTensor) else grad
if local_storage.is_contiguous():
    dist.all_reduce(local_storage, op=dist.ReduceOp.SUM, group=ep_group)
else:
    buf = local_storage.contiguous()
    dist.all_reduce(buf, op=dist.ReduceOp.SUM, group=ep_group)
    local_storage.copy_(buf)
```

Non-blocking — but the silent-grad-drift failure mode is exactly the scary kind.

veomni/optim/optimizer.py
```diff
@@ -620,6 +660,19 @@ def _build(groups: Sequence[Dict[str, Any]]) -> Optimizer:
...
+    # ``non_extra_parallel`` optimizer — its DTensors are on a different mesh and
+    # ``foreach_*`` ops reject mixed-mesh tensor lists), but unlike the EP-sliced
+    # base experts and Mode-1 per-expert LoRA they are *replicated* across the
+    # EP group (kept in lock-step by ``LoraSharedExperts._ensure_ep_grad_sync_hooks``).
+    # ``extra_parallel_fsdp2_clip_grad_norm`` uses this set to skip the EP all-reduce
+    # for these params; otherwise their squared norms are summed once per EP rank
+    # and the global norm comes out larger than the EP=1 baseline, breaking
+    # cross-EP grad-norm parity (and therefore any optimizer step that depends
+    # on the clipped grad).
+    model._ep_replicated_lora_param_ids = _collect_ep_replicated_lora_param_ids(model)
```
Luosuu, line 698: **Concern (MEDIUM): `id(p)` stability across FSDP2 parallelization is order-sensitive.**

`_collect_ep_replicated_lora_param_ids` snapshots `id(p)` here. After `build_parallelize_model` / `fully_shard`, FSDP2 replaces the original `nn.Parameter` objects with new DTensor-backed ones — the captured IDs become stale, and Python may reuse those addresses for unrelated new objects, causing false positives in the `_ep_replicated_lora_param_ids` membership check downstream (e.g. in `extra_parallel_fsdp2_clip_grad_norm`).

This is safe only if `build_optimizer` is **always** invoked after `build_parallelize_model`. Could you either:
1. Add an assertion at the top of `_build` confirming the params are already DTensors (`isinstance(p, DTensor)`), or
2. Key the set by FQN instead of `id()` so it survives re-wrapping?

At minimum, please add a comment here documenting the required call order — otherwise this is a foot-gun for anyone adding a new training-script flow.

veomni/utils/save_safetensor_utils.py
```diff
@@ -262,6 +262,42 @@ def save_lora_adapter_with_dcp(
...
+        # ``get_peft_model_state_dict`` strips the adapter name from keys
+        # (``...gate_proj.lora_A.default.weight`` -> ``...gate_proj.lora_A.weight``)
+        # while ``_fqn2spec_info`` is built from ``model.named_parameters()``
+        # and keeps the adapter name. Re-insert it (matching the save's
+        # ``adapter_name``) when looking up the spec, otherwise no LoRA
+        # tensor matches and the EP gather is silently skipped.
+        for name in list(lora_state):
+            spec = fqn2spec_info.get(name)
+            if spec is None and name.endswith(".weight"):
+                spec_key = f"{name[: -len('.weight')]}.{adapter_name}.weight"
```
Luosuu, line 298: **Minor (LOW): fragile adapter-name injection.**

```python
spec_key = f"{name[: -len('.weight')]}.{adapter_name}.weight"
spec = fqn2spec_info.get(spec_key)
```

This silently returns `None` (skipping the EP gather) on any PEFT key-format change or multi-adapter scenario, with no warning. Two cheap improvements:
1. Assert `spec is not None` after the lookup if you expect every `.weight` key to map.
2. Iterate `model.peft_config.keys()` to handle multi-adapter setups instead of injecting a single `adapter_name`.

Likely fine for the current single-adapter LoRA path but worth hardening.

Luosuu:
maybe we can verify whether this agent speaks correctly..

Coach257:
> maybe we can verify whether this agent speaks correctly..

It seems your agent is better than my agent haha

veomni/utils/lora_utils.py
Coach257: A lora layer is `all loaded` or `nothing to load`. So here init a full lora layer is acceptable.

veomni/utils/save_safetensor_utils.py
Coach257: Will be updated when multi-adapter is supported.

machiavellierik-oss (approved):

Coach257:
The points below came out of review and are intended as incremental work on top of that baseline.

---

## 1. MoE-LoRA variants & granular targeting

Today `target_parameters` wraps the whole fused experts module as one unit (`gate_up_proj` + `down_proj` together), so LoRA is always attached to all three projections at once. We cannot currently attach an adapter to just one projection. Proposed extensions:

- **Per-projection targeting** — allow LoRA on `gate` / `up` / `down`  individually (e.g. up-only or down-only), instead of the all-or-nothing  wrap of the fused experts module.
- **Hot-expert LoRA** — attach (or up-rank) LoRA only on the most  frequently-routed experts, leaving cold experts untouched, to spend adapter  capacity where it matters.
- **Variable / non-uniform LoRA count** — drop the assumption of one uniform LoRA layout across all experts; support a configurable number of adapters  (e.g. a small pool of shared adapters that is smaller than the expert count but larger than 1), interpolating between Mode 1 and Mode 2.
- **Multi-LoRA adapters** — support multiple named adapters coexisting on the same MoE module (load/switch/merge), as a foundation for multi-task and serving scenarios.

This directly addresses the open design question on Mode 2: a shared adapter can cap per-expert specialization, and these variants are the path to recover it without paying the full per-expert parameter/activation cost.

## 2. Performance optimizations (shared-LoRA path)

Both items are specific to `LoraSharedExperts`:

- **Bucket the EP gradient all-reduce.** Under expert parallelism each rank only sees a subset of tokens, so the shared LoRA params need a cross-EP gradient sync. The current `_ensure_ep_grad_sync_hooks` registers a `register_post_accumulate_grad_hook` *per LoRA parameter*, and each hook issues its **own** `dist.all_reduce` on `ep_group`. These should be coalesced into a single flattened/bucketed collective (one `all_reduce` over a contiguous buffer) to cut launch and latency overhead.
- **Remove redundant per-(token, expert) LoRA recompute in the forward.** Because the adapter is *identical* across experts in shared mode, the LoRA projection of a given token does not depend on which expert it is routed to. The fused forward currently still evaluates the LoRA contribution per token-expert pair, duplicating work. Instead, compute the shared LoRA contribution **once per token** and scatter-add it onto each selected expert's output. (The base `gate_up_proj` / `down_proj` GEMMs remain per-expert; only the shared LoRA term is deduplicated.)

## 3. Reduce PEFT coupling → VeOmni-native LoRA adapter stack

The current integration leans on PEFT (`PeftModel`, PEFT-style key layout, `base_layer` / `lora_A` / `lora_B` naming) for compatibility, but PEFT's coverage of MoE / fused-expert layouts is incomplete and forces several workarounds (custom expert wrappers, FQN remapping, manual EP gradient sync, fused-kernel dispatch). As the variants in §1 and the optimizations in §2 grow, the PEFT compatibility layer becomes a constraint rather than a convenience.

Proposal: incrementally build a self-contained VeOmni LoRA adapter stack (adapter definition, attach/merge/save/load, checkpointing, EP/FSDP-aware gradient handling) that does not depend on PEFT, while keeping a PEFT-format
import/export shim for interop with the broader ecosystem.

tests/distributed/test_extra_parallel_clip_grad_norm.py
```diff
@@ -0,0 +1,159 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
```
Coach257: del this file

Coach257:
#866  @Liuweixiong0118 cc

Luosuu (approved):
This is much cleaner than I thought it would be. LGTM! Thanks

PR #760, opened 2026-05-17
[model, ci] test: add implicit CUDA sync gate for qwen3_5 generated modeling

TimYangst:
### What does this PR do?

Adds a unit-level **ratchet test** that detects implicit host↔device CUDA syncs in VeOmni-patched v5 modeling under `veomni/models/transformers/*/generated/`. The gate runs each model's forward under `torch.cuda.set_sync_debug_mode("warn")`, captures the warnings, and fails if a new sync site shows up that isn't on the per-case allowlist. Companion to the manual `debug-cuda-sync` skill (real-weights / SP/EP investigation flow) — this test is the cheap PR-time regression gate.

### The principle (two axes)

For each surfaced sync site, classify along **two independent axes**:

1. **Owner** — VeOmni-patched (line in patchgen `.diff`) vs HF-verbatim (unchanged from upstream, even though it lives inside `generated/`).
2. **Path** — production (runs every step, no override above) vs eager-only fallback (production bypasses via OpSlot / other VeOmni override).

The rule:

| | Production path | Eager-only fallback (production bypasses) |
|---|---|---|
| **VeOmni-patched** | Fix the patch | Fix the patch |
| **HF-verbatim** | **Add an `override_method` patch and fix** (becomes ours) | Leave alone / allowlist |

`_ALLOWED_SYNCS` entries are tagged in the reason string:

- `"HF-eager-only: ..."` — accepted long-term (production bypasses).
- `"HF-prod-pending-fix: ..."` — production-path HF-verbatim; tracked for fix; reason names the follow-up branch / PR.
- `"algorithm-essential: ..."` — accepted by design.

### Coverage

3 cases, ~13s on A100, wired into `gpu_unit_tests_v5` next to `test_models_logits_equal_v5.py`:

- `qwen3_5-text-eager` — non-MoE, eager attention through our patched `Qwen3_5Model.forward`.
- `qwen3_5-text-fa2` — non-MoE, FA2 path through the same patched forward.
- `qwen3_5_moe-text-fa2-fused` — production path: FA2 + `fused_triton` MoE, exercising VeOmni's OpSlot short-circuit in `Qwen3_5MoeExperts.forward`.

**No eager-MoE case**. Eager `Qwen3_5MoeExperts.forward` body is HF-verbatim (VeOmni only adds the OpSlot short-circuit at the top); production never runs it because `fused_triton` bypasses the loop. Gating it would force the allowlist to absorb 5 HF-verbatim sync sites from an algorithm-inherent loop that's not on any real codepath, while adding no signal for VeOmni-patched code. The fa2-fused case covers everything we own on the MoE side.

The fa2-fused case also doubles as **structural validation of the OpSlot short-circuit**: if `fused_triton` silently fell back to the HF eager loop, those 5 verbatim sync sites would re-fire and the test would fail. It passes → the short-circuit is intact.

### Findings from baseline

3 entries in `_ALLOWED_SYNCS`, all the same upstream method `_update_linear_attn_mask`:

- `patched_modeling_qwen3_5_gpu.py:1750` (qwen3_5-text-eager + qwen3_5-text-fa2)
- `patched_modeling_qwen3_5_moe_gpu.py:1955` (qwen3_5_moe-text-fa2-fused)

The method is HF-verbatim (not in either Model class's `Methods patched:` list) but it's called unconditionally from `Qwen3_5{,Moe}Model.forward` and there's no VeOmni override above it — i.e. **production path**. Per the rule it should be fixed via an `override_method` patch (which makes it ours). Tagged `HF-prod-pending-fix` and pointing at branch `tingyang/fix/qwen3_5_key_fix` (where the fix already exists; needs rebase). When that fix lands the allowlist entries become dead and the gate's dead-entry detection forces the cleanup.

### Test

- `pytest tests/models/test_model_forward_no_implicit_sync.py` — 3 passed in ~13s on A100.
- `make quality` — passes.
- `make patchgen` — no `_ALLOWED_SYNCS` line numbers drift.

### API and Usage Example

No API changes — test-only addition plus one CI step.

### Design & Code Changes

- `tests/models/test_model_forward_no_implicit_sync.py` (new): runs forward under `set_sync_debug_mode("warn")`, captures `UserWarning`s matching `r"called a synchronizing"`, fails on any site whose `WarningMessage.filename` lives under `veomni/models/transformers/*/generated/` that isn't in the per-case `_ALLOWED_SYNCS` allowlist. Also fails on **dead** allowlist entries (listed but no longer observed) so a `make patchgen` line-shift can't silently rot the gate.
- Cases declared explicitly (not imported wholesale from `test_models_logits_equal_v5`) because the gate wants a different MoE backend than the logits test forces for HF parity, and prefers FA2 over SDPA. `_MOE_IMPL_BY_CASE` maps `case_id` → MoE backend.
- Warmup forward outside debug mode (so one-time rotary/kernel-autotune syncs don't pollute), then the measured forward inside the warning-capture block.
- `is_fused_moe_available()` skip for fused_triton cases — clear skip on a triton-less CI host instead of a confusing failure inside `apply_ops_config`.
- Assertion failure message walks the maintainer through the two-axis triage (read the line + check the patchgen `.diff` + decide path) and the right reason-tag.
- `.github/workflows/gpu_unit_tests.yml`: one pytest invocation in `Run models tests`, right after `test_models_logits_equal_v5.py`.
- `tests/special_sanity/check_device_api_usage.py`: add the test file to the CUDA-keyword allowlist (`torch.cuda.{get,set}_sync_debug_mode` has no `veomni.utils.device` equivalent and is intrinsically CUDA-only; test is `IS_CUDA_AVAILABLE`-gated).

### Follow-up work (not in this PR)

- **Land the existing `tingyang/fix/qwen3_5_key_fix` work** to fix `_update_linear_attn_mask` via `override_method` and kill the 3 `HF-prod-pending-fix` allowlist entries. Also lands the ViT-side fixes already on that branch.
- **Extend gate** to `qwen3_vl`, `qwen3_vl_moe`, `qwen3_omni_moe` — first cut in stacked PR #761.
- **Refactor**: extract the shared v5 test helpers (`_DTYPE_MAP`, `Case`, `_apply_determinism`, `_forward_target`, `_make_config`, `_make_inputs`, `_release`) into `tests/models/_v5_helpers.py` so the sync test doesn't import 7 private names from the sibling logits test.

- [x] Added/updated documentation (test docstring is comprehensive; no separate docs file warranted)
- [x] Added tests to CI workflow (this IS a new CI test, wired into `gpu_unit_tests_v5`)

### DIFF
--- a/tests/models/test_model_forward_no_implicit_sync.py
+++ b/tests/models/test_model_forward_no_implicit_sync.py
@@ -1,8 +1,24 @@
 """Implicit CUDA-sync gate for VeOmni-patched v5 modeling code.
 
 Runs each model's forward under ``torch.cuda.set_sync_debug_mode("warn")``
-and fails if any implicit host<->device sync originates from
+and fails if any new implicit host<->device sync site shows up in
 ``veomni/models/transformers/<model>/generated/``.
+
+The principle
+-------------
+We are gating **VeOmni's patches** (the code we wrote), not HF upstream.
+``generated/`` files are mostly HF source copied verbatim with imports
+resolved; only the bits marked ``[PATCHED CLASS]`` / ``[MODIFIED CLASS]
+Methods patched: ...`` are ours. The patchgen ``.diff`` next to each
+generated file is the authoritative ours-vs-HF view.
+
+The test inevitably observes warnings from *all* code in the generated
+file (patched and verbatim alike), so the allowlist exists to mark
+HF-verbatim sites as "not ours, leave alone". Concretely:
+
+  - **VeOmni-patched code introducing a sync** → fix the patch.
+  - **HF-verbatim code that syncs** → allowlist with reason "HF-verbatim;
+    out of scope". Do not try to improve upstream HF here.
 
 Why this matters
 ----------------
@@ -14,10 +30,9 @@ under SP/EP and small micro-batches, and can cascade into NCCL watchdog
 timeouts. See the ``debug-cuda-sync`` skill for the manual investigation
 flow against real weights.
 
-This test is the *unit-level ratchet* for the same property: it runs on
-toy configs (so only catches sync sites reachable from a tiny forward),
-but it's cheap and catches "someone added a ``.item()`` to a patch" at
-PR time. Real-model SP/EP coverage stays with the skill.
+This test is the *unit-level ratchet*: it runs on toy configs (so only
+catches sync sites reachable from a tiny forward) and is cheap enough
+to gate PRs. Real-model SP/EP coverage stays with the skill.
 
 Extending to more models
 ------------------------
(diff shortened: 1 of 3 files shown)


tests/models/test_model_forward_no_implicit_sync.py
```diff
@@ -0,0 +1,400 @@
...
+    for c in _ALL_CASES:
+        if c.case_id == case_id:
+            return c
+    raise KeyError(f"{case_id!r} not in test_models_logits_equal_v5.CASES")
+
+
+def _toy(name: str) -> str:
+    return os.path.join(
+        os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))), "tests", "toy_config", name
+    )
```
ZihaoW123: refactor: reuse shared _toy helper from test_models_logits_equal_v5 or extract to common test utils to avoid duplicate path resolution

TimYangst: Done, thanks!

Luosuu (approved):

PR #761, opened 2026-05-18
[model, ci] test: extend sync gate to qwen3_vl / qwen3_vl_moe / qwen3_omni_moe (skipped pending fix)

TimYangst:
> Stacked on top of #760 (`tingyang/ci/sync_test`). Merge that first or rebase this onto main once #760 lands.

### What does this PR do?

Adds three new cases to the implicit-CUDA-sync gate introduced in #760:

- `qwen3_vl-fa2` — non-MoE VLM, full multimodal forward (dummy 2x2 image), FA2 + bf16.
- `qwen3_vl_moe-fa2-fused` — VLM + production FA2 + `fused_triton` MoE.
- `qwen3_omni_moe-fa2-fused` — Omni thinker forward + production FA2 + `fused_triton` MoE.

The three new cases are **declared but skipped at runtime** via `_PENDING_FIX_CASES` because their initial run surfaces 21 VeOmni-touched sync sites that need to be fixed in the patches before the gate can be tightened. The skip preserves the cases as visible `SKIPPED` rows in pytest output and gives a clear conversion path: fix the patches, remove from `_PENDING_FIX_CASES`, sub-triage any HF-verbatim sites that remain (per the two-axis principle in #760).

### Findings from the first run

49 sites surfaced, triaged via the patchgen `.diff` (lines added/modified in the diff are ours; unchanged lines are HF-verbatim):

| case                            | VeOmni-touched | HF-verbatim |
| ------------------------------- | -------------: | ----------: |
| `qwen3_vl-fa2`                  |              9 |          12 |
| `qwen3_vl_moe-fa2-fused`        |              9 |          12 |
| `qwen3_omni_moe-fa2-fused`      |              3 |           4 |
| **total**                       |         **21** |      **28** |

VeOmni-touched concentration (same shape as the existing `.pr-drafts/tingyang-fix-qwen3_5_key_fix.md` draft, but for the VL/Omni files instead of qwen3_5* text-side):

- `rot_pos_ids` helper, `rot_pos_emb`, `fast_pos_embed_interpolate` — vision tower metadata derived directly from GPU `grid_thw`.
- Vision + Model `forward` — `cu_seqlens` built from GPU `repeat_interleave`, `n_image_tokens = image_mask.sum().item()`.
- qwen3_vl + qwen3_vl_moe mirror each other; qwen3_omni_moe has 3 forward-path sites.

[...]

PR #762, opened 2026-05-18
[model, perf] fix: remove production-path CPU syncs from Qwen3.5 / Qwen3.5-MoE patches

TimYangst:
### What does this PR do?

Removes production-path implicit host↔device CUDA syncs from VeOmni's `qwen3_5` and `qwen3_5_moe` patches. Surfaces:

1. **ViT `fast_pos_embed_interpolate`** — was iterating GPU `grid_thw` directly; every iteration forced syncs (`torch.linspace(steps=h)` with `h` 0-D GPU, `reshape(h // m_size, ...)`, `expand(t, ...)`) *per image per micro-batch*. Now takes a Python list materialised once by the ViT `forward`.
2. **ViT `forward`** — re-derived `cu_seqlens` / `total_seq_len` from GPU `grid_thw` via `torch.repeat_interleave(GPU_repeats)` + `[-1]` indexing, and read `max_seqlen` back from `(cu_seqlens[1:] - cu_seqlens[:-1]).max().detach().cpu().item()` for FA. Now materialises `grid_thw.tolist()` once at the top, builds `cu_seqlens` host-side as Python ints, carries `total_seq_len` as a plain Python `int`, and computes `vision_max_seqlen` purely on the host from `max(h*w for grid_thw_list, pad_seq_len)` — fed to every ViT block via `**kwargs` so the FA varlen call gets an int directly.
3. **ViT `rot_pos_emb`** — used to start its body with `grid_thw.tolist()`, a second per-forward sync on the same tensor. Adds an `override_method` that accepts the host-materialised list from the caller (with `torch.is_tensor` fallback for external callers). Combined with (2), the ViT forward goes from ~5–6 `grid_thw`-related syncs/forward down to **1** (the single `.tolist()` at the top, now shared across `fast_pos_embed_interpolate`, the `cu_seqlens` build, `rot_pos_emb`, and the host-side `max_seqlen`).
4. **`Qwen3_5(Moe)TextModel._update_linear_attn_mask`** — was reading GPU scalars (`cache_position[0] > 0`, `torch.all(attention_mask == 1)`) to decide a no-op. The `torch.all` short-circuit is dropped outright (returning the all-ones mask makes `apply_mask_to_padding_states` a no-op multiply — equivalent to upstream's `None`, minus the sync). The cached-forward branch is kept as a correctness guard but detected host-side from tensor shapes (`attention_mask.shape[-1] != cache_position.shape[-1]`). Result: 1–2 syncs/forward → 0, no behaviour change.
5. **`Qwen3_5(Moe)Model.forward`** — `n_image_tokens = image_mask.sum().item()` / `n_video_tokens = video_mask.sum().item()` were 1 sync each. Now computed host-side from the precomputed metadata: `sum(prod(image_grid_thw)) // spatial_merge_size**2`.

This is the patchgen `override_method` lane: the 3 sites in `_update_linear_attn_mask` and the second `.tolist()` in `rot_pos_emb` go from HF-verbatim (those methods were not in any `Methods patched:` list, so the lines were upstream-unchanged inside a MODIFIED CLASS) to VeOmni-patched. Per the two-axis principle landed in #760: production-path HF-verbatim syncs get fixed via `override_method` (which makes them ours), not allowlisted as "leave alone".

### Effect on the sync gate (#760)

After this lands, the 3 `HF-prod-pending-fix` allowlist entries from #760 become dead:

- `patched_modeling_qwen3_5_gpu.py:1750` (qwen3_5-text-eager + qwen3_5-text-fa2)
- `patched_modeling_qwen3_5_moe_gpu.py:1955` (qwen3_5_moe-text-fa2-fused)

The gate's dead-entry detection fires, forcing the cleanup in commit 6 (`[model, ci] test: drop dead _update_linear_attn_mask entries from sync gate allowlist`). After: qwen3_5 / qwen3_5_moe allowlist is **empty** — the gate is fully clean for these models.

### Commits (squash on merge if preferred; history is informative as-is)

1. `[model] perf: avoid host-device sync in Qwen3.5 ViT fast_pos_embed_interpolate` — Python-list iteration.
2. `[model] perf: remove host-device syncs from Qwen3.5 ViT / linear-attn forward path` — main change; introduces `override_method` for `_update_linear_attn_mask` + the ViT `forward` `cu_seqlens`-host-built path.
3. `[model] perf: drop n_image/video_tokens host-device sync in Qwen3.5-VL forward`.
4. `[model] fix: import gather_outputs/slice_input_tensor from the right module in qwen3_5 patchgen` — patchgen import cleanup that came with commit 2.
5. `[model] chore: regenerate qwen3_5/qwen3_5_moe patched modeling + add missing vision_attention import` — mechanical cleanup after rebase onto current main (regenerated `.diff` files + added the `qwen3_5_vision_attention_forward_patched` import to `qwen3_5_moe_gpu_patch_gen_config.py` which the cherry-pick referenced but didn't import).
6. `[model, ci] test: drop dead _update_linear_attn_mask entries from sync gate allowlist` — the consequence of the fix in #760's gate.
7. `[model, ci] fix: avoid .cuda substring in patchgen comment for device-api check` — `tests/special_sanity/check_device_api_usage.py` does plain substring matching on `.cuda` / `"cuda"` / `"nccl"` (comments included). Rephrased one comment to `set_sync_debug_mode("warn")` to clear the check; no behaviour change.
8. `[model, perf] perf: drop second per-forward grid_thw sync + host-side max_seqlen in Qwen3.5 ViT` — addresses PR review. Adds `override_method` for `Qwen3_5VisionModel.rot_pos_emb` to accept the host-materialised list (reused via name_map by the npu / moe-gpu / moe-npu configs since upstream `Qwen3_5MoeVisionModel.rot_pos_emb` is byte-identical), and replaces the `(cu_seqlens[1:] - cu_seqlens[:-1]).max().item()` reduce with a pure-Python `max(h*w for grid_thw_list, pad_seq_len)`.

### Test

- `pytest tests/models/test_model_forward_no_implicit_sync.py` — 3 passed in ~12s on A100. Allowlist for qwen3_5* is now empty; the sync sites the gate previously listed under `HF-prod-pending-fix` are gone.
- `make quality` — passes.
- `python -m veomni.patchgen.check_patchgen` — all generated files in sync with their patchgen configs.

### API and Usage Example

No public API change. Patchgen-internal: adds `Qwen3_5(Moe)TextModel._update_linear_attn_mask` and `Qwen3_5VisionModel.{forward, fast_pos_embed_interpolate, rot_pos_emb}` to the `Methods patched:` list. NPU configs reuse the GPU helpers via Python import, so the same fix flows through automatically.

### Design & Code Changes

- `veomni/models/transformers/qwen3_5/qwen3_5_gpu_patch_gen_config.py` — adds the `qwen3_5_text_model_update_linear_attn_mask` / `qwen3_5_vision_model_fast_pos_embed_interpolate` / `qwen3_5_vision_model_forward` / `qwen3_5_vision_model_rot_pos_emb` / `qwen3_5_vision_attention_forward_patched` helpers + their `@config.override_method` registrations.
- `veomni/models/transformers/qwen3_5/qwen3_5_npu_patch_gen_config.py` — register the same 5 overrides on NPU (functions reused from the GPU module).
- `veomni/models/transformers/qwen3_5_moe/qwen3_5_moe_{gpu,npu}_patch_gen_config.py` — mirror: import the GPU helpers and register equivalent overrides.
- `veomni/models/transformers/qwen3_5{,_moe}/generated/patched_modeling_qwen3_5{,_moe}_{gpu,npu}.{py,diff}` — regenerated via `python -m veomni.patchgen.check_patchgen --fix`. Only these 8 files change in `generated/` (other models' generated/ stays untouched).
- `tests/models/test_model_forward_no_implicit_sync.py` — `_ALLOWED_SYNCS` becomes empty (the 3 `HF-prod-pending-fix` entries for `_update_linear_attn_mask` are dropped; the surrounding comment block updated to record that the previous entries were cleared by this branch).

### Safety

- `grid_thw.tolist()` is one synchronising H2D-then-D2H copy of a tiny `(num_images, 3)` int tensor; downstream code is the same arithmetic on Python ints — numerically identical, GPU-only behaviour unchanged. The host-built `cu_seqlens` is exactly `F.pad(repeat_interleave(h*w, t).cumsum(0), (1, 0))`.
- `rot_pos_emb` — the patched body is byte-equivalent to upstream's once `grid_thw_list` is in hand (`tolist()` of the same tensor is the only change). The `torch.is_tensor(grid_thw)` fallback preserves the original semantics for external callers (unit tests, ad-hoc scripts) that still pass a raw tensor.
- `vision_max_seqlen` — `max(h*w for grid_thw_list, pad_seq_len)` equals `(cu_seqlens[1:] - cu_seqlens[:-1]).max()` by construction: `cu_seqlens` is built by appending `h*w` deltas for each `(t,h,w)` in `grid_thw_list` plus (under SP) one final `pad_seq_len` segment. The consumer-side fallback in `Qwen3_5VisionAttention.forward` (recomputes from `cu_seqlens` if the kwarg is absent) is unchanged, so external callers behave identically.
- `_update_linear_attn_mask` — prefill/training returns the (possibly all-ones) 2-D mask: at worst one no-op elementwise multiply per linear-attn layer when it is all-ones; cached forwards still return `None`; identical to upstream in every case.
- `n_*_tokens` — derived from `image_grid_thw` shapes the data pipeline already passes in; same value the GPU `.item()` would have produced.

### Follow-up work (not in this PR)

- The corresponding VL/Omni patches (`qwen3_vl`, `qwen3_vl_moe`, `qwen3_omni_moe`) have the same pattern of VeOmni-touched syncs (21 sites) and are tracked separately: see #761 (the gate extension that surfaces them) and `tingyang/cpu_sync_fix/fix-vl-omni.md`.

- [x] Added/updated documentation (patchgen-internal change; no docs/ touched)
- [x] Added tests to CI workflow (validated by the sync gate from #760; this PR's commit 6 cleans up the corresponding allowlist entries)


Luosuu (approved):

Coach257 (approved):

veomni/models/transformers/qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_npu.py
```diff
@@ -1455,9 +1459,15 @@ def __init__(self, config, *inputs, **kwargs) -> None:
...
+    def rot_pos_emb(self, grid_thw) -> torch.Tensor:
         merge_size = self.spatial_merge_size
-        grid_thw_list = grid_thw.tolist()
+
+        # Modification: reuse the host-materialized grid metadata from the caller when available.
+        # The upstream body always starts with `grid_thw.tolist()`, which means the patched ViT
+        # forward was synchronizing the same tensor twice per call after its own
+        # `grid_thw_list = grid_thw.tolist()`. Accept a Python list directly and fall back to
+        # `.tolist()` only for non-patched callers that still pass the raw tensor.
+        grid_thw_list = grid_thw.tolist() if torch.is_tensor(grid_thw) else grid_thw
```
ZihaoW123, line 1470: `grid_thw.tolist() if torch.is_tensor(grid_thw) else grid_thw appears` in multiple places (fast_pos_embed_interpolate, rot_pos_emb). Consider extracting this into a small helper function (e.g., _to_host_list) to reduce code duplication and improve maintainability.

TimYangst: Thanks for the suggestion! I've noted it, and I will do the extraction once I found more vl model could reuse the function.

PR #771, opened 2026-05-19
[model, ops] feat: chunked fused-linear top-k forward-KL distillation kernel

Luosuu:
## Summary

- Add `chunk_topk_distill_function` — a chunked fused-linear top-k forward-KL kernel that lets a downstream verl VeOmni engine compute distillation loss **without materializing the `[T, V]` student logits tensor**. Rides the same chunked autograd pattern as `chunk_logprobs.py`; reuses its per-token NLL / entropy helpers to keep bitwise-vs-verl on the shared outputs.
- Route `return_log_probs=True` + `teacher_topk_ids` + `teacher_topk_log_probs` (+ optional `log_prob_min_clamp`) through `ForCausalLMLoss` / `_chunk_loss_dispatch` to the new kernel. Loss wrappers now return a uniform 7-tuple `(loss, logits, log_probs, entropy, distillation_losses, student_mass, teacher_mass)`; non-distill paths fill trailing slots with `None`.
- Add `distillation_losses` / `student_mass` / `teacher_mass` (`Optional[Tensor]`, fp32, `labels` shape) to `CausalLMOutputWithLogProbs`, `MoeCausalLMOutputWithLogProbs`, and all per-model subclasses (Qwen2-VL, Qwen2.5-VL, Qwen3-VL, Qwen3-VL-MoE, Qwen2.5-Omni, Qwen3-Omni-MoE). `student_mass` / `teacher_mass` are detached (metrics-only, matching verl's contract).
- Update every per-model patchgen config to unpack the 7-tuple and thread the three new fields into the output dataclass. Inline output-dataclass declarations (`add_helper_after`) also extended. Regenerated all 22 modeling files via `make patchgen`.

## What verl needs

Two additive patches documented in `docs/design/verl_topk_distill_integration.md` — both symmetric with the existing `output.log_probs` / `output.entropy` handling, no impact on the non-fused path:

1. `verl/workers/engine/veomni/transformer_impl.py:prepare_model_inputs` — thread teacher tensors into `model_inputs` when `distillation_use_topk=True` (~15 lines).
2. `verl/workers/engine/fsdp/transformer_impl.py:prepare_model_outputs` — extend the `use_fused_kernels=True` arm to read the three new fields off `output` and wrap as nested tensors (~15 lines).

No changes to `compute_forward_kl_topk`, `distillation_ppo_loss`, `DistillationConfig`, micro-batch fields, or estimator-only loss modes (k1/k2/k3/kl/...).

## Numerics

Kernel forward-KL formula is verbatim verl's `kl_divergence(log_q, log_p) = (exp(log_p) * (log_p - log_q)).sum(-1)` on the teacher's top-k support. Backward is closed-form: three additive `dlogits` contributions (log_probs / entropy / topk-KL) matmul-folded to `(dhidden, dweight)`, identical dtype/temperature op order to `chunk_logprobs.py` so the bitwise-vs-verl story carries over on the shared outputs.

## Test plan

- [x] `pytest tests/ops/test_chunk_topk_distill.py` — 9 passed locally, 1 skipped (`verl` import — `ray` not in venv). On H100 worker: 9 passed + 1 skipped.
- [x] `pytest tests/ops/test_chunk_logprobs.py` (regression) — 7 passed + 3 skipped (verl-equivalence subcases gated on same `ray` import). On H100: 7 passed + 3 skipped.
- [x] `pytest tests/models/test_return_log_probs_e2e.py` — 8/8 pass on H100, including the new `test_return_log_probs_with_topk_distill_populates_three_fields` parametrized over `qwen3-text` and `qwen3_vl-vlm`. Asserts: three new fields populated, `distillation_losses ≥ 0`, `student_mass`/`teacher_mass ∈ [0, 1]`, mass tensors detached, **bitwise-equal to the kernel called directly on `model.lm_head.weight`**, backward through `distillation_losses` reaches `lm_head.weight.grad`.
- [x] `make patchgen && git diff` — only the expected per-model unpack + output-dataclass changes (plus a small amount of pre-existing patchgen formatting drift against current HF v5.2.0).
- [x] `make quality` clean.
- [ ] Verl-side dry run once the two integration patches land in a verl branch.

## File-by-file

| File | Change |
|------|--------|
| `veomni/ops/kernels/cross_entropy/chunk_topk_distill.py` | **new** kernel (`_ChunkedLinearTopkDistill` autograd Function + wrapper) |
| `veomni/ops/kernels/cross_entropy/__init__.py` | dispatch branch + 7-tuple return shape |
| `veomni/utils/model_outputs.py` | three new fields × 8 dataclasses |
| `veomni/models/transformers/*/*_patch_gen_config.py` | per-model unpack + output-dataclass extensions (17 configs) |
| `veomni/models/transformers/*/generated/*.py` | `make patchgen` regenerated (do not hand-edit) |
| `tests/ops/test_chunk_topk_distill.py` | **new** kernel tests (correctness + closed-form backward + clamp + IGN masking + verl equivalence) |
| `tests/models/test_return_log_probs_e2e.py` | **new** e2e contract test on `qwen3` and `qwen3_vl` |
| `docs/design/verl_topk_distill_integration.md` | **new** integration guide with the exact two verl-side patches |


veomni/ops/kernels/cross_entropy/chunk_topk_distill.py
```diff
@@ -0,0 +1,409 @@
...
+                    student_log_probs = logits.log_softmax(dim=-1)
+                    student_topk_lp = student_log_probs.gather(dim=-1, index=ids_chunk)
+                    active = (student_topk_lp >= ctx.log_prob_min_clamp).float()
+                else:
+                    active = None
+
+                # Dense ``p_t[v]`` via scatter_add (handles duplicate ids
+                # by summing — the chance verl emits duplicates is zero
+                # in practice but the math is well-defined either way).
+                p_t_k = tlp_f32.exp()  # [chunk, K]
```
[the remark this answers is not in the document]
Luosuu: @gemini-code-assist fixed.

TimYangst (approved):
Overall looks good to me, however please wait for other reviewers' feedback.

Coach257:
### Follow-up idea: nest per-token fused outputs instead of growing every `*OutputWithLogProbs`

PR #771 adds three more top-level optional fields on every output dataclass (`distillation_losses`, `student_mass`, `teacher_mass`) on top of the existing `log_probs` / `entropy` pattern. That works for landing verl top-k distill, but the flattening approach doesn't scale — the next metric (value head, reward, another KL variant) would again touch all `*OutputWithLogProbs` subclasses + every patchgen `forward` + verl reads.

**Proposal (ModelOutput-only refactor; loss wrapper internals can stay as-is for now):**

1. Introduce a single payload type, e.g. **`FusedLinearAuxOutput`**, holding all per-token tensors produced by the fused-linear path (`log_probs`, `entropy`, and the distill trio from this PR). New fields only extend this one dataclass.

2. Add **`FusedLinearAuxOutputMixin`** with one field: `fused_linear_aux: Optional[FusedLinearAuxOutput] = None`. Every `CausalLMOutputWithLogProbs` / `Qwen*OutputWithLogProbs` / `MoeCausalLMOutputWithLogProbs` inherits the mixin + its HF base — **no repeated optional tensor fields on each subclass**.

3. Patched `forward` keeps calling `loss_function` as today (still the 7-tuple from the wrapper), then packs slots 3–7 via `FusedLinearAuxOutput.from_loss_slots(...)` and passes **one** kwarg into the output constructor: `fused_linear_aux=...`.

4. **verl** (small additive change, symmetric with existing fused-kernel handling): read `output.fused_linear_aux.log_probs` / `.entropy` / `.distillation_losses` instead of top-level `output.log_probs`, etc. No change to `compute_forward_kl_topk` or distillation config — only the read path moves one level down.

**Why not change `ForCausalLMLoss` in the same PR:** this PR is already large (kernel + patchgen + e2e). The pack step in `forward` is a few lines per model; we can later shorten the wrapper return to `(loss, logits, fused_linear_aux)` without touching kernels.

**Non-goals:** no long-lived `@property` aliases on `output.log_probs` (avoid two APIs); seq-cls outputs stay unchanged (no fused-linear aux path).

Happy to do this as an immediate follow-up PR on top of #771 if folks agree on the naming (`fused_linear_aux` + `FusedLinearAuxOutput`).

tests/models/test_return_log_probs_e2e.py
```diff
@@ -398,3 +398,147 @@ def test_return_log_probs_backward_flows_gradients(toy_path, family):
...
+    wrapper routes through ``chunk_topk_distill_function`` and assigns
+    ``output.distillation_losses`` / ``output.student_mass`` /
+    ``output.teacher_mass`` automatically.
+
+    This test pins exactly that contract end-to-end on a real toy model
+    (both text and VLM families):
+
+    1. The three new ``output`` fields are populated, finite, and have
+       the same shape as the existing ``log_probs`` / ``entropy``.
+    2. ``output.distillation_losses >= 0`` (forward KL is non-negative).
```
FoolPlayer: Why distillation_losses >= 0

veomni/ops/kernels/cross_entropy/__init__.py
```diff
@@ -303,38 +353,71 @@ def _resolve_cross_entropy_fn(impl: str) -> Callable:
...
         labels = kwargs.get("labels")
         if hidden_states is None or weights is None:
             raise ValueError("return_log_probs=True requires hidden_states and weights (fused-linear path).")
+        if (teacher_topk_ids is None) != (teacher_topk_log_probs is None):
+            raise ValueError(
+                "teacher_topk_ids and teacher_topk_log_probs must be provided together for "
+                "the top-k distillation path."
+            )
+        if teacher_topk_ids is not None:
+            log_probs, entropy, distill, student_mass, teacher_mass = chunk_topk_distill_function(
```
FoolPlayer, line 414: it seems `chunk_size` didn't used here

FoolPlayer (approved):

veomni/ops/kernels/cross_entropy/__init__.py
Luosuu: Good catch — fixed in 26242a2. The wrapper now pops `chunk_size` from kwargs (default 1024 to preserve current behavior) and forwards it to both `chunk_topk_distill_function` and `chunk_logprobs_function`. `ForSequenceClassificationLoss` also pops it defensively so callers that always forward it don't leak it into the inner CE kernel.

tests/models/test_return_log_probs_e2e.py
Luosuu: Fair question — added an explanatory comment in 26242a2. Truncated-support forward KL `Σ_k p_t,k · (log p_t,k - log q_s,k)` is not unconditionally non-negative (it would be for full-support KL via Gibbs; on the top-k subset the contributions can be of either sign). In this test the teacher tensors come from `log_softmax(teacher_logits).topk(K)` so the top-k slots carry the peakiest teacher mass, and the student is a random-init small model with near-uniform log-probs `~ log(1/V)`. Per-slot `log p_t,k > log q_s,k` holds, so per-position sum is `>= 0`. The assertion is intentional: it pins the `(log p_t - log q_s)` sign convention and catches a kernel regression where someone accidentally flips it.

Luosuu:
Thanks @Coach257 — the `FusedLinearAuxOutput` mixin proposal makes sense and I agree the flat-field pattern won't scale to the next metric (value head, reward, alt-KL). Keeping #771 as-is to land verl top-k distill without re-touching every patchgen `forward`, and happy to do the mixin refactor as the immediate follow-up PR with naming `fused_linear_aux: FusedLinearAuxOutput`, leaving the wrapper return shape alone for now (separate PR can shorten it to `(loss, logits, fused_linear_aux)`).

PR #787, opened 2026-05-23
[trainer, ops] feat: rework MoE load-balance monitor (model-agnostic, EP/DP-aware)

Luosuu:
## Summary

The previous `veomni/utils/moe_monitor.py` was dead code: the architecture depended on each model's router patch calling `register_forward_hook(router_forward_hook)` at construction, but **no patch in the repo ever did**. Only `Qwen3MoeTopKRouter` even returned `router_indices` in its forward output, and there was no all-reduce across DP/EP, so the heatmap would have been per-rank even if the hook had fired. Enabling `moe_load_balance_monitor_interval > 0` produced only a "no recorded data" warning at step time.

This PR replaces it with a working, model-agnostic, distributed-aware design covering **Qwen3-MoE family and DeepSeek-V3** out of the box.

### Two-track recognition

The Qwen3 family and DeepSeek-V3 surface their routing decisions at different points in the forward path. `attach_moe_router_monitor(model, monitor)` handles both:

- **`ROUTER_EXTRACTORS` registry** — routers whose forward output exposes `[num_tokens, top_k]` indices directly (Qwen3 family: `Qwen3MoeTopKRouter`, `Qwen3VLMoeTopKRouter`, `Qwen3OmniMoeTopKRouter`). A forward hook reads `output[2]` and accumulates.
- **`EXTERNAL_RECORD_ROUTERS` set** — routers whose patched MoE block calls `record_router_indices(self.gate, topk_indices)` explicitly (DeepSeek-V3, whose top-k math — sigmoid + `e_score_correction_bias` + group-limited top-k — lives in `DeepseekV3MoE.route_tokens_to_experts`, not in the router itself). No hook is registered; the layer is pre-registered so heatmap row order is stable across resumes.

`record_router_indices` is symmetric to the existing `maybe_replay_indices` — same on/off singleton gating, no-op when no monitor is active or it's paused, safe to leave at the call site on non-monitored runs.

### Other changes

- **DP+SP all-reduce of counts** before normalization, using `parallel_state.fsdp_group` (the `dp_sp` mesh dim — every rank with a distinct token slice). EP siblings replicate the router gate and produce identical counts, so they are deliberately excluded to avoid inflating by `ep_size`.
- **Pause / resume / sticky disable** for RL phase scoping (`pause` during rollout, `resume` during update; `disable` survives subsequent `resume` calls).
- **Backend-agnostic `compute_metrics(step, format_only_on)`** so non-source ranks participate in the all-reduce without paying for the heatmap build. Caller wraps the returned dict (PIL image + scalars) for whatever logger backend it has.
- **Fail loud, not fall back**: a registered extractor returning `None` is now an `AssertionError`; matplotlib is a hard dep (`pyproject.toml`), not optional.

The trainer callback no longer requires `wandb.enable` to construct the monitor; only the `wandb.log` emission is gated by it. `on_train_begin` attaches the hooks (so the model is fully parallelized first) and self-disables cleanly if zero routers are recognized.

`make patchgen` regenerated the DeepSeek-V3 GPU/NPU modeling files for the new `DeepseekV3MoE.forward` patch; no other family's generated files changed.

## Future work

- **verl `VeOmniEngine` integration.** The pause/resume/disable surface and the backend-agnostic `compute_metrics(step, format_only_on=...)` were designed to drop into verl's RL engine: monitor created in `_init_moe_monitor()` with `dp_group = parallel_state.fsdp_group`, lazily attached on the first `forward_backward_batch`, paused when `forward_only=True` (rollout / log-prob) and resumed when `forward_only=False` (update). The interval counter increments per `train_batch`; on each interval the engine merges `compute_metrics(...)` into `outputs["metrics"]` on global rank 0 only, so verl's existing `Tracking` logger picks it up for free. This wiring is staged locally and will land in a separate verl PR.
- **GLM-MoE-DSA, Qwen3.5-MoE coverage.** Same pattern as DeepSeek-V3 — register the router class in `EXTERNAL_RECORD_ROUTERS` and insert one `record_router_indices(self.gate, topk_indices)` line in the patched MoE block.

## Test plan

- [x] `pytest tests/utils/test_moe_monitor.py tests/utils/test_moe_monitor_wandb_smoke.py` — 19 unit + wandb-offline-smoke tests, all green (includes DeepSeek-V3-style external-record path + paused/inactive no-op + idempotent re-registration)
- [x] `pytest tests/utils/test_moe_router_replay.py tests/models/test_moe_router_replay_invariants.py` — 16 passed, 4 skipped (replay path untouched)
- [x] `make quality` — clean
- [x] `make patchgen` — DeepSeek-V3 GPU/NPU regen verified; no other family touched
- [x] **End-to-end on mlx worker** against Qwen3-30B-A3B, 8×H100, EP=8, 6 training steps with `moe_load_balance_monitor_interval=2`:
  - `MoE router monitor created: num_experts=128, interval=2, ep_size=8`
  - `attached to 48 router module(s)` (every MoE layer)
  - 3 logging events at steps 2 / 4 / 6 with sensible early-training imbalance:
    | step | max_vio.max | avg_vio.max | avg_vio.avg |
    |------|-------------|-------------|-------------|
    | 2    | 14.62       | 1.54        | 1.42        |
    | 4    | 15.00       | 1.75        | 1.51        |
    | 6    | 15.00       | 1.80        | 1.60        |
  - 3 well-formed 1280×960 RGBA heatmap PNGs in the wandb offline run
  - Clean shutdown, `EXIT=0`


Luosuu:
Addressed both Gemini findings in `5009c05`:

1. `_register_layer` now dedupes against `_layer_order` directly (was checking `_counts`, which is only populated by the first `record()` call). Re-attaching to the same model is now idempotent.
2. `_stack_and_reduce` synthesizes a zero row for any layer present in `_layer_order` but absent from `_counts`. Conditionally-routed / partial-forward layers no longer crash `compute_metrics`; they just show up cold in the heatmap.

Two new unit tests (`test_attach_is_idempotent`, `test_unfired_layers_appear_as_zero_rows`) cover both paths. Full suite: 16 passed, `make quality` clean.

Luosuu:
Landed DeepSeek-V3 coverage in `48ebdc6`:

- New `EXTERNAL_RECORD_ROUTERS` set for families whose top-k math lives downstream of the router (`DeepseekV3TopkRouter` registered).
- New `record_router_indices(router_module, indices)` helper, symmetric to `maybe_replay_indices` — same on/off singleton gating, no-op when paused/inactive.
- Patched `DeepseekV3MoE.forward` (both GPU and NPU configs) calls `record_router_indices(self.gate, topk_indices)` right after `route_tokens_to_experts` returns. `make patchgen` regenerated only the DeepSeek-V3 files; no other family touched.
- 3 new tests cover the external-record path end-to-end, the paused/inactive no-op, and idempotent re-registration. Full suite: 19 monitor tests pass; replay/invariant tests still green; `make quality` clean.

So the monitor now works out of the box for **Qwen3-MoE family and DeepSeek-V3**. GLM-MoE-DSA / Qwen3.5-MoE can follow the same one-line `record_router_indices` pattern when needed (called out in the updated PR description under Future work).

TimYangst (approved):

PR #789, opened 2026-05-25
 [data, model] feat: Implemented dynamic batch logic optimization

zhihaofang1017:
### What does this PR do?
When dynamic batching is enabled, we find that the dynamic batching time is too long in long sequence scenarios. In the 128k scenario of qwen35-35B, profiling revealed a significant processing time before each step, accounting for approximately half of the single-step time.
image
This PR adds a background prefetcher class to preload batches of data in a background thread, allowing data loading to overlap with GPU computation, thereby improving training efficiency.
In the original training, the assembly line was:
`Load data → GPU computation → Load data → GPU computation → ...`
Our expectation is this:
```
Backend:   Load Data 1 →    Load Data 2   →         Load Data 3       → ... 
Foreground:                 GPU Computation 1 → GPU Computation 2 → ...
```
I made the following main modifications.
1. Added BackgroundPrefetcher logic to `veomni/trainer/vlm_trainer.py`
2. Weight saving and loading are performed in `_load_checkpoint` and `_save_checkpoint` within `veomni/trainer/callbacks/checkpoint_callback.py`.
3. Add CI test for the code（next PR）

> Concise overview of the change. Reference related issues/PRs.

### Checklist Before Starting

### Test
```
- Hardware：Ascend 910B
- CANN：8.5.0
- main software version：
torch                         2.7.1+cpu
torch_npu                     2.7.1
triton-ascend                 3.2.0
transformers                  5.2.0
```
I'm using the qwen3.5-0.8B model for validation, mainly focusing on the following logic:
1. The modified loss needs to be aligned with the original loss.
2. After saving checkpoints, the loss should be aligned.
3. After loading checkpoints, the loss should be perfectly aligned.
4. Ensure that dynamic batch operations are completely masked.

The loss and profiling of the original logic:
image
image

The loss  and profiling after modifying dynamic batch:
image
image
imageReload the logic after the checkpoint：
image

I find that the precision can be perfectly aligned, and the time of dynamic batches can be completely masked.

### API and Usage Example
Test script for qwen3.5-0.8B：
```
cat > /tmp/tulu_sharegpt4v_llavavideo.yaml << 'EOF'
sources:
- /mnt/bn/lrm-hl/shared/data/sharegpt4v_instruct_gpt4-vision_cap100k_coco.json
- /mnt/bn/lrm-hl/shared/data/tulu-3-sft-mixture/data
- /mnt/bn/lrm-hl/shared/data/llava-video-178k/0_30_s_academic_v0_1/video.json
names:
- sharegpt4v_captioner_sft
- tulu-3-sft-mixture
- LLaVA-Video-178K
schedule:
- schedule_type: const
  weights: [0.4, 0.2, 0.4]
EOF

cat > /tmp/training_config.yaml << 'EOF'
model:
  model_path: /mnt/bn/llm-lr-hl/shared/models/Qwen3.5-0.8B
  ops_implementation:
    attn_implementation: veomni_flash_attention_2_with_sp
    rms_norm_implementation: npu
    rotary_pos_emb_implementation: npu
    cross_entropy_loss_implementation: chunk_loss
    swiglu_mlp_implementation: eager
    moe_implementation: fused_npu
    rms_norm_gated_implementation: eager
    causal_conv1d_implementation: eager
    chunk_gated_delta_rule_implementation: eager

data:
  train_path: /tmp/tulu_sharegpt4v_llavavideo.yaml
  data_type: conversation
  chat_template: qwen3vl
  max_seq_len: 131072
  train_size: 80000000
  source_name: sharegpt4v_sft
  dataloader:
    num_workers: 8
    use_background_prefetcher: true
  mm_configs:
    image_max_pixels: 602112 # 28 * 28 * 768
    video_max_pixels: 602112 # 28 * 28 * 768
    max_frames: 16
    fps: 2.0
    use_audio_in_video: false

train:
  accelerator:
    ulysses_size: 2
    fsdp_config:
      fsdp_mode: fsdp2
  gradient_checkpointing:
    enable_reentrant: false
  wandb:
    enable: false
    name: Qwen3.5-08B-sft-vision-npu-2n-sp1
    project: fzh-Veomni-test
  freeze_vit: false
  optimizer:
    lr: 1.0e-5
    lr_decay_style: cosine
  num_train_epochs: 1
  micro_batch_size: 1
  max_steps: 50000
  init_device: meta
  profile:
    enable: true
    start_step: 10
    end_step: 11
    record_shapes: true
    with_stack: false
  checkpoint:
    output_dir: /opt/tiger/newveomni/Open-VeOmni/checkpoint
    manager: dcp
    save_steps: 20
    save_epochs: 1
    save_hf_weights: true
EOF

MULTI_STREAM_MEMORY_REUSE=2 bash train.sh tasks/train_vlm.py /tmp/training_config.yaml 

```

### Design & Code Changes

- Applied pre-commit checks
- Added/updated documentation


veomni/trainer/vlm_trainer.py
```diff
@@ -30,6 +36,85 @@
 from .base import BaseTrainer, _collect_muon_kwargs
 
 
+logger = helper.create_logger(__name__)
```
FoolPlayer: Duplicate logger with line 118

veomni/trainer/vlm_trainer.py
```diff
@@ -12,6 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import queue
+
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
```
FoolPlayer: Duplicate with Line11-13

veomni/trainer/vlm_trainer.py
```diff
@@ -332,7 +417,9 @@ def train(self):
             self.on_epoch_begin()
 
             # Create a batch generator
-            data_iterator = iter(self.base.train_dataloader)
+            # data_iterator = iter(self.base.train_dataloader)
```
FoolPlayer: Remove this line

veomni/trainer/vlm_trainer.py
```diff
@@ -30,6 +36,85 @@
...
+                raise StopIteration
+            if isinstance(res, Exception):
+                raise res
+            return res
+
+    def state_dict(self):
+        import torch.distributed as dist
+
+        if dist.is_available() and dist.is_initialized():
+            dist.barrier()
```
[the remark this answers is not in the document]
FoolPlayer: Agree. It make sense.

veomni/trainer/vlm_trainer.py
```diff
@@ -332,7 +417,9 @@ def train(self):
             self.on_epoch_begin()
 
             # Create a batch generator
-            data_iterator = iter(self.base.train_dataloader)
+            # data_iterator = iter(self.base.train_dataloader)
+            data_iterator = BackgroundPrefetcher(self.base.train_dataloader)
```
FoolPlayer: Maybe remove every `BackgroundPrefetcher` after the current epoch ends

veomni/trainer/vlm_trainer.py
```diff
@@ -30,6 +36,85 @@
...
+
+        if dist.is_available() and dist.is_initialized():
+            dist.barrier()
+        if self.current_state is not None:
+            return self.current_state
+        if self.original_state_dict:
+            return self.original_state_dict()
+        return {}
+
+    def stop(self):
```
FoolPlayer: I think we need add some timeout setting

veomni/trainer/callbacks/checkpoint_callback.py
```diff
@@ -105,15 +105,20 @@ def _save_checkpoint(self, state: TrainerState):
...
             "optimizer": self.trainer.optimizer,
             "extra_state": {
                 "global_step": state.global_step,
                 "lr_scheduler": self.trainer.lr_scheduler.state_dict(),
-                "train_dataloader": self.trainer.train_dataloader.state_dict()
-                if self.trainer.train_dataloader is not None
-                else {},
+                "train_dataloader": train_dataloader_state,
                 "environ_meter": self.trainer.environ_meter.state_dict(),
                 "torch_rng_state": torch.get_rng_state(),
```
FoolPlayer, line 126: If save ckpt and prefetch dataloader at the same time, the prefetch action may change rng state. Maybe stop prefetch when saving.

tests/models/test_vlm_trainer.py
```diff
@@ -73,3 +74,39 @@ def test_freeze_vit_on_vlm_model(config_path, freeze_vit):
...
+    def __iter__(self):
+        for item in self.data:
+            self.state["count"] += 1
+            yield item
+
+    def state_dict(self):
+        return self.state.copy()
+
+
+def test_background_prefetcher():
```
FoolPlayer: The test seems not cover some traininig case, maybe we need add more case for this PR, you can do at the next PR.

veomni/trainer/vlm_trainer.py
```diff
@@ -30,6 +36,85 @@
 from .base import BaseTrainer, _collect_muon_kwargs
 
 
+logger = helper.create_logger(__name__)
+
+
+class BackgroundPrefetcher:
```
FoolPlayer: I think we can add it on BaseTrainer and add extra args to select it

FoolPlayer (changes requested):

Luosuu:
cc @LiuzcEECS 
please take a look

veomni/trainer/vlm_trainer.py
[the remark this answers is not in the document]
zhihaofang1017: it has been removed

veomni/trainer/vlm_trainer.py
zhihaofang1017: done

veomni/trainer/vlm_trainer.py
zhihaofang1017: done

veomni/trainer/vlm_trainer.py
zhihaofang1017: I have removed the `BackgroundPrefetcher` at the end of each epoch

veomni/trainer/vlm_trainer.py
zhihaofang1017: I have added 5s timeout for this stop function

veomni/trainer/callbacks/checkpoint_callback.py
zhihaofang1017: I think I have already considered this. When I save ckpt, I'm actually saving the dataloader from the previous step. I specifically considered this in _save_checkpoint, so I saved the current state and the original state in BackgroundPrefetcher; therefore, the prefetch operation shouldn't change the rng state

tests/models/test_vlm_trainer.py
zhihaofang1017: OK, i have deleted the CI tests, and i can consider more tests for next PR

veomni/trainer/vlm_trainer.py
zhihaofang1017: I'm agree with you, so I move this main class to BaseTrainer, and also add an extra args(use_background_prefetcher) to control whether to use BackgroundPrefetcher

veomni/trainer/base.py
```diff
@@ -656,7 +733,10 @@ def train(self):
             self.on_epoch_begin()
 
             # Create a batch generator
-            data_iterator = iter(self.train_dataloader)
+            if args.data.dataloader.use_background_prefetcher:
```
FoolPlayer: Great! Can we combine these 2 iter into  `checkpoint_callback.py` as like `VeOmniIter` and use same api in training loop, it's easier for adding new trainer.

veomni/trainer/vlm_trainer.py
```diff
@@ -27,7 +27,7 @@
 from ..utils.device import synchronize
```
FoolPlayer, line 27: Only adding BackgroundPrefetcher in vlm_trainer now, please support BackgroundPrefetcher in other trainers as well.

veomni/trainer/vlm_trainer.py
```diff
@@ -332,7 +332,12 @@ def train(self):
             self.on_epoch_begin()
 
             # Create a batch generator
-            data_iterator = iter(self.base.train_dataloader)
+            if args.data.dataloader.use_background_prefetcher:
+                data_iterator = BackgroundPrefetcher(self.base.train_dataloader)
+            else:
+                data_iterator = iter(self.base.train_dataloader)
+
+            self.base.data_iterator = data_iterator
```
FoolPlayer: Why set `self.base.data_iterator = data_iterator` here but not in others

veomni/trainer/callbacks/checkpoint_callback.py
```diff
@@ -109,26 +105,25 @@ def _save_checkpoint(self, state: TrainerState):
 
         save_checkpoint_path = os.path.join(args.train.checkpoint.save_path, f"global_step_{state.global_step}")
 
+        if hasattr(self.trainer, "data_iterator") and hasattr(self.trainer.data_iterator, "state_dict"):
```
FoolPlayer, line 111: If using prefetch but not set `data_iterator` in trainer, dataloader state will use `trainer.train_dataloader.state_dict()` which is the next batch state when saving ckpt (prefetch batch n+1 when fwd batch n, prefetcher holding state batch n, but batch n+1 for train_dataloader). We need fix it in BaseTrainer, and maybe add assert or raise error for that case. Now it's a silent bug.

veomni/trainer/base.py
zhihaofang1017: I have added an iterator called `VeOmniIter` in` base.py`

veomni/trainer/vlm_trainer.py
zhihaofang1017: done

veomni/trainer/callbacks/checkpoint_callback.py
zhihaofang1017: This is indeed a problem I didn't fully consider. I simply added `self.base.data_iterator = data_iterator` to `vlm_trainer` without giving it much thought. The current solution is to directly mount `data_iterator` as an instance property `self.data_iterator` (or `self.base.data_iterator` for composite classes like `Trainer`).

FoolPlayer (approved):

LiuzcEECS:
@zhihaofang1017 Hi Zhihao, excuse me for the late reply. I would like to explain what we have done in https://github.com/ByteDance-Seed/VeOmni/pull/603 and how that might help.

We have 2 "dyn_bsz_runtime" mode:

1. [Default] "main", where the dyn batch construction is running in a single process instead of multi-worker processes. 

Under the "main" mode, from the profile timeline screenshot, I guess the processes work like this:
image

You can see there are two major problems in the process:
1. Because the batching logic is written in the main process, the main process is actually IDLE when the GPU is working
2. Although we have 8 workers, because the seq is very long, if you check what happened between T4 and T5, the whole(100%) buffer is not enough for one batch seq, so it is IDLE before batching but only resumes working when batching starts, as the result, GPU is not well utilized and has lots of IDLE blocks

2. "worker" mode, added in the PR #603 . Under the "worker" mode, multi-worker processes are working on the fetching-batching loop
image

You can see that the GPU is fully utilized and it is well parallelized with dataloading because now the workers' buffer is enough for `pytorch_prefetch_factor*num_workers` batch seq for the GPU to process, and also the batching is parallelized under multiple workers, which also speeds up the data processing.

@zhihaofang1017 I sincerely suggest you to try disabling use_background_prefetcher and using "worker" as "dyn_bsz_runtime" to see if it can resolve the issue because
(1) we are actually going to deprecate "main" mode in future
(2) if it can help, we can resolve your performance issue in a simpler way
cc: @FoolPlayer

PR #803, opened 2026-05-28
[ckpt] refactor: deduplicate ExtraParallel dim preprocessing

Luosuu:
## Summary

``ModelState.get_state_dict_with_extra_parallel_dim_preprocess`` and ``OptimizerState.get_state_dict_with_extra_parallel_dim_preprocess`` were ~80% the same code (130 lines total). Extract the shared logic into two module-level helpers:

- ``_validate_extra_parallel_meshes`` — fail-fast precondition checks (previously duplicated as ~25 lines of mesh-dict construction + 4 asserts in both methods)
- ``_apply_extra_parallel_dim`` — the drop/restore loop, with a ``key_match`` parameter to switch between exact (model) and substring (optimizer) FQN lookup

Both wrappers become one-liners delegating to the shared function.

### Net change

```
1 file changed, 100 insertions(+), 120 deletions(-)
```

### Behavior

No functional change.  The non-tensor and 0-D guards previously only applied in the OptimizerState path are now applied unconditionally — these are safe no-ops on model state dicts because model params are never non-tensors or 0-D scalars.

The only meaningful difference between the two callers is how state-dict keys map to ExtraParallel FQNs:

| Path | Key format | Lookup mode |
|------|-----------|-------------|
| Model | ``"model.layers.0.mlp.experts.gate_proj"`` (FQN is the key) | ``key_match="exact"`` |
| Optimizer | ``"state.model.layers.0.mlp.experts.gate_proj.exp_avg"`` (FQN is substring) | ``key_match="substring"`` |

## Test plan

- [x] Unit tests in ``tests/checkpoints/`` — 23 pass + 2 xfail
- [x] E2E ``test_trainer_saveload[qwen3_moe-1]`` (no EP, ExtraParallel preprocessing not exercised) — pass on 8× H100
- [x] E2E ``test_trainer_saveload[qwen3_moe-8]`` (ep_size=8, exercises the refactored ExtraParallel path) — pass on 8× H100
- [x] ``make quality`` passes

### DIFF
--- a/veomni/checkpoint/dcp_checkpointer.py
+++ b/veomni/checkpoint/dcp_checkpointer.py
@@ -51,6 +51,94 @@ _EXTRA_STATE_FORMAT = "extra_state_rank_{}.pt"
 _EXTRA_STATE_DIR = "extra_state"
 
 
+def _validate_extra_parallel_meshes(parallel_state) -> None:
+    """Fail-fast precondition for ExtraParallel state dict preprocessing.
+
+    At least one ExtraParallel mesh must be non-None, and at least one
+    of those meshes must be 2D (the ExtraParallel + FSDP composition).
+    """
+    extra_parallel_mesh = {
+        para: parallel_state.extra_parallel_fsdp_device_mesh[para][para]
+        if parallel_state.extra_parallel_fsdp_device_mesh[para] is not None
+        else None
+        for para in parallel_state.extra_parallel_names
+    }
+    assert any(m is not None for m in extra_parallel_mesh.values()), (
+        "At least one extra_parallel mesh should be not None"
+    )
+    assert any(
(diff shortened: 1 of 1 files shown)


Luosuu:
The flagged behaviour matches the pre-refactor code byte-for-byte — both `ModelState` and `OptimizerState` were already passing `spec_info.para_fsdp_mesh[f"{para_name}_fsdp"]` as the third argument to `restore_extra_parallel_dim`. The refactor preserves this exactly. E2E `test_trainer_saveload[qwen3_moe-8]` (which exercises the ExtraParallel preprocessing path) passes on 8× H100, so the FSDP+EP path is verified. If the no-FSDP-but-EP-only branch in `restore_extra_parallel_dim` has a real bug, it's a pre-existing one — out of scope for this dedup PR.

Coach257 (approved):

PR #805, opened 2026-05-28
Feat/fused ulysses qkv projection

Butterfingrz:
This PR is intended to explore the optimization potential of communication-computation fusion in training workloads, especially the possibility of using PyTorch Symmetric Memory to build fine-grained fused communication-computation operators.

## Acknowledgement

Special thanks to [[@Gin-Sin](https://github.com/Gin-Sin)](https://github.com/Gin-Sin), who originated the symmetric-memory–based Ulysses fused-kernel idea this PR builds and adapts on, presented at the NVIDIA CUDA meetup.

## TL;DR

This PR introduces a **fused qkv-proj GEMM + all-to-all** kernel for the Ulysses pre-attention path. Design rationale:

1. **Async-Ulysses underperforms on GQA.** Q-side GEMM FLOPs and all-to-all payload dwarf KV's, so the asymmetric Q work has no peer to hide behind and the overlap window collapses.

2. **Timeline overlap ≠ throughput overlap.** NCCL collectives are GPU kernels that contend with GEMMs for SM cycles; `torch.profiler` shows all-to-all firing late and the two streams throttling each other. CUDA green contexts may ease this but not remove it.

3. **Splitting qkv into 3 GEMMs is wasteful** — a single fused qkv GEMM is strictly more efficient.

4. **Async-Ulysses cannot remove the pre/post reshuffles** Ulysses inherently requires.

Fusing the qkv GEMM with the all-to-all into one kernel sidesteps cross-stream SM contention, recovers single-GEMM efficiency, and absorbs the reshuffle into the epilogue.

## Summary

This PR adds the new package `veomni/ops/kernels/fused_ulysses_projection/` — a single-kernel **GEMM + all-to-all** for the Ulysses pre-attention path — plus the framework glue that drops it into the existing `AsyncUlyssesQKVProjection` without touching model authors' code. `qwen3_vl` and `qwen3_vl_moe` are the reference integrations.

- fused kernel arch 
fused_kernel_host_design

### 1. Kernel layer — `_kernels/kernel/`

* `gemm_ws_packqkv.py`: `GemmWsPreAttnPackQKVSm90` extends quack 0.3.4's persistent `GemmSm90` via a new `PackQKVEpilogueMixin`. The GEMM main loop is unchanged; the epilogue is rewritten so each output sub-tile is **TMA-stored directly into the destination peer's symmetric-memory `[Q | K | V]` slot** instead of into a local `mD`. Routing ("which sub-tile → which peer's Q/K/V") is derived from the GEMM N-axis coordinate by `input_nh_to_dst_rank_and_pack_head` (`gemm_ws_packqkv.py:30-59`), so the all-to-all *is* the epilogue store path.

* `quack_patch.py`: minimally invasive monkey-patch on `GemmSm90.__call__/kernel/epilogue` that inserts four named hooks (`epi_get_output_tensor`, `epi_has_output_stage`, `epi_make_peer_gmem_copy_fns`, `epi_store_subtile`). The mixin overrides only these — we never fork quack.

* `pack_qkv_scheduler.py`: `PackQKVTileScheduler` round-robins N-axis cluster emission across rank-sections so TMA pushes spread across NVLinks instead of hammering one peer per phase; identity fallback when a section has fewer clusters than `world_size` (the GQA K/V edge case).

- custom scheduler round robin optimization
custom_scheduler_v2

* `gpu_barrier.py`: `GpuBarrierAll` — a PTX system-scope CAS kernel launched post-GEMM that spins on per-peer sync slots in symm-mem. Replaces the implicit NCCL barrier a standalone all-to-all would provide.

- Framework integration of fused kernels
fused_kernel_pipeline

### 2. Custom op + state layer — `ws_push.py`, `state_manager.py`, `_config.py`

* `torch.ops.ulysses.ws_push_gemm_a2a` is the registered custom op; `ws_push_forward_impl` validates state, hands symm-mem views to the kernel, and returns zero-copy `(q, k, v)` views over the receive buffer (cloned by default to survive grad-accum overwrites).

* `WsPushState` owns the symm-mem `blob` (`peer_out | pad | sync` regions), per-peer 2D views, the int64 sync-slot pointer table for `GpuBarrierAll`, and a CPU metadata tensor crossing the custom-op boundary. `WsPushStateManager` is the process-wide holder; its `__init__` is **collective** — picks tile config via `choose_tile_config(local_seq, world_size)`, calls `init_ws_push_state` (`_symmetric_memory.empty + rendezvous`) across the SP group, zeroes sync slots, and barriers. `is_compatible(bs, local_seq, dtype)` is the only per-forward check; structural quantities would have failed at init. (`_active_impl_name` flips at config; `_active_manager` is published in `on_train_begin`, since symm-mem needs the SP group.)

* `WsPushDispatch.try_resolve_auto / try_resolve_fused` returns `None` (→ eager) on no-manager / shape-mismatch, else a token `(state, W_qkv_B, bias_B)`.

* `_config.py` registers `ulysses_qkv_projection` as a GLOBAL `OpSpec` with backends `eager` / `ws_push`. Its `_ws_push_side_effect` runs a preflight (CUDA + SM90 + `torch.distributed._symmetric_memory`) so bad hosts fail at config-resolution, not deep in the kernel.

### 3. Integration into `AsyncUlyssesQKVProjection` (`veomni/distributed/sequence_parallel/async_ulysses.py`)

The autograd `Function.forward` gains one branch (`async_ulysses.py:263-291`): if `dispatch is not None`, call `validate_ws_push_dispatch(...)` + `_qkv_via_ws_push(...)` (one fused kernel call); else fall back to 3×`F.linear` + 3× async-a2a. The signature is extended with `qkv_weight / qkv_bias / n_q / n_kv` (mutually exclusive with the legacy `q/k/v_weight` triplet); `_split_fused_qkv_views` slices Q/K/V views off the packed weight *only* for the eager fallback and the backward GEMMs. Backward stays on the existing 3×GEMM + 3× async-a2a path (fused backward is out of scope); a fused-signature branch at the tail of `backward` (`async_ulysses.py:562-594`) re-packs the three partial weight grads into one contiguous `grad_qkv_weight` via slice-`copy_` so the gradient lands on the single `FusedQKVLinear.weight` parameter.

### 4. Framework wiring — `FusedQKVLinear`, patchgen, callback, checkpoint converter

* **`FusedQKVLinear`** (`veomni/models/modules/fused_qkv_linear.py`) — one `nn.Parameter weight` of shape `[n_q*hd + 2*n_kv*hd, hidden]`, row-concat `[Q | K | V]`. `from_separate_linears(q, k, v, head_dim)` does the host-side `torch.cat` **once at model init**, so FSDP2 sees one shard unit. Its own `forward` is the eager fallback; ws_push bypasses it and reads `self.weight` directly.

* **Patchgen** (`qwen3_vl_gpu_patch_gen_config.py`, reused by `qwen3_vl_moe_gpu_patch_gen_config.py` via `name_map={"Qwen3VL": "Qwen3VLMoe"}`):

  * `modify_init` wraps HF `__init__` to build `self.qkv_proj = FusedQKVLinear.from_separate_linears(...)` and `del self.q_proj, self.k_proj, self.v_proj`.
  * `override_method` swaps in `_qwen3_vl_async_ulysses_attention_forward`, which calls `async_ulysses_qkv_projection(qkv_weight=self.qkv_proj.weight, n_q=…, n_kv=…, …)` and **never** passes `dispatch=` — model code stays oblivious to the feature flag.

* **Checkpoint conversion** — `Qwen3VLAttentionCheckpointTensorConverter` regex-matches HF's `{q,k,v}_proj.{weight,bias}` and emits one `qkv_proj.{weight,bias}` via `torch.cat(dim=0)` at load time. For `qwen3_vl_moe`, a thin composite (`Qwen3VLMoeFullCheckpointTensorConverter`) routes each tensor to either this QKV converter or the existing MoE expert-layout converter, since the loader accepts only one converter per model. HF export is handled offline by `scripts/qkv_split_dcp_to_hf.py`.

* **Callback** (`fused_ulysses_state_callback.py`, instantiated unconditionally at `base.py:462`):

  * `__init__` silently self-disables (`_disabled_reason`) unless `ulysses_qkv_projection_implementation == "ws_push"`, then snapshots `(sp_group, bs, local_seq, dtype, nheads_q/k/v, head_dim)` via `_prepare_init_kwargs` (warning_rank0 for `dyn_bsz=True` / mismatched `pad_to_length`, which would silently waste the symm-mem buffer).
  * `on_train_begin` collectively builds `WsPushStateManager(...)`, runs `all_reduce(local_ok, MIN, sp_group)` so any single-rank failure cleanly falls **all** ranks back to eager (no next-collective deadlock), then `set_active_manager(...)`.
  * `on_train_end` → `manager.teardown()` (`cuda.synchronize → dist.barrier(sp_group) → state.close`) + `clear_active_manager()`.

### End-to-end (`qwen3_vl_moe`)

```text
YAML: ulysses_qkv_projection_implementation=ws_push
  ① register_op + apply_global_ops → _active_impl_name="ws_push"; preflight CUDA/SM90/symm-mem
  ② patchgen modify_init       → self.qkv_proj = FusedQKVLinear (one Parameter, row-concat QKV)
     patchgen override_method  → forward calls async_ulysses_qkv_projection(qkv_weight=…)
  ③ FusedUlyssesStateCallback.on_train_begin → WsPushStateManager(...)
        → init_ws_push_state (symm-mem empty+rendezvous) → all_reduce(MIN) → set_active_manager
  ④ per forward → async_ulysses_qkv_projection
        → WsPushDispatch.try_resolve_auto → try_resolve_fused (get_active_manager + is_compatible)
        → AsyncUlyssesQKVProjection.apply → validate_ws_push_dispatch (14 gates)
        → _qkv_via_ws_push → ws_push_forward_impl → torch.ops.ulysses.ws_push_gemm_a2a
        → GemmWsPreAttnPackQKVSm90 (PackQKV epilogue TMA-push to peer symm-mem) + GpuBarrierAll
  ⑤ backward → eager 3×GEMM + 3× async-a2a; fused-signature branch re-packs to grad_qkv_weight
  ⑥ on_train_end → teardown + clear_active_manager
```

Five-layer guardrail: preflight → callback warning → `all_reduce(MIN)` → `is_compatible` → `validate_ws_push_dispatch`.

## Numerics

The fused kernel does not alter the GEMM math; it only redirects WHERE each epilogue sub-tile is stored:

* quack 0.3.4's persistent SM90 main loop, K-axis reduction tree, and fp32→bf16 quantization are **unchanged**. `PackQKVEpilogueMixin` overrides only the four named destination hooks (`epi_get_output_tensor`, `epi_has_output_stage`, `epi_make_peer_gmem_copy_fns`, `epi_store_subtile`), deciding sub-tile **destination**, not **value**.
* The PackQKV TMA-store is **bit-exact data movement** (routing via `input_nh_to_dst_rank_and_pack_head`, `gemm_ws_packqkv.py:30-59`); `PackQKVTileScheduler` is a cluster-ID permutation; `GpuBarrierAll` is a PTX CAS spin. None touch the payload.

Hence the only place fused-vs-eager numerics can diverge is the fact that the fused path runs **one** GEMM `A @ [Q|K|V]_B^T` while the eager path runs **three** separate `F.linear` calls — a property of GEMM fusion, not of communication fusion.

## Tests

Two parity tests against eager 3×`F.linear` + 3× async-a2a, bf16, `atol=1e-2 rtol=1e-2`:

* **`tests/parallel/ulysses/test_fused_ulysses_pre_attn.py`** — forward parity, ws=2, GQA `nheads_q=8 / nheads_kv=4`, `head_dim=64`, `local_seq=64`, ± bias + RMSNorm.
* **`tests/parallel/ulysses/test_fused_vs_eager_numerical.py`** — forward + backward sweep at `VEOMNI_TEST_WORLD_SIZE ∈ {2,4,8}`; adds `unpadded%ws ≠ 0` coverage + buffer-alias safety; prints per-tensor noise stats so regressions surface as measurable drift.

Launch:

```bash
pytest tests/parallel/ulysses/test_fused_ulysses_pre_attn.py
VEOMNI_TEST_WORLD_SIZE=8 pytest tests/parallel/ulysses/test_fused_vs_eager_numerical.py -s
```

## Profiler

End-to-end perf was captured on 1 node × 4 H200 via `bash scripts/multimodal/train_qwen3_vl_moe_ws_push.sh` (`DummyVLMTrainer` + `configs/multimodal/qwen3_vl/qwen3_vl_moe_ws_push_dummy.yaml`). The built-in `ProfileTraceCallback` captures one step at `profile.start_step=20, end_step=22` on rank 0. Full rationale: `dev_md/ulysses/PR_doc/train_qwen3_vl_moe_ws_push.md`.

### Configuration

| Field                                   | Value             | Why this value                                                                                      |
| --------------------------------------- | ----------------- | --------------------------------------------------------------------------------------------------- |
| `ulysses_qkv_projection_implementation` | `ws_push`         | The only switch that arms the fused kernel                                                          |
| `ulysses_size`                          | `4`               | Capped at `num_key_value_heads=4`; ws_push has no `need_repeat_kv` workaround                       |
| `dyn_bsz` / `pad_to_length`             | `false` / `false` | `(bs, local_seq, dtype)` is frozen at `on_train_begin`; `dyn_bsz=true` silently falls back to eager |
| `micro_batch_size`                      | `1`               | Must equal the frozen `bs`; mismatch silently falls back to eager                                   |
| `max_seq_len`                           | `131072`          | Must be divisible by `ulysses_size`                                                                 |

Model is **Qwen3-VL-MoE 30B-A3B**; data is fully synthetic via `build_dummy_dataset(task_type='qwen3vl')`, so the profiler captures kernel and collective time only.

fused_ulysses_profiler

baseline_ulysses_profiler

### Results

Same config, same seed; `ws_push` on vs. off (off = the async-Ulysses fallback — the `dispatch=None` path of `AsyncUlyssesQKVProjection`):

| Metric                                | async-Ulysses | ws_push (fused) | Δ                           |
| ------------------------------------- | ------------- | --------------- | --------------------------- |
| pre-attention `qkv_proj + all_to_all` | 2.500 ms      | 1.325 ms        | **−1.175 ms (−47%, 1.89×)** |
| training step (e2e)                   | 17.497 ms     | 17.405 ms       | −0.092 ms                   |

The pre-attention block sees the intended ≈2× speedup from collapsing 3× GEMM + 3× async-a2a into one SM-resident kernel. The step-level delta is much smaller — pre-attention is only one phase of the MoE step (MoE FFN, attention compute, backward, optimizer all unchanged), and on this 30B-A3B/sp=4/128K config the rest of the step dominates the wall clock.

### Framework integration

* The fused kernel needs the qkv-proj weight **pre-packed**. Packing has to move out of the forward path into **model init** — `FusedQKVLinear` does a real host-side `torch.cat` once per layer at startup, never per step. Requires patchgen to wire it in.
* Re-exporting back to HF Hub needs an **offline split** (`scripts/qkv_split_dcp_to_hf.py`); it cannot be done by an automatic module hook.
* Invasive modification of `AsyncUlyssesQKVProjection` in `async_ulysses.py` is unavoidable.
* No framework-level symmetric-memory manager exists yet. Allocation/teardown sits outside the standard `nn.Module` forward allocation cycle and must be driven explicitly; lifetime is not auto-released the way regular PyTorch tensors are.

### Operator

* Under GQA, `sp_size` is capped at `num_kv_heads`: the fused kernel does not implement `repeat_kv`, so `sp_size > num_kv_heads` is unsupported.

* The kernel returns **views** into symm-mem, so the `AsyncUlyssesQKVProjection` forward must `.clone()` them to survive grad-accum overwrites. (Unavoidable for training; inference could skip the 3× D2D copy cost.)

* Performance is dominated by the underlying GEMM launch params. Today only a heuristic pick between **pingpong** and **cooperative** schedulers (driven by `local_seq` under Ulysses sp) with fixed tile sizes is in place — no autotuning. Hardware with different compute capability (e.g., H20) requires separate tuning; reusing H200's GEMM launch configuration as-is may degrade performance.

* Symm-mem capacity is sized from `(bs, local_seq, dtype)`. With `dyn_bsz=True`, `pad_to_length=True` must be set so the actual `local_seq` on any sp-rank cannot exceed the allocated buffer.


Butterfingrz:
@Luosuu  @FoolPlayer  @Fazziekey  Hi , could you please take a look at this PR . I would love to discuss it with everyone. Thanks!

Luosuu:
@Butterfingrz Thank you for the amazing work! I will take a look soon. I am curious whether the performance gain is related to the sequence length and do we have benchmarks for very long context e.g. 128K inputs?

Butterfingrz:
@Luosuu I think the correlation between the current operator’s performance benefit and the sequence length should be considered from the following two aspects， the benefits of fusing Ulysses QKV-proj GEMM with all-to-all mainly come from two aspects：

1. Under GQA, the output dimension along the N dimension of the KV-proj GEMM is very small. When the M dimension, namely the local sequence length, is very large, the GEMM shape becomes tall and skinny. Packing the QKV projection into a single GEMM makes the shape more favorable for matrix multiplication.

2. The fused kernel implements an operation equivalent to the all-to-all transpose in the epilogue by combining TMA store with symmetric memory. Therefore, to hide the communication overhead during GEMM tile computation, the ping-pong scheduling strategy is preferred initially, as it can almost fully hide the all-to-all overhead in the epilogue.

2.1. However, when `local_seq` is relatively large, a larger GEMM tile has to be selected to prioritize GEMM execution efficiency. In this case, only the cooperative scheduling strategy can be used. Due to the nature of cooperative scheduling, each CTA can process multiple GEMM tiles, so the prologue overhead can be hidden starting from the second tile. However, the epilogue cannot be overlapped. Still, thanks to inter-warp communication-computation fusion, the overhead of launching a separate communication kernel and reloading the data is eliminated, so the performance improvement remains significant.


**Benchmark script:**

```bash
ULYSSES_WORLD_SIZE=8 pytest tests/parallel/ulysses/test_pre_attn_sweep.py
```

pre_attn_sweep_P8_20260529_085856

Butterfingrz:
The purple result in the benchmark is intended to separately evaluate the performance overhead introduced by `.copy()`. There is still room for further optimization in the operator. Before training starts, some tile and cluster shapes can be configured through the autotuning functionality inside `quack-kernel`. The current operator does not apply shape-specific tuning; it only uses two simple sets of launch parameters for the ping-pong and cooperative strategies, selected based on `local_seq`, that is, the size of the M dimension.

veomni/models/transformers/qwen3_vl_moe/qwen3_vl_moe_gpu_patch_gen_config.py
```diff
@@ -53,6 +53,7 @@
     qwen3_vl_model_get_placeholder_mask_patched,
     qwen3_vl_rmsnorm_forward_patched,
     qwen3_vl_text_attention_forward_patched,
+    qwen3_vl_text_attention_init_patched,
```
Luosuu, line 56: This symbol does not exist in `qwen3_vl_gpu_patch_gen_config.py`, so `make patchgen` will fail. Also, the reused forward helper below still calls `self.q_proj/k_proj/v_proj`, while the generated MoE file deletes those modules and uses `self.qkv_proj`; please add a MoE-specific fused-QKV forward helper and verify regeneration.

veomni/distributed/sequence_parallel/async_ulysses.py
```diff
@@ -393,6 +555,48 @@ def backward(ctx: Any, *grad_output: Tensor):
...
+                *ref.shape[:-2],
+                total_out,
+                in_dim,
+                dtype=ref.dtype,
+                device=ref.device,
+            )
+            grad_qkv_weight[..., :_q_out, :].copy_(grad_q_weight)
+            grad_qkv_weight[..., _q_out : _q_out + _kv_out, :].copy_(grad_k_weight)
+            grad_qkv_weight[..., _q_out + _kv_out :, :].copy_(grad_v_weight)
+            if grad_q_bias is not None and grad_k_bias is not None and grad_v_bias is not None:
```
Luosuu, line 582: In fused-signature mode the separate `q_bias/k_bias/v_bias` inputs are `None`, so the earlier `ctx.needs_input_grad[3/5/7]` gates prevent these partial bias grads from being created. `qkv_bias` then receives no grad. Please gate fused bias grads on the `qkv_bias` input slot and add a `FusedQKVLinear(bias=True)` backward test.

veomni/models/transformers/qwen3_vl_moe/__init__.py
```diff
@@ -33,7 +37,9 @@ def register_qwen3_vl_moe_modeling(architecture: str):
         )
 
     for model_cls in (Qwen3VLMoeForConditionalGeneration, Qwen3VLMoeModel, Qwen3VLMoeTextModel):
-        model_cls._create_checkpoint_tensor_converter = staticmethod(create_qwen3_vl_moe_checkpoint_tensor_converter)
+        model_cls._create_checkpoint_tensor_converter = staticmethod(
```
Luosuu, line 40: This registers the QKV-fusing converter for NPU too, but the NPU generated model still has separate `q_proj/k_proj/v_proj` modules. On NPU, HF keys will be converted to unexpected `qkv_proj.*` keys. Please make the QKV converter GPU-only unless the NPU model is also fused.

veomni/distributed/sequence_parallel/async_ulysses.py
```diff
@@ -523,7 +732,29 @@ def async_ulysses_qkv_projection(
...
     group: Optional[ProcessGroup] = None,
+    dispatch: Optional["WsPushDispatch"] = None,
+    # ---------- Fused signature (FusedQKVLinear callers) ----------
+    qkv_weight: Optional[Tensor] = None,
+    qkv_bias: Optional[Tensor] = None,
+    n_q: Optional[int] = None,
+    n_kv: Optional[int] = None,
 ):
+    # Transparent dispatch: when the caller exposes a fused ``qkv_weight``
+    # (the FusedQKVLinear path used by qwen3_vl / qwen3_vl_moe) and did not
```
Luosuu, line 743: This says the fused `qkv_weight` path is used by both `qwen3_vl` and `qwen3_vl_moe`, but only the MoE generated GPU file passes `self.qkv_proj.weight`. Plain Qwen3-VL still passes separate q/k/v weights, so `ws_push` silently falls back to eager there. Please either wire Qwen3-VL through `FusedQKVLinear` too or narrow the claim/docs to MoE only.

.agents/knowledge/architecture.md
```diff
@@ -141,6 +143,12 @@ VeOmni uses FSDP2 exclusively.
...
 
+## Fused Ulysses Pre-attention Kernel (opt-in, Hopper SM90+)
+
+When `model.ops_implementation.ulysses_qkv_projection_implementation: ws_push` is set, Ulysses pre-attention QKV projection + first all-to-all is replaced by a single fused kernel at `veomni/ops/kernels/fused_ulysses_projection/`. The kernel uses PyTorch symmetric memory to push the per-rank Q/K/V tile directly to peer buffers in the GEMM epilogue (one system-scope CAS barrier instead of NCCL a2a). Lifecycle is owned by `FusedUlyssesStateCallback` (`on_train_begin` allocates symm-mem, `on_step_end` invalidates cached `W_qkv_B`, `on_train_end` tears down). Dispatch is transparent: `async_ulysses_qkv_projection(qkv_weight=...)` queries `WsPushDispatch.try_resolve_fused` and falls back silently to the eager 3× a2a path when the active manager is absent or batch shape is incompatible. Models opt in by installing `FusedQKVLinear` (single `[n_q*hd + 2*n_kv*hd, H]` Parameter) via the patchgen `INIT_MODIFICATION` primitive; on-disk checkpoints stay in fused layout, with HF-format export handled by the offline `scripts/qkv_split.py` (mirroring `scripts/moe_ckpt_merge/moe_split.py`).
```
Luosuu, line 148: This lifecycle text is stale: the callback no longer invalidates a cached `W_qkv_B`, and the export script is named `scripts/qkv_split_dcp_to_hf.py`, not `scripts/qkv_split.py`. Please update the docs so future agents do not rely on the removed cache behavior.

Luosuu (commented):
Concise inline review comments for the main blocking issues I found.

Luosuu (commented):
overall I think this is a very good high level research idea. However, from maintainability perspective it is better to prevent monkey-patching the quack cute-dsl kernel package. For very long context training, we may seek solutions like magi-attention instead of the async ulysses. Thank you very much for your efforts but I think it would be hard to merge this PR considering the challenges in maintainability.

Butterfingrz:
Thanks for the review and feedback. I fully agree that we should not introduce a quack-kernel-version-dependent monkey patch into Veomni, as it would go against the current design goals of flexibility and modularity. Async Ulysses was first proposed in Veomni, and this PR is mainly intended to further explore the potential of the Async Ulysses paradigm, particularly its possibilities for deeper integration of communication and computation.

Butterfingrz:
By the way, I’m very interested in the Veomni project, and I’d be happy to contribute to future roadmap items related to kernel and operator optimization. I’m still learning and trying to keep up with the latest techniques in training-kernel and operator development.

I’ve been researching magi-attention, and my main takeaway is that CP design for ultra-long-context multimodal training may require framework-level support. Thanks again for your suggestion! I would really appreciate a community-facing contribution list if one becomes available.

PR #820, opened 2026-06-03
[ops, ci] feat: extend NPU kernel test coverage and add CPU-runnable registry/eager sanity tests

Kirrito-k423:
## Summary

This PR adds extended unit-test coverage for the NPU kernels shipped under `veomni/ops/`, plus a set of CPU-runnable sanity tests for the kernel registry and eager-reference helpers. All tests pass on an Ascend 910 cluster and a subset (registry sanity + eager reference) runs on plain CPU so the suite can be wired into any CI runner.

### 1. Extended NPU numerical tests - `tests/ops/test_npu_kernels_extended.py` (+326 lines, 26 tests)

Broadens the existing NPU kernel tests with production-sized shapes, additional dtypes, and edge cases that the previous suite did not cover:

- **RMSNorm** - hidden sizes 256 / 1024 / 2048 in both `bfloat16` and `float16`, plus a `qwen3_5` variant at 512/1024, zero-input / unit-weight / eps-robustness edge cases.
- **RoPE** - full and partial variants with `head_dim` 64 and 128, `bfloat16` and `float16`, with both implicit position and explicit `position_ids` paths, plus a pass-through preservation check for the un-rotated tail.
- **RMSNormGated** - shape and dtype contract for batch 2 and 4, zero-gate-output behavior, eps sweep at `1e-5 / 1e-6 / 1e-7`, and `nn.Module` smoke test.

Tolerances are matched to the empirical NPU bf16 ULP drift (e.g. `atol=1e-2` for 256-dim, `atol=5e-2` for 2048-dim); tighter values fail on Ascend 910.

### 2. Kernel-registry sanity tests - `tests/ops/test_kernel_registry_sanity.py` (+247 lines, 23 tests)

CPU-runnable, no device required. Validates the registry contract:

- Every declared `(op, variant)` slot has a corresponding `npu` implementation.
- No duplicate `(op, variant)` registrations.
- `resolve("eager")` returns `None`; `resolve("unknown")` raises `KeyError`.
- `OpSlot` behavior in unbound / bound / error paths.
- `KernelSpec` immutability.
- `HardwareRequirement` parsing for `device_type in {gpu, npu, any}` and `compute_capability` bounds.

### 3. Eager reference tests - `tests/ops/test_eager_kernels_cpu.py` (+153 lines, 9 tests)

CPU-runnable, checks algebraic invariants of `load_balancing_loss_pytorch`:

- Uniform routing => loss equals `top_k`.
- One-hot (imbalanced) routing => loss equals `E`.
- Attention-mask correctly excludes padded positions.
- Layer-independence when the same gate logits are reused.
- `None` / non-tuple inputs return `0` instead of raising.

## Test results

Environment: `root@192.168.13.154` (Ascend 910 x 16), CANN 8.5 / 9.0, `torch_npu 2.7.1.post5`, Python 3.11.

```
$ pytest tests/ops/test_eager_kernels_cpu.py \
         tests/ops/test_kernel_registry_sanity.py \
         tests/ops/test_npu_kernels_extended.py -v

collected 58 items

tests/ops/test_eager_kernels_cpu.py ............ [9 passed]
tests/ops/test_kernel_registry_sanity.py .......................s... [22 passed, 1 skipped]
tests/ops/test_npu_kernels_extended.py .......................... [26 passed]

======================== 57 passed, 1 skipped in 8.36s ==============================
```

The one `skipped` test (`test_resolve_hardware_mismatch_raises_runtimeerror`) is conditional on the running host's hardware - on a non-matching device the registry returns `None` rather than raising, which is the documented behavior. The skip is intentional, not a failure.

## Files

```
 tests/ops/test_eager_kernels_cpu.py      | 153 +++++++++++++++
 tests/ops/test_kernel_registry_sanity.py | 247 +++++++++++++++++++++
 tests/ops/test_npu_kernels_extended.py   | 326 +++++++++++++++++++++++++++
 3 files changed, 726 insertions(+)
```

### DIFF
--- a/tests/ops/test_eager_kernels_cpu.py
+++ b/tests/ops/test_eager_kernels_cpu.py
@@ -1,0 +1,153 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""CPU-runnable tests for the eager reference implementations used by VeOmni.
+
+These tests don't require any accelerator.  They are useful when:
+  * Iterating on kernel math changes locally on a Mac (no NPU).
+  * Catching regressions in the eager fallback that the NPU/GPU
+    kernels are compared against.
+
+The tests are deliberately small so they run in <1s on a laptop.
+
+``load_balancing_loss_pytorch`` expects ``gate_logits`` to be a tuple
+of per-layer tensors with shape ``[batch_size * seq_len, num_experts]``;
+that convention is used throughout these tests.
+
+Algebraic reference (Switch Transformer):
+    loss = E * sum_e(f_e * P_e)
+where f_e = (count of times e is in top-k) / N and P_e = mean probability
+assigned to e.  With uniform probs (1/E), f_e = top_k / E, P_e = 1/E,
+so loss = E * E * (top_k/E) * (1/E) = top_k.
+"""
+
+import pytest
+import torch
+
+import veomni.ops  # noqa: F401 -- trigger KERNEL_REGISTRY registrations
+from veomni.ops.kernels.load_balancing_loss.eager import load_balancing_loss_pytorch
+from veomni.ops.dispatch import OpSlot
+
+
+# ---------------------------------------------------------------------------
+# Load balancing loss (pure-PyTorch) eager reference
+# ---------------------------------------------------------------------------
+
+
+class TestLoadBalancingLossEager:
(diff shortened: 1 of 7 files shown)


tests/ops/test_eager_kernels_cpu.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    def test_uniform_distribution_well_conditioned(self):
+        """With uniform random gate logits, the loss is in (0, top_k + 1)."""
+        num_experts, top_k, num_layers, N = 8, 2, 3, 1024
+        torch.manual_seed(0)
+        gate_logits = tuple(torch.randn(N, num_experts, dtype=torch.float32) for _ in range(num_layers))
+        loss = load_balancing_loss_pytorch(gate_logits, num_experts, top_k, None)
+        assert loss.item() > 0
+        assert loss.item() < 5.0
```
ZihaoW123, line 61: The range (0, 5.0) is logical, but the random inputs make the test non‑deterministic, and the coarse interval provides low precision. Recommendation: Use a fixed random seed (e.g., torch.manual_seed(42)) to generate reproducible inputs, and assert against a concrete expected loss value for that specific input to improve test precision.

Crystal-jiang (approved):
look good to me

tests/ops/test_eager_kernels_cpu.py
Kirrito-k423: I will fix this in next PR

PR #833, opened 2026-06-12
[data, trainer] feat: support effective-token dynamic batching

zhangxin81:
## Summary

Add a new `train.dyn_bsz_count_mode` switch (`"total"` | `"effective"`, default `"total"`, fully backward compatible). When set to `"effective"`, dynamic batching seals a micro-batch by **effective tokens** (`labels != IGNORE_INDEX`) instead of total `attention_mask` tokens.

This addresses a long-standing imbalance in SFT / DPO / multimodal pipelines: sequence packing already evens out the **wall-clock compute** across DP ranks, but it does **not** even out the **effective-token count**. Prompt-heavy samples can starve some DP ranks of gradient signal while others train on dense supervision, inflating gradient variance and loss noise even though the global loss reduction is mathematically correct.

### Status quo (`total` mode, current default)

- `DynBszBuffer` / `TextBatchingStrategy` / `DynamicBatchingSizeDataset` all measure per-sample length via `attention_mask.sum()`.
- A micro-batch is sealed once `Σ attention_mask >= micro_batch_size * max_seq_len`.
- Total tokens per DP rank (≈ wall-clock compute) are naturally balanced.
- **However, the effective-token count (loss-mask=1) can differ by tens of times across DP ranks**, e.g.:
  - Rank A: long system prompt + short reply → effective ≈ 200
  - Rank B: pure pre-training long doc → effective ≈ 7900
  - Rank C: many short multi-turn dialogs → effective ≈ 4000

`mean_global_loss` (`veomni/utils/loss_utils.py:54`) already reduces loss with a global effective-token weighting, so the **loss value is correct**. But the per-step gradient signal-to-noise ratio still differs by tens of times across ranks, hurting effective convergence rate.

### New `effective` mode

- Per-sample length becomes `(labels != IGNORE_INDEX).sum()`.
- A micro-batch is sealed once the **effective-token sum** crosses the budget.
- DP-rank effective gradient signal becomes balanced; gradient variance becomes more stable.
- Trade-off: the actual packed sequence may be slightly longer than `max_seq_len` (prompt tokens occupy extra physical space).

## Design

| Counting scheme | Meaning | When to use |
|---|---|---|
| `attention_mask.sum()` | Tokens that participate in attention | Pre-training / CPT, latency-sensitive setups |
| `(labels != IGNORE_INDEX).sum()` | Tokens that contribute to the loss | SFT, DPO, multimodal, prompt-heavy data |

### What this PR explicitly does NOT do

- **No cross-DP all-to-all redistribution.** Sequence packing already balances compute time for text LLMs; adding a runtime redistribute would cost more in communication than it saves. This change only swaps the "is the micro-batch ready?" predicate — zero extra communication.
- **No default behaviour change.** `dyn_bsz_count_mode` defaults to `"total"`; existing configs are unaffected.
- **No change to loss normalization.** `mean_global_loss` is already correct; left untouched.

### New argument

- `veomni/arguments/arguments_types.py`: `TrainingArguments.dyn_bsz_count_mode: Literal["total", "effective"] = "total"`

### Data layer

- `veomni/data/dataset.py`
  - Add a top-level `IGNORE_INDEX` import (avoid per-call import on the hot path).
  - New `get_length_by_labels_fn(sample)`: counts via `labels != IGNORE_INDEX`; gracefully falls back to `attention_mask` when `labels` is missing.
  - New `get_length_fn_by_count_mode(count_mode)`: mode → length-callable selector.
- `veomni/data/dynamic_batching.py`
  - `DynBszBuffer.__init__` now accepts an optional `get_length_fn`; default behaviour unchanged (still uses `attention_mask`).
  - `TextBatchingStrategy.__init__` forwards `get_length_fn` into its internal buffer.
- `veomni/data/data_loader.py`
  - `build_native_dataloader` adds the new `dyn_bsz_count_mode` argument.
  - Both `main` and `worker` dyn_bsz runtimes now use the same length callable.

### Trainer plumbing

- `veomni/trainer/base.py`
- `veomni/trainer/base_rl_trainer.py`
- `veomni/trainer/dit_trainer.py`

Each forwards `dyn_bsz_count_mode=args.train.dyn_bsz_count_mode` to `build_dataloader` / `build_native_dataloader`.

## Backward Compatibility

- Default `count_mode="total"` is bit-for-bit equivalent to the pre-change code path.
- `DynBszBuffer.__init__`'s new parameter defaults to `None`; external callers constructing the buffer directly are not broken.
- No existing YAML config needs updating.

## Usage

```yaml
train:
  dyn_bsz: true
  dyn_bsz_count_mode: effective   # new; defaults to "total"
```

Or via CLI:

```bash
--train.dyn_bsz_count_mode=effective
```

### Recommended setting

| Scenario | Recommended |
|---|---|
| Pre-training / CPT | `total` |
| SFT with stable prompt-to-response ratio | `total` |
| SFT with highly variable prompt length (multi-turn history / long system prompt) | `effective` |
| DPO / preference alignment | `effective` |
| Multimodal SFT (vision tokens are not in the loss) | `effective` |
| NPU / strong wall-clock-balance requirement | `total` |

## Testing

- All edited files pass `python3 -m ast.parse` syntactic sanity check.
- Both `total` and 'effective' mode test existing on: `tests/data/test_dataloader.py`, `tests/data/test_dynamic_batching_dataset.py`, `tests/data/test_multisource_dataset.py`.

### DIFF
--- a/veomni/data/dataset.py
+++ b/veomni/data/dataset.py
@@ -288,7 +288,7 @@ def get_length_by_input_ids_fn(sample):
     return _numel_sequence_like(sample["input_ids"])
 
 
-def get_length_by_labels_fn(sample):
+def get_length_by_labels_fn(sample: Any) -> int:
     """Return effective token length from ``labels`` (i.e. tokens contributing to loss).
 
     A token contributes to loss iff its label is not ``IGNORE_INDEX``. Falls back to
@@ -298,6 +298,13 @@ def get_length_by_labels_fn(sample):
     set additional ``IGNORE_INDEX`` positions later, but balancing is decided here at
     sample ingestion time.
     """
+    if sample is None:
+        return 0
+    if isinstance(sample, list):
+        return sum(get_length_by_labels_fn(item) for item in sample)
+    if not isinstance(sample, dict):
+        return 1
+
     if "labels" in sample:
         labels = sample["labels"]
         if isinstance(labels, torch.Tensor):
@@ -307,7 +314,25 @@ def get_length_by_labels_fn(sample):
         if isinstance(labels, list):
             return sum(1 for label in labels if label != IGNORE_INDEX)
 
-    return get_length_by_attention_mask_fn(sample)
+    if "attention_mask" in sample:
+        attention_mask = sample["attention_mask"]
+        if isinstance(attention_mask, torch.Tensor):
+            return int(attention_mask.sum().item())
+        if isinstance(attention_mask, np.ndarray):
+            return int(attention_mask.sum())
+        if isinstance(attention_mask, list):
+            return int(sum(attention_mask))
+
+    if "input_ids" in sample:
+        input_ids = sample["input_ids"]
+        if isinstance(input_ids, torch.Tensor):
+            return int(input_ids.numel())
+        if isinstance(input_ids, np.ndarray):
+            return int(input_ids.size)
+        if isinstance(input_ids, list):
+            return len(input_ids)
+
+    return 1
 
 
 def get_length_fn_by_count_mode(count_mode: str):
--- a/veomni/data/dataset.py
+++ b/veomni/data/dataset.py
@@ -298,17 +298,16 @@ def get_length_by_labels_fn(sample):
     set additional ``IGNORE_INDEX`` positions later, but balancing is decided here at
     sample ingestion time.
     """
-    if "labels" not in sample:
-        return get_length_by_attention_mask_fn(sample)
+    if "labels" in sample:
+        labels = sample["labels"]
+        if isinstance(labels, torch.Tensor):
+            return int((labels != IGNORE_INDEX).sum().item())
+        if isinstance(labels, np.ndarray):
+            return int((labels != IGNORE_INDEX).sum())
+        if isinstance(labels, list):
+            return sum(1 for label in labels if label != IGNORE_INDEX)
 
-    labels = sample["labels"]
-    if isinstance(labels, torch.Tensor):
-        return int((labels != IGNORE_INDEX).sum().item())
-    if isinstance(labels, np.ndarray):
-        return int((labels != IGNORE_INDEX).sum())
-    if isinstance(labels, (list, tuple)):
-        return sum(label != IGNORE_INDEX for label in labels)
-    return int(labels != IGNORE_INDEX)
+    return get_length_by_attention_mask_fn(sample)
 
 
 def get_length_fn_by_count_mode(count_mode: str):
--- a/veomni/data/dataset.py
+++ b/veomni/data/dataset.py
@@ -252,13 +252,40 @@ class EnergonDataset(IterativeDataset):
             logger.debug(f"Dataset {type(self._data).__name__} does not support set_epoch or state reset")
 
 
+def _sum_sequence_like(value: Any) -> int:
+    """Return the sum of tensor, ndarray, list, tuple, or scalar values."""
+    if isinstance(value, torch.Tensor):
+        return int(value.sum().item())
+    if isinstance(value, np.ndarray):
+        return int(value.sum())
+    if isinstance(value, (list, tuple)):
+        return int(sum(value))
+    return int(value)
+
+
+def _numel_sequence_like(value: Any) -> int:
+    """Return the flattened element count of tensor, ndarray, list, tuple, or scalar values."""
+    if isinstance(value, torch.Tensor):
+        return int(value.numel())
+    if isinstance(value, np.ndarray):
+        return int(value.size)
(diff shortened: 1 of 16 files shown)


veomni/data/dataset.py
```diff
@@ -260,6 +261,38 @@ def get_length_by_attention_mask_fn(sample):
...
+def get_length_by_labels_fn(sample):
+    """Return effective token length from ``labels`` (i.e. tokens contributing to loss).
+
+    A token contributes to loss iff its label is not ``IGNORE_INDEX``. Falls back to
+    ``attention_mask`` when ``labels`` is absent (e.g. some unsupervised pipelines).
+
+    Note: this counts pre-pack labels; ``PackingCollator`` and SP label shifting may
+    set additional ``IGNORE_INDEX`` positions later, but balancing is decided here at
+    sample ingestion time.
+    """
```
TimYangst, line 300: Please refer to  func `_default_sample_token_len`, we should handle the `sample` in different types:  ('torch.Tensor', 'np.ndarray', 'list')

veomni/data/dataset.py
TimYangst, line 1031: When `dyn_bsz_count_mode="effective"`, length only tracks loss-contributing tokens. If we encounter samples with massive prompts but very few labels, this greedy loop will keep packing samples to meet the effective budget, causing the accumulated physical token count to grow unpredictably and trigger out-of-memory (OOM) or CUDA indexing errors. 
I'm wondering if we should introduce a hard physical token cap inside this loop to seal the micro-batch early if the total physical tokens exceed a safe boundary (e.g., micro_batch_size * max_seq_len).

cc: @Coach257

Luosuu:
@LiuzcEECS please take a look

veomni/data/dataset.py
zhangxin81: Modified as above said

veomni/data/dataset.py
zhangxin81: Modified as above said

zhangxin81:
Please take a look when you free, thanks

Coach257:
effective mode is effectively a no-op because the physical cap equals the effective budget

In build_native_dataloader (veomni/data/data_loader.py:170) the physical cap is set to the same value as the effective budget:

`physical_token_cap = batching_token_len if dyn_bsz_count_mode == "effective" else None`

# token_micro_bsz / micro_batch_seq_length is also batching_token_len
So we have effective_budget == physical_cap == B.

There's an invariant that holds for every sample: effective_len (labels != IGNORE_INDEX) <= physical_len (attention_mask.sum()), since loss-contributing tokens are a strict subset of attention-valid tokens. Summed over a batch, cum_effective <= cum_physical.

In DynBszBuffer.get_samples (veomni/data/dynamic_batching.py:95-105) each sample must satisfy both:

```
fits_effective_budget = seq_len      <= B - cum_seq_len        # labels
fits_physical_cap      = physical_len <= B - cum_physical_len   # attention
```

Because physical_len >= seq_len and cum_physical >= cum_effective, the physical constraint is always the tighter one. The micro-batch therefore fills up until cum_physical ≈ B and seals there — which is exactly the sealing criterion of total mode (attention tokens reaching B). The same applies to is_ready_for_micro_batch: physical_ready (all_physical_token_cnt >= B) always fires no later than effective_ready.

Net result: with physical_cap == effective_budget, each micro-batch ends up with ~B physical tokens and essentially the same sample set as total mode, so the effective-token balancing has no observable effect.

This also contradicts the PR's stated trade-off ("the actual packed sequence may be slightly longer than max_seq_len"): that requires the physical cap to be strictly larger than the effective budget, but cap == B forbids exactly that.

Suggested fix: decouple the physical cap from the effective budget — allow controlled physical overflow, e.g. an overflow_ratio (or a separate dyn_bsz_physical_token_cap / max_seq_len multiplier):

```
physical_token_cap = (
    math.ceil(batching_token_len * dyn_bsz_physical_overflow_ratio)
    if dyn_bsz_count_mode == "effective"
    else None
)
```
with overflow_ratio > 1.0 (default e.g. 1.5–2.0). This keeps the OOM guard that was requested earlier while actually letting effective mode pack to a balanced effective-token budget.

zhangxin81:
Thank you for your very constructive feedback; it has been revised accordingly. @Coach257

Coach257:
LGTM

Coach257 (approved):

zhangxin81:
This is my first PR in the veomni codebase. Thank you all for the thorough review — it's much more rigorous and formal than what I typically receive internally at my company, and I've learned a great deal from it. Thanks again！ @Coach257 @Luosuu

PR #853, opened 2026-06-18
[data] feat: support map-style datasets in worker-side dynamic batching

LiuzcEECS:
### What does this PR do?

From the discussion of https://github.com/ByteDance-Seed/VeOmni/pull/789#issuecomment-4606881927, we think it would be better to try to use worker mode dynamic batch size dataset.
However, currently we cannot use map-style datasets together with the worker mode.
Now, the PR is to enable **map-style datasets** on the worker-side dynamic-batching path
(`dyn_bsz_runtime="worker"`). That path previously accepted iterable upstreams
only, so a map-style dataset silently lost its DP-rank sharding and then crashed.

We created a small internal wrapper now re-creates `DistributedSampler` semantics inline, so `DynamicBatchingSizeDataset` can accept map-style datasets while keeping random access.

Related issues/PRs: https://github.com/ByteDance-Seed/VeOmni/pull/789

### Checklist Before Starting

- Search for relative PRs/issues and link here: _none_
- [x] PR title follows `[{modules}] {type}: {description}` — `[data] feat: support map-style datasets in worker-side dynamic batching` (`data` module, `feat` type, non-breaking)

### Test

CPU-only unit tests added to `tests/data/test_datasets.py` (run in CI):

- `test_map_style_sampler_wrapper_rank_parity` — per-rank indices **bit-identical**
  to `torch.utils.data.DistributedSampler` across `shuffle × drop_last ×
  num_replicas × epoch`.
- `test_map_style_sampler_wrapper_worker_split` — disjoint / balanced-within-1 /
  strided worker shares (1 and 8 workers).
- `test_map_style_sampler_wrapper_resume` — end-to-end through
  `_MapStyleSamplerWrapper → DynamicBatchingSizeDataset → StatefulDataLoader`; an
  interrupted-and-resumed run is tensor-identical to the uninterrupted run over
  `num_workers ∈ {0, 2} × save_by_idx ∈ {False, True}`.

Local run: **30 passed**; `test_dynamic_batching_dataset.py` non-distributed
**6 passed** (no regression); `ruff check` clean. No training-curve/eval change —
this only affects data ordering/sharding, validated for exact equivalence above.

### Design & Code Changes

`_MapStyleSamplerWrapper` (`veomni/data/dataset.py`) presents a map-style dataset
as a per-rank, per-worker stateful iterable:

- **Rank split** bit-identical to `DistributedSampler` (`seed + epoch` permutation,
  wrap-around padding / `drop_last` truncation, `indices[rank::num_replicas]`).
- **Worker split** strides the rank's indices by `worker_id` inside `__iter__`
  (`[worker_id::num_workers]`) (`DistributedSampler` has no worker logic; the
  DataLoader's batch round-robin does not apply on the iterable path, so the
  wrapper does it explicitly.)
- **Resume** skips the consumed prefix of the deterministic permutation — the same
  mechanism as `StatefulDistributedSampler`; per-worker `yielded` composes with
  `StatefulDataLoader` snapshots and `DynamicBatchingSizeDataset`'s upstream-state
  nesting for exact mid-epoch resume.

Change list:
- `veomni/data/dataset.py` — add `_MapStyleSamplerWrapper`.
- `veomni/data/data_loader.py` — auto-wrap map-style datasets in the worker
  `dyn_bsz` branch (`shuffle` / `seed` / DP rank-size matching the sampler branch).
- `tests/data/utils.py` — add `ShardedMappingDataset` (reuses
  `ShardedIterableDataset`'s deterministic sample generation).
- `tests/data/test_datasets.py` — add the three tests.

No behavior change for iterable upstreams or `dyn_bsz_runtime="main"`.

- [x] Applied pre-commit checks (`ruff format` + `ruff check` clean)
- [ ] Added/updated documentation — _N/A: internal wrapper, no public API/config/workflow change_
- [ ] Moved/renamed `tasks/` training scripts — _N/A: not touched_
- [x] Added tests to CI workflow (`tests/data/test_datasets.py`)

### DIFF
--- a/tests/data/test_datasets.py
+++ b/tests/data/test_datasets.py
@@ -4,18 +4,33 @@ import random
 import subprocess
 import sys
 from functools import partial
+from types import SimpleNamespace
 from typing import Any, Dict, List
+from unittest.mock import patch
 
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
 
 import pytest
+import torch
 import yaml
 from tools import resolve_ops_overrides
+from torch.utils.data import DistributedSampler
+from torchdata.stateful_dataloader import StatefulDataLoader
 from transformers import PretrainedConfig
-from utils import DummyDataset, FakeModel, compare_global_batch, compare_items, compare_metrics, process_dummy_example
+from utils import (
+    DummyDataset,
+    FakeModel,
+    ShardedMappingDataset,
+    compare_global_batch,
+    compare_items,
+    compare_metrics,
+    process_dummy_example,
+)
 
 from veomni.arguments import parse_args
+from veomni.data.data_collator import MainCollator
+from veomni.data.dataset import DynamicBatchingSizeDataset, _MapStyleSamplerWrapper
 from veomni.distributed.parallel_state import get_parallel_state
 from veomni.trainer.base import BaseTrainer, VeOmniArguments
 from veomni.trainer.callbacks import (
@@ -320,3 +335,114 @@ def test_native_dataset(dataset_type: str, dyn_bsz: bool, dummy_native_dataset_c
     command = build_command(dataset_type, dyn_bsz, data_path=data_path)
     result = subprocess.run(command, check=True)
     assert result.returncode == 0
+
+
+# ---------------------------------------------------------------------------
+# _MapStyleSamplerWrapper unit tests (CPU-only, no torch.distributed init:
+# DistributedSampler is constructed with explicit num_replicas / rank).
+# ---------------------------------------------------------------------------
+
+_WRAPPER_DATASET_SIZE = 50
+_WRAPPER_MICRO_BATCH_SEQ_LENGTH = 64
+_WRAPPER_READY_THRESHOLD = 4
+
+
+def _wrapper_get_length(item):
+    return int(item["attention_mask"].sum())
+
+
+def _wrapper_take_first(micro_batches):
+    # batch_size=1 -> one micro batch per item; module-level so it stays picklable for num_workers > 0.
+    return micro_batches[0]
+
+
+def _wrapper_worker_info(worker_id, num_workers):
+    return None if num_workers == 1 else SimpleNamespace(id=worker_id, num_workers=num_workers)
+
+
+@pytest.mark.parametrize("shuffle", [False, True])
+@pytest.mark.parametrize("drop_last", [False, True])
+@pytest.mark.parametrize("num_replicas", [1, 2, 4])
+@pytest.mark.parametrize("epoch", [0, 3])
+def test_map_style_sampler_wrapper_rank_parity(shuffle, drop_last, num_replicas, epoch):
+    """Per-rank index assignment must be bit-identical to ``DistributedSampler``."""
+    dataset = ShardedMappingDataset(size=_WRAPPER_DATASET_SIZE)
+    seed = 7
+    for rank in range(num_replicas):
+        wrapper = _MapStyleSamplerWrapper(
+            dataset, num_replicas=num_replicas, rank=rank, shuffle=shuffle, seed=seed, drop_last=drop_last
+        )
+        wrapper.set_epoch(epoch)
+        sampler = DistributedSampler(
+            dataset, num_replicas=num_replicas, rank=rank, shuffle=shuffle, seed=seed, drop_last=drop_last
+        )
+        sampler.set_epoch(epoch)
+        assert wrapper._rank_indices() == list(sampler)
+        assert len(wrapper) == sampler.num_samples
+
+
+@pytest.mark.parametrize("num_workers", [1, 8])
+def test_map_style_sampler_wrapper_worker_split(num_workers):
+    """Workers of one rank read disjoint, balanced, strided shares of the rank's indices."""
+    dataset = ShardedMappingDataset(size=_WRAPPER_DATASET_SIZE)
+    wrapper = _MapStyleSamplerWrapper(dataset, num_replicas=2, rank=1, shuffle=True, seed=0)
+    rank_indices = wrapper._rank_indices()
+
+    seen_per_worker = []
+    for worker_id in range(num_workers):
+        worker_wrapper = copy.deepcopy(wrapper)  # real workers iterate their own copy
+        worker_wrapper.output_index_for_resume = True
+        # The wrapper binds get_worker_info via a direct import, so patch it in its own module.
+        with patch("veomni.data.dataset.get_worker_info", return_value=_wrapper_worker_info(worker_id, num_workers)):
+            seen_per_worker.append([idx for _, idx in worker_wrapper])
+
+    counts = [len(seen) for seen in seen_per_worker]
+    assert max(counts) - min(counts) <= 1
+    assert sorted(idx for seen in seen_per_worker for idx in seen) == sorted(rank_indices)
+    for worker_id, seen in enumerate(seen_per_worker):
+        assert seen == rank_indices[worker_id::num_workers]
+
+
+def _build_wrapper_pipeline(num_workers, save_by_idx):
+    """mapping dataset -> _MapStyleSamplerWrapper -> DynamicBatchingSizeDataset -> StatefulDataLoader."""
+    dataset = ShardedMappingDataset(size=_WRAPPER_DATASET_SIZE)
+    wrapper = _MapStyleSamplerWrapper(dataset, num_replicas=2, rank=0, shuffle=True, seed=11)
+    dynamic_ds = DynamicBatchingSizeDataset(
+        dataset=wrapper,
+        micro_batch_seq_length=_WRAPPER_MICRO_BATCH_SEQ_LENGTH,
+        ready_for_micro_batch_threshold=_WRAPPER_READY_THRESHOLD,
+        dynamic_batching_collate_fn=MainCollator(),
+        get_length_fn=_wrapper_get_length,
+        save_by_idx=save_by_idx,
+    )
+    return StatefulDataLoader(dynamic_ds, batch_size=1, num_workers=num_workers, collate_fn=_wrapper_take_first)
+
+
+@pytest.mark.parametrize("num_workers", [0, 2])
+@pytest.mark.parametrize("save_by_idx", [False, True])
+def test_map_style_sampler_wrapper_resume(num_workers, save_by_idx):
+    """An interrupted+resumed run reproduces the uninterrupted batch stream exactly.
+
+    Exercises both the wrapper's own state_dict/load_state_dict (skip the consumed prefix
+    of the deterministic permutation) and its composition with StatefulDataLoader's
+    per-worker snapshots and DynamicBatchingSizeDataset's upstream-state nesting.
+    """
+    golden = list(_build_wrapper_pipeline(num_workers, save_by_idx))
+    assert len(golden) > 4, "test setup must produce enough micro batches"
+
+    dataloader = _build_wrapper_pipeline(num_workers, save_by_idx)
+    it = iter(dataloader)
+    head = [next(it) for _ in range(3)]
+    state = dataloader.state_dict()
+    del it, dataloader
+
+    resumed = _build_wrapper_pipeline(num_workers, save_by_idx)
+    resumed.load_state_dict(state)
+    batches = head + list(resumed)
+
+    assert len(batches) == len(golden)
+    for got, want in zip(batches, golden):
+        assert got.keys() == want.keys()
+        for key in got:
+            if torch.is_tensor(got[key]):
+                assert torch.equal(got[key], want[key]), f"mismatch in {key}"
--- a/tests/data/test_datasets.py
+++ b/tests/data/test_datasets.py
@@ -446,3 +446,34 @@ def test_map_style_sampler_wrapper_resume(num_workers, save_by_idx):
         for key in got:
             if torch.is_tensor(got[key]):
                 assert torch.equal(got[key], want[key]), f"mismatch in {key}"
+
+
+def test_map_style_sampler_wrapper_state_dict_after_resume_before_iter():
+    """state_dict() right after load_state_dict() but before __iter__ must report the
+    restored progress, not 0.
+    """
+    dataset = ShardedMappingDataset(size=_WRAPPER_DATASET_SIZE)
+    kwargs = dict(num_replicas=2, rank=0, shuffle=True, seed=5)
+
+    src = _MapStyleSamplerWrapper(dataset, **kwargs)
+    src.set_epoch(2)
+    it = iter(src)
+    consumed = 3
+    for _ in range(consumed):
+        next(it)
+    state = src.state_dict()
+    assert state == {"epoch": 2, "yielded": consumed}
+
+    # Resume, then query state BEFORE iterating: must still reflect the restored progress.
+    resumed = _MapStyleSamplerWrapper(dataset, **kwargs)
+    resumed.load_state_dict(state)
+    assert resumed.state_dict() == state  # used to return yielded == 0
+
+    # And the resumed iteration actually skips the already-consumed prefix.
+    full = _MapStyleSamplerWrapper(dataset, **kwargs)
+    full.set_epoch(2)
+    full.output_index_for_resume = True
+    resumed.output_index_for_resume = True
+    full_indices = [idx for _, idx in full]
+    resumed_indices = [idx for _, idx in resumed]
+    assert resumed_indices == full_indices[consumed:]
--- a/tests/data/test_datasets.py
+++ b/tests/data/test_datasets.py
@@ -4,9 +4,7 @@ import random
 import subprocess
 import sys
 from functools import partial
-from types import SimpleNamespace
 from typing import Any, Dict, List
-from unittest.mock import patch
 
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
@@ -16,7 +14,6 @@ import torch
 import yaml
 from tools import resolve_ops_overrides
 from torch.utils.data import DistributedSampler
-from torchdata.stateful_dataloader import StatefulDataLoader
 from transformers import PretrainedConfig
 from utils import (
     DummyDataset,
@@ -30,6 +27,7 @@ from utils import (
 
 from veomni.arguments import parse_args
 from veomni.data.data_collator import MainCollator
+from veomni.data.data_loader import DistributedDataloader
 from veomni.data.dataset import DynamicBatchingSizeDataset, _MapStyleSamplerWrapper
 from veomni.distributed.parallel_state import get_parallel_state
 from veomni.trainer.base import BaseTrainer, VeOmniArguments
@@ -337,11 +335,6 @@ def test_native_dataset(dataset_type: str, dyn_bsz: bool, dummy_native_dataset_c
     assert result.returncode == 0
 
 
-# ---------------------------------------------------------------------------
-# _MapStyleSamplerWrapper unit tests (CPU-only, no torch.distributed init:
-# DistributedSampler is constructed with explicit num_replicas / rank).
-# ---------------------------------------------------------------------------
-
 _WRAPPER_DATASET_SIZE = 50
 _WRAPPER_MICRO_BATCH_SEQ_LENGTH = 64
 _WRAPPER_READY_THRESHOLD = 4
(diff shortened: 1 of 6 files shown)


Luosuu:
@LiuzcEECS maybe we need to address merge conflict first

veomni/data/dataset.py
```diff
@@ -721,6 +722,144 @@ def load_state_dict(
...
+                yield sample, int(idx)
+            else:
+                yield sample
+
+    def get_item(self, idx: int):
+        """Refetch a sample by global dataset index (used by ``save_by_idx`` resume)."""
+        return self.dataset[idx]
+
+    def state_dict(self) -> Dict[str, Any]:
+        return {"epoch": self.epoch, "yielded": self._yielded}
```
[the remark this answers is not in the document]
FoolPlayer: Gemini looks right,can you check that in resume case?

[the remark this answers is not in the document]
LiuzcEECS: fixed

FoolPlayer (commented):
Three P2 observations from a local review — all non-blocking, feel free to defer.

veomni/data/dataset.py
```diff
@@ -812,6 +813,149 @@ def load_state_dict(
...
+        num_workers = worker_info.num_workers if worker_info is not None else 1
+        # Strided worker split keeps per-worker counts balanced (difference <= 1).
+        indices = self._rank_indices()[worker_id::num_workers]
+
+        if not self._just_resumed:
+            self._yielded = 0
+        else:
+            # Resuming: keep the restored progress (clamped to this worker's index count).
+            self._yielded = min(self._yielded, len(indices))
+            self._just_resumed = False
```
FoolPlayer, line 919: P2 (robustness): the `min(_yielded, len(indices))` clamp silently masks a `num_workers` change on resume — StatefulDataLoader will already shuffle per-worker snapshots to different workers, and this clamp lets training continue on misaligned shards without any warning. The docstring says the same `num_workers` must be used, but nothing enforces it.

Suggestion: store `num_workers` in `state_dict()` and raise (or at least `logger.warning_rank0`) in `__iter__` / `load_state_dict` when it changes; or turn the clamp into an explicit `raise RuntimeError` when `_yielded > len(indices)`.

veomni/data/dataset.py
```diff
@@ -812,6 +813,149 @@ def load_state_dict(
...
+        self.dataset = dataset
+        self.num_replicas = num_replicas
+        self.rank = rank
+        self.shuffle = shuffle
+        self.seed = seed
+        self.drop_last = drop_last
+        self.epoch = 0
+        # Set by DynamicBatchingSizeDataset.save_by_idx; when True, yield (sample, global_idx).
+        self.output_index_for_resume = False
+
```
FoolPlayer, line 881: Minor (robustness): `num_replicas == 0` isn't rejected here — `rank < 0 or rank >= num_replicas` passes for `rank=0, num_replicas=0`, and the next line hits `len(dataset) % 0` → `ZeroDivisionError`. Real callers pass `parallel_state.dp_size` so this is purely defensive; consider `if num_replicas <= 0: raise ValueError(...)` before the rank check for a clearer error.

veomni/data/data_loader.py
```diff
@@ -192,6 +197,17 @@ def build_native_dataloader(
...
         else:
+            if not isinstance(dataset, IterableDataset):
+                # Map-style datasets lose their DistributedSampler once wrapped into the
+                # (iterable) DynamicBatchingSizeDataset. Adapt them to a per-rank,
+                # per-worker iterable with sampler-equivalent index assignment.
+                dataset = _MapStyleSamplerWrapper(
+                    dataset,
+                    num_replicas=parallel_state.dp_size,
+                    rank=parallel_state.dp_rank,
+                    shuffle=shuffle,
```
FoolPlayer, line 208: P2 (test coverage): this is the only place the new wrapper is wired into the production path (dp_size/dp_rank/shuffle/seed injection), but no end-to-end test exercises it — existing `test_native_dataset[True-mapping]` / `test_multisource_dataset[True-mapping]` run with the default `dyn_bsz_runtime="main"`, and the new `test_map_style_sampler_wrapper_*` tests only cover the wrapper in isolation.

Suggestion: add a `dyn_bsz_runtime="worker"` parametrization (ideally also `dyn_bsz_dataset_save_by_idx=True`) to the existing torchrun-based mapping-dataset tests to catch regressions on the actual `build_native_dataloader → wrapper → DynamicBatchingSizeDataset → DistributedDataloader` chain, including multi-epoch + resume.

veomni/data/dataset.py
```diff
@@ -812,6 +813,153 @@ def load_state_dict(
...
+            if self.output_index_for_resume:
+                yield sample, int(idx)
+            else:
+                yield sample
+
+    def get_item(self, idx: int):
+        """Refetch a sample by global dataset index (used by ``save_by_idx`` resume)."""
+        return self.dataset[idx]
+
+    def state_dict(self) -> Dict[str, Any]:
```
Coach257, line 956: P2: The overrun check only rejects some `num_workers` changes on resume. If the restored `yielded` still fits the new worker’s shorter shard, iteration continues even though the strided index assignment has changed, which can skip or replay samples. Since exact resume requires the same worker count, please persist `num_workers` in `state_dict()` and raise on any mismatch.

Coach257: `StatefulDataLoader` has checked the restored `num_workers`.

Coach257 (approved):

veomni/data/dataset.py
```diff
@@ -812,6 +813,153 @@ def load_state_dict(
...
+        return {"epoch": self.epoch, "yielded": self._yielded}
+
+    def load_state_dict(self, state_dict: Dict[str, Any]) -> None:
+        self.epoch = state_dict["epoch"]
+        self._yielded = state_dict["yielded"]
+        self._just_resumed = True
+
+    def set_epoch(self, epoch: int) -> None:
```
FoolPlayer, line 985: **P1**: `set_epoch()` doesn't clear the resume latch, so an entire epoch can be silently skipped.

`load_state_dict()` arms `_just_resumed=True` and stores the restore point in the same field used for live progress (`_yielded`). Because `__iter__` is a generator, the latch is only disarmed on the first `next()`. With `num_workers=0`, `CheckpointerCallback` calls `iter(train_dataloader)` when `start_step == 0` (`checkpoint_callback.py:104-106`) — that applies the checkpoint state to the long-lived wrapper but never advances the generator. `set_epoch(E+1)` then changes the permutation while `_yielded` still holds the previous epoch's final count, so `indices[self._yielded:]` is empty and the whole epoch yields nothing.

`StatefulDistributedSampler` avoids exactly this by keeping the restore point in a separate `next_yielded` and resetting `yielded = 0` unconditionally at the top of `__iter__`. Minimal fix:

```python
def set_epoch(self, epoch: int) -> None:
    self.epoch = epoch
    self._yielded = 0
    self._just_resumed = False
```

This stays correct for mid-epoch resume: with `start_step != 0` the callback doesn't call `iter()`, so the order is `set_epoch` -> `iter()` -> state applied.

veomni/data/dataset.py
```diff
@@ -812,6 +813,153 @@ def load_state_dict(
...
+    def __len__(self) -> int:
+        """Number of samples assigned to this rank (across all of its workers)."""
+        return self.num_samples
+
+    def _rank_indices(self) -> list:
+        """Compute this rank's global indices, mirroring ``DistributedSampler.__iter__``."""
+        if self.shuffle:
+            g = torch.Generator()
+            g.manual_seed(self.seed + self.epoch)
+            indices = torch.randperm(len(self.dataset), generator=g).tolist()
```
FoolPlayer: **P2**: `.tolist()` materializes the full global permutation in every worker, every epoch, before slicing down to that worker's share. `DistributedSampler` pays this once in the main process. Measured on a 5M-sample dataset: ~2.9s and ~200MB of Python heap per call (the int64 tensor itself is only 40MB — the `.tolist()` is the cost). With 8 workers that's ~1.6GB per rank and ~23s of added latency at every epoch boundary, all recomputing an identical result.

Keeping it a tensor through both slices and converting only the worker's share fixes it:

```python
perm = torch.randperm(len(self.dataset), generator=g)
# ... pad / truncate with torch.cat / slicing ...
indices = perm[self.rank : self.total_size : self.num_replicas][worker_id::num_workers].tolist()
```

veomni/data/dataset.py
FoolPlayer, line 956: **P2**: the resume state carries no topology fingerprint. The only restore-time validation is the `_yielded > len(indices)` check in `__iter__`. If `num_replicas`, `rank`, `seed`, or `len(dataset)` changes between save and resume, `_rank_indices()` produces a different permutation and the skip-prefix is meaningless — the run silently trains on the wrong samples. Recording those four values here and raising on mismatch in `load_state_dict()` would make it fail loudly instead.

FoolPlayer (commented):
Review notes from a local pass over this diff. Verdict: **needs-attention**.

The sharding math is solid — I extracted `_MapStyleSamplerWrapper` into a standalone module and verified `_rank_indices()` is bit-identical to `torch.utils.data.DistributedSampler` across dataset size {1, 7, 50} x num_replicas {1, 2, 4} x shuffle x drop_last x epoch, including the empty-dataset and `num_replicas > len(dataset)` cases. Since map-style + `dyn_bsz_runtime="worker"` previously raised `TypeError` outright, there's no silent-regression risk either.

The findings below are about resume-state lifecycle and cost. One P1, the rest P2.

veomni/data/data_loader.py
```diff
@@ -192,6 +197,17 @@ def build_native_dataloader(
 
             collate_fn = UnpackDataCollator()
         else:
+            if not isinstance(dataset, IterableDataset):
+                # Map-style datasets lose their DistributedSampler once wrapped into the
+                # (iterable) DynamicBatchingSizeDataset. Adapt them to a per-rank,
+                # per-worker iterable with sampler-equivalent index assignment.
+                dataset = _MapStyleSamplerWrapper(
```
FoolPlayer, line 204: **P2**: `drop_last` isn't forwarded, so production always gets `drop_last=False` and the tail is padded by wrapping indices around (duplicating samples) even when `args.data.dataloader.drop_last=True`. This matches the `StatefulDistributedSampler` construction below, so it isn't a regression — but it does mean the wrapper's `drop_last` parameter is unreachable from any caller. Either pass `drop_last=drop_last` or drop the parameter.

tests/data/test_dynamic_batching_dataset.py
```diff
@@ -532,22 +533,24 @@ def test_save_load_state_dict(save_by_idx):
...
-        (False, True, False),
-        (True, False, False),
-        (True, True, False),
-        (True, True, True),
+        ("iterable", False, False, False),
+        ("iterable", False, True, False),
+        ("iterable", True, False, False),
+        ("iterable", True, True, False),
+        ("iterable", True, True, True),
+        ("mapping", True, True, False),
```
FoolPlayer, line 543: **P2**: only one map-style combination is covered. Two gaps worth closing:

- `multi_sample_per_iteration=True` never runs with `mapping`, which is exactly the path the wrapper's docstring warns about (one index expanding to `list[dict]`, resumed via `get_item(idx)` plus `sample_idx` re-selection). It's the most fragile interaction this PR introduces.
- `save_by_idx=False` never runs with `mapping`, so the full-sample-buffer resume path is untested for map-style.

Adding `("mapping", True, True, True)` and `("mapping", True, False, False)` would cover both. Note also that the epoch-boundary bug flagged on `dataset.py` can't reproduce here, since this test runs with `num_workers=2`.

tests/data/test_datasets.py
```diff
@@ -320,3 +335,145 @@ def test_native_dataset(dataset_type: str, dyn_bsz: bool, dummy_native_dataset_c
     command = build_command(dataset_type, dyn_bsz, data_path=data_path)
     result = subprocess.run(command, check=True)
     assert result.returncode == 0
+
+
+# ---------------------------------------------------------------------------
```
FoolPlayer: Nit: these banner dividers are forbidden by the repo's `no-section-divider-comments` rule. The explanation would fit naturally in the docstring of the first test.

veomni/data/dataset.py
LiuzcEECS: I am thinking if we can do the same way as DistributedSampler

veomni/data/dataset.py
LiuzcEECS: num_replicas, seed and len(dataset) should be kept. About rank, I am wandering in dcp save/load, how is the rank aligned?

veomni/data/dataset.py
FoolPlayer: You can keep `DistributedSampler`'s index math exactly as-is — the problem isn't the algorithm, it's *where* it runs.

For a map-style dataset, `DistributedSampler.__iter__` is consumed in the **main process**: the DataLoader iterates `_index_sampler` there and ships index lists to the workers over the index queues. So `randperm(N).tolist()` is paid once per rank per epoch.

Here the wrapper is an `IterableDataset`, so `__iter__` runs **inside each worker**, and `_rank_indices()` recomputes the identical global permutation in every worker. That's where the `num_workers` multiplier comes from — so "the same way as DistributedSampler" would still leave the duplication in place.

The cheap fix keeps the math bit-identical and only changes when the tensor becomes a Python list:

```python
if self.shuffle:
    g = torch.Generator()
    g.manual_seed(self.seed + self.epoch)
    indices = torch.randperm(len(self.dataset), generator=g)
else:
    indices = torch.arange(len(self.dataset))
# pad / truncate with torch.cat + slicing instead of list ops
indices = indices[self.rank : self.total_size : self.num_replicas]
```

then in `__iter__`: `indices[worker_id::num_workers].tolist()`. Same output, but each worker only materializes its own share instead of all N.

Caching `_rank_indices()` wouldn't help, for what it's worth — `persistent_workers` is off, so `__iter__` already runs exactly once per worker per epoch.

veomni/data/dataset.py
FoolPlayer: `rank` is worth keeping too, and the reason is exactly the save/load asymmetry you're asking about.

The dataloader state travels in `extra_state`, and `DCPCheckpointer` keys those files by **global** rank:

```python
extra_state_path = os.path.join(extra_state_dir, _EXTRA_STATE_FORMAT.format(dist.get_rank()))
```

(`veomni/checkpoint/dcp_checkpointer.py:661` on save, `:676` on load — i.e. `extra_state_rank_{global_rank}.pt`.) So global rank *i* always reloads what global rank *i* wrote; the alignment is purely positional.

But the wrapper shards by `parallel_state.dp_rank`, and the global-rank -> dp_rank mapping depends on the mesh layout. `init_parallel_state` builds dims in a fixed order — `pp`, `dp_replicate`, `dp_shard`, `ulysses`, `cp`, `tp` — so `pp` sits *before* the DP dims while `ulysses`/`cp`/`tp` sit *after*. Move parallelism across that boundary while keeping `dp_size` fixed (e.g. 16 ranks going from `pp=2, dp_shard=8` to `dp_shard=8, ulysses=2`: `dp_size` is 8 either way, but `dp_rank` goes from `g % 8` to `g // 2`) and global rank *i* now loads a state that belonged to a different `dp_rank`. `num_replicas` alone won't catch that; `rank` will.

So I'd store all four: `num_replicas`, `rank`, `seed`, `len(dataset)`. Cheap, and it turns a silent wrong-data run into a clear error.

veomni/data/dataset.py
LiuzcEECS: it is right and we might also have some similar problem in `veomni/data/dataset.py`, let me try to fix them together

veomni/data/dataset.py
LiuzcEECS: The extra state is saved and restored from a global-rank-specific file. With VeOmni’s currently supported topology (pp_size == 1) and the fixed mesh-dimension ordering, the flattened DP rank is determined by the global rank and DP size. Therefore, once num_replicas is validated and the world size is unchanged, validating rank separately would be redundant. I’ll persist and validate num_replicas, seed, dataset length, shuffle, and drop_last.

FoolPlayer: Agreed, you're right — and I shouldn't have leaned on the `pp` case: `arguments_types.py:802` asserts `pp_size == 1` ("Pipeline parallel size not supported yet"), so it isn't reachable today.

With `pp == 1` the mesh is `(dp_replicate, dp_shard, ulysses, cp, tp)` and every non-DP dim sits *after* the DP dims, so `dp_rank == g // (world_size / dp_size)`. Validating `num_replicas` with an unchanged world size does pin it down. `shuffle` and `drop_last` are good additions I'd missed — both change the permutation or the truncation.

Two optional notes:

- The residual gap is a *shrinking* world size that preserves `dp_size` — e.g. 16 ranks at `sp=4` -> 8 ranks at `sp=2`, `dp_size=4` throughout. The surviving ranks find their `extra_state_rank_{g}.pt` files and `num_replicas` matches, but `dp_rank` shifts from `g // 4` to `g // 2`. Growing world size fails loudly instead (missing files). Constraint 16 already says changing rank count requires resharding, so this is arguably out of scope — storing `world_size` would close it for one int if you want the insurance.
- If `pp` support ever lands, `rank` becomes load-bearing again, since `pp` sits *before* the DP dims in the mesh order. Probably worth a short comment next to the fingerprint so the assumption is written down rather than implicit.

veomni/data/data_loader.py
LiuzcEECS: let me drop it

veomni/data/dataset.py
FoolPlayer: Verified the fix — and I owe you a correction: your conditional is right, and my suggested *unconditional* reset would have been wrong.

I claimed the ordering was `set_epoch` -> `iter()` -> state applied. It's the reverse. `BaseTrainer.train()` calls `self.on_train_begin()` first (`base.py:849`), which is where `CheckpointerCallback` runs `train_dataloader.load_state_dict(...)`, and only then enters the epoch loop that calls `set_epoch()` (`base.py:858-860`). So an unconditional reset would have zeroed `_yielded` and cleared `_just_resumed` immediately after the restore, and a mid-epoch resume would have replayed the epoch from the start.

The `epoch != self.epoch` guard separates the two cases exactly. With `start_epoch = global_step // train_steps` and `start_step = global_step % train_steps` (`checkpoint_callback.py:85-86`):

- saved at an epoch boundary -> `global_step == (E+1) * train_steps` -> `start_epoch == E+1` while the restored `self.epoch == E`, so the guard fires and the new epoch starts clean;
- saved mid-epoch -> `start_epoch == E == self.epoch`, the guard doesn't fire, and the restore point survives.

Correct in both directions. Thanks for catching my mistake by construction.

FoolPlayer:
Re-reviewed at `cde890fa`. Most of the previous round is addressed — thanks. Leading with the CI failure since it's what's blocking.

## CI is red on a pre-existing bug your new test just found

`gpu_unit_tests_v5` and `VeOmni Ascend test (self-host)` fail on the same test with the same error, so it's neither NPU-specific nor flaky:

```
tests/data/test_datasets.py:430: in test_map_style_sampler_wrapper_resume
>               resaved_state = resaved_dataloader.state_dict()
torchdata/stateful_dataloader/worker.py:57: in try_to_deserialize
    obj.load_state_dict(state_dict)
state_dict = {'buffer_physical_token_count': 106, 'buffer_token_count': 106, 'dynamic_batch_upstream_dataset_state': {...}, 'save_by_idx': False}
>               for item in state_dict["buffer"]
E           KeyError: 'buffer'
veomni/data/dataset.py:1359: KeyError
```

The cause is `DynamicBatchingSizeDataset.load_state_dict`, which ends with `del state_dict["buffer"]` (`dataset.py:1379`) — it destroys the caller's dict. That line is pre-existing (`main` has it at :1207), but nothing had ever reused a state object before. The new "resave after load" block hands the same `dataloader_state` to a second `load_state_dict`, and by then `buffer` is gone.

I'd fix it in `dataset.py` rather than deep-copying in the test. `load_state_dict` mutating its input is surprising on its own, and `torchdata`'s `try_to_deserialize` can invoke it on a shared snapshot, so the test isn't the only caller that can trip over it. Dropping the `del` (or operating on a shallow copy) should be enough — it reads like a memory optimization, and the caller discards the dict anyway.

Good find, incidentally: the rewritten test is doing exactly what it should.

## Addressed

- `set_epoch()` resume latch (P1) — verified, details in that thread. Your conditional is better than what I suggested.
- `_rank_indices()` stays a tensor through the rank slice; `.tolist()` now only materializes this worker's share.
- Sampler fingerprint on `state_dict()` with validation on restore, plus a dedicated rejection test.
- `drop_last` dropped rather than left half-wired.
- Section dividers gone; the distributed matrix gained `("mapping", True, True, True)` and `("mapping", True, False, False)`; the `pytest.skip` in `test_dataloader.py` is removed so that path is covered too.

## Two coverage points, non-blocking

**1. `test_map_style_sampler_wrapper_worker_split` was deleted.** It was the only direct check that a rank's workers receive disjoint, balanced-within-1, strided shares that together cover exactly `_rank_indices()`. The end-to-end resume test can't stand in for it: golden and resumed both run with the same `num_workers`, so a split that is wrong but *consistently* wrong — overlapping shares, which would duplicate samples — still passes. Restoring it needs little more than a `.tolist()` on the now-tensor `_rank_indices()`.

**2. `sample_length_range=(6, 7)` now applies to the `iterable` cases as well.** Those previously ran with lengths `1..50` against a 32-token budget, which exercised the greedy first-fit skip branch in `_get_micro_batch` (while, admittedly, silently discarding everything longer than 32). With uniform 6/7 lengths every micro batch packs identically. Worth keeping at least one `iterable` parametrization on the original variable-length generator so the packing diversity doesn't disappear from the suite.

FoolPlayer:
Re-reviewed at `0792e21d`. CI is green (18/18) and everything blocking from the earlier rounds is resolved. LGTM.

Removing `del state_dict["buffer"]` fixes the failure directly. More interesting is that you dropped the `epoch != self.epoch` guard I signed off on last round and replaced it with something better: splitting `__iter__` into an eager, non-generator function that handles the resume latch, with `_iter()` holding the generator body. That is the real mechanism, and it makes correctness independent of epoch bookkeeping. I traced all three paths:

- **`num_workers=0`** — `load_state_dict` mutates the live object, so the latch is real state. `CheckpointerCallback` calls `iter()` on an epoch-boundary resume (`checkpoint_callback.py:104-106`); because `__iter__` is now eager, that call actually consumes the latch instead of deferring it to a first `next()` that never comes. The epoch-3 `iter()` then sees `_just_resumed == False` and zeroes `_yielded`.
- **`num_workers>0`** — `StatefulDataLoader` never applies the snapshot to the main-process dataset; it hands it to workers through `try_to_deserialize`. So the main-process wrapper keeps `_just_resumed == False` / `_yielded == 0` throughout, and the second `iter()` forks clean workers.
- **Across the two-layer nesting** — `DynamicBatchingSizeDataset._iter()` calls `iter(self.dataset)` lazily, so the outer `iter()` does not reach the wrapper's latch. The new `if hasattr(self.dataset, "_just_resumed"): self.dataset._just_resumed = False` in the non-resuming branch is what closes that gap: it pushes the "not resuming" decision down a level, and the wrapper zeroes `_yielded` itself once its own `__iter__` runs.

So three flags across two classes have to agree, and the whole thing rests on `__iter__` not being a generator. Your five cases do lock that contract down — adding a `yield` back into either `__iter__` would fail Case 3 — but the constraint is invisible at the call site. A one-line comment on each split, saying the body lives in `_iter()` so that latch handling runs eagerly at `iter()` time, would save the next reader from rediscovering this.

Worth calling out in the PR description that the eager split also fixes the same latent bug on the pre-existing **iterable** path, not just the new map-style one: the callback's epoch-boundary `iter()` previously did nothing at all there either.

The two coverage points from last round are still open — `test_map_style_sampler_wrapper_worker_split` is deleted, and `sample_length_range=(6, 7)` still applies to the `iterable` cases (`test_dynamic_batching_dataset.py:666,674`). Neither blocks, both are your call. I would take the worker-split test back at some point, since nothing else checks that a rank's workers get non-overlapping shares.

PR #857, opened 2026-06-22
[ops] feat: stream all2all chunks to reduce FusedMoe with EP memory

Luosuu:
## Summary
- stream the fused_triton EP merged-fc1 post-all2all path by hidden chunks to avoid materializing the full routed expert output
- add a fixed-meta fused grouped GEMM helper for merged gate/up weights
- keep split EP post-all2all weighting dense before unpermute and update shared routing-map callers

## Tests
- `ruff format` on changed files
- `python -m py_compile` on changed files
- focused EP merged-fc1 post-all2all correctness check
- 2-step fused_triton EP training smoke test

### DIFF
--- a/veomni/distributed/moe/moe_layer.py
+++ b/veomni/distributed/moe/moe_layer.py
@@ -283,15 +283,16 @@ def _weighted_unpermute_forward(
     output: torch.Tensor,
 ):
     hidden_dim = output.size(-1)
-    output.zero_()
+    fp32_output = torch.zeros(output.shape, device=output.device, dtype=torch.float32)
     for start in range(0, tokens.size(0), _UNPERMUTE_CHUNK_SIZE):
         end = min(start + _UNPERMUTE_CHUNK_SIZE, tokens.size(0))
         chunk_mapping = permutation_mapping[start:end]
-        output.scatter_add_(
+        fp32_output.scatter_add_(
             0,
             chunk_mapping.unsqueeze(1).expand(-1, hidden_dim),
-            (tokens[start:end] * tokens_weight[start:end].unsqueeze(-1)).to(output.dtype),
+            (tokens[start:end] * tokens_weight[start:end].unsqueeze(-1)).float(),
         )
+    output.copy_(fp32_output.to(output.dtype))
     return output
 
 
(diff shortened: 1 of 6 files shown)


Luosuu:
Additional verification update:

- Confirmed the existing same_nk/same_mn grouped GEMM tuned dispatch remains unchanged.
- Confirmed the new merged gate/up helper uses fixed launch meta and no pretuned fallback path.
- Ran a two-step fused_triton EP MoE training smoke; both backward steps completed.

veomni/distributed/moe/moe_layer.py
```diff
@@ -112,29 +235,389 @@ def tokens_post_all2all(
...
+    permutation_mapping: torch.Tensor,
+    output: torch.Tensor,
+):
+    hidden_dim = output.size(-1)
+    output.zero_()
+    for start in range(0, tokens.size(0), _UNPERMUTE_CHUNK_SIZE):
+        end = min(start + _UNPERMUTE_CHUNK_SIZE, tokens.size(0))
+        chunk_mapping = permutation_mapping[start:end]
+        output.scatter_add_(
+            0,
```
Luosuu, line 291: This new merged EP path accumulates scatter results in the output dtype, while the existing `unpermute` path uses FP32 accumulation before casting back. For top-k > 1 this can regress numerical parity in BF16/FP16; consider accumulating this chunk in FP32 and casting once into `output`.

veomni/distributed/moe/moe_layer.py
```diff
@@ -112,29 +235,389 @@ def tokens_post_all2all(
...
+                local_output = _all_to_all_no_autograd(
+                    ctx.ep_group,
+                    sorted_output,
+                    ctx.input_splits,
+                    ctx.output_splits,
+                )
+                grad_tokens_weight.add_((grad_local * local_output).sum(dim=-1).to(tokens_weight.dtype))
+
+        grad_permute_tokens = permute_tokens
+        for token_start, token_end in _iter_token_chunks(permute_tokens.shape[0]):
```
Luosuu, line 548: `grad_permute_tokens` aliases the saved forward input here, and the loop below overwrites it with `copy_()`. That can corrupt retained/repeated backward or any path that still needs the saved tensor. Use `torch.empty_like(permute_tokens)` instead.

Luosuu (commented):
Two small correctness comments from review.

PR #864, opened 2026-06-25
[data] feat: support infinity / infinity_padding in the worker dynamic-batching path

LiuzcEECS:
### What does this PR do?

This PR adds two new options, `infinity` and `infinity_padding`, for worker-side dynamic batching. They allow a `dyn_bsz_runtime="worker"` dataloader to handle upstream dataset exhaustion without immediately stopping.

- `infinity`: restart the upstream iterator when the source is exhausted and continue producing micro batches.
- `infinity_padding`: after the source is exhausted and the buffer is drained, emit padding micro batches deep-copied from the last real micro batch and marked with `padding_flag=True`.

Padding batches are excluded by `count_loss_token()`. While at least one rank still has real data, padding ranks therefore contribute zero token-weighted gradients. When all ranks reach padding, `mean_global_loss()` raises `StopIteration` to end training.

Both options default to `False`; existing behavior remains unchanged when they are disabled.

### Key Modifications

- **`veomni/data/data_loader.py`**
  - Add `infinity` and `infinity_padding` parameters to `build_native_dataloader()`.
  - Forward them to `DynamicBatchingSizeDataset` only when `dyn_bsz_runtime="worker"`.
  - Leave the main-process dynamic batching path unchanged.

- **`veomni/data/dataset.py`**
  - Store the two options in `DynamicBatchingSizeDataset`.
  - Restart the upstream iterator after exhaustion when `infinity=True`.
  - Drain the remaining buffer and then emit deep-copied padding micro batches when `infinity_padding=True`.

- **`veomni/utils/loss_utils.py`**
  - Detect when all ranks have zero loss tokens under infinity-padding mode and raise `StopIteration` to end training.

- **`veomni/trainer/base.py`**
  - Enable the all-ranks-padding check only for worker-side infinity-padding mode.
  - Avoid accessing the underlying dataset for the main-process dynamic batching runtime.

## Summary by CodeRabbit

* **New Features**
  * Added infinite iteration for worker-side dynamic batching.
  * Added optional zero-loss padding batches after data exhaustion.
  * Added checkpoint and resume support for both modes.
* **Bug Fixes**
  * Improved iterator cleanup and worker shutdown at the end of training.
  * Excluded padding batches from loss and token-count calculations.
  * Added safeguards for incompatible or mutually exclusive data-loading options.
* **Documentation**
  * Updated configuration references and usage examples with the new options and limitations.
* **Tests**
  * Expanded distributed training coverage for cycling, padding, exhaustion, and resume behavior.


veomni/data/batch_metadata.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
FoolPlayer: This file is unused anywhere on the branch and unrelated to the infinity feature (it landed in the `fix comments` commit). Suggest dropping it from this PR, or moving it to the collator PR that actually consumes it.

veomni/data/dataset.py
```diff
@@ -1021,13 +1043,45 @@ def __iter__(self):
...
+                                "infinity_padding is enabled but no micro batch was ever produced, "
+                                "there is nothing to pad from; stopping iteration."
+                            )
+                            return
+                        # Keep yielding padding micro batches forever; downstream groups them
+                        # into num_micro_batch-sized steps and loss accounting skips them via
+                        # the padding_flag marker. One deep copy suffices: downstream is
+                        # read-only on the yielded batch (preforward rebuilds a new dict and
+                        # moves tensors to device before the model sees them).
+                        padding_batch = copy.deepcopy(last_micro_batch)
```
FoolPlayer: Padding batches keep the real `labels`, so the model still computes a non-zero loss while `count_loss_token` counts 0 tokens. If an entire global step is all-padding (e.g. `train_steps` overruns the data, or all DP ranks exhaust in the same step), `mean_global_loss` raises `ValueError` (`all_reduced_len==0` but `loss!=0`). Safe only in the ragged case where >=1 rank still has real data. Consider emitting a genuinely zero-loss batch (labels all `IGNORE_INDEX`) or guaranteeing >=1 real micro-batch per step.

veomni/data/dataset.py
```diff
@@ -1021,13 +1043,45 @@ def __iter__(self):
...
                 if isinstance(e, StopIteration):
+                    if self._infinity:
+                        if just_restarted:
+                            raise RuntimeError(
+                                "infinity=True but the upstream dataset produced nothing after a restart; "
+                                "refusing to spin forever on an empty dataset."
+                            ) from None
+                        # Restart the upstream dataset and keep going forever; the buffer is
+                        # preserved so batching continues to accumulate across epochs.
+                        self._data_iter = iter(self.dataset)
```
FoolPlayer: Restart re-iterates without bumping the epoch, and the trainer calls `set_epoch` only once for an infinite run, so epoch-seeded shufflers (`_MapStyleSamplerWrapper`, `WeightedMultiSourceDataset`) replay the identical order every cycle. Suggest `self.set_epoch(self.epoch + 1)` before re-iter, or document the fixed permutation. Also: no tests cover either new mode yet.

FoolPlayer (commented):
Core logic mirrors the existing `dyn_bsz_runtime="main"` padding path (`dynamic_batching.py`) and the `padding_flag` contract is honored by `count_loss_token`, so the approach is sound. A few things worth addressing before merge (inline).

veomni/data/batch_metadata.py
LiuzcEECS: Yes, it was a temp file used to fix another feature; however, we decided not to support it anymore.

I will remove this file

veomni/data/dataset.py
LiuzcEECS: That is the behavior we want it to follow. I will add it to a doc and give it a test func.

FoolPlayer:
Hi, please fix ci~

veomni/data/data_loader.py
```diff
@@ -84,6 +84,8 @@ def build_native_dataloader(
     prefetch_factor: int = 2,
     shuffle: bool = True,
     seed: int = 0,
+    infinity: bool = False,
```
Coach257, line 190: `infinity` / `infinity_padding` are only kwargs on `build_native_dataloader`, not fields on the dataloader/training args, so the normal `BaseTrainer` yaml path cannot turn them on. If this is meant to be user-facing, please wire them into `arguments_types` + docs.

veomni/utils/loss_utils.py
```diff
@@ -73,18 +82,23 @@ def mean_global_loss(
...
+        elif zero_len_nonzero_loss is None and not torch.allclose(cur_loss, torch.zeros_like(cur_loss)):
+            zero_len_nonzero_loss = (loss_name, cur_loss)
 
         if get_parallel_state().sp_enabled:
             cur_loss = cur_loss / get_parallel_state().sp_size
 
         loss_dict[key] = cur_loss
 
+    if infinity_padding and all_reduced_lens and all(all_reduced_len == 0 for all_reduced_len in all_reduced_lens):
+        raise StopIteration("All ranks reached infinity padding.")
```
FoolPlayer: **Blocking**: raising here aborts `train_step` mid-step — `postforward` runs after `self.model(**micro_batch)` and before `loss.backward()`. When this fires at `micro_step == 0`, `BaseTrainer.train_step` has already executed:

- `on_step_begin()`, with no matching `on_step_end()` ever running (`ChannelLossCallback` `begin_step` is only torn down in `on_train_end`).
- `model_reshard(0, n)` -> `set_reshard_after_backward(False)` and `_configure_hsdp_allreduce(0, n)` -> `set_requires_all_reduce(False)`. Both are only restored at `micro_step == n-1`, which never happens; with `num_train_epochs > 1` they stay latched into the next epoch.
- `veomni_clip_grad_norm` and `optimizer.zero_grad()` are skipped, leaving gradients accumulated by earlier micro-steps on the params.
- The traceback pins `outputs` and `loss`, keeping the whole autograd graph alive through `on_epoch_end` -> collective DCP save, i.e. peak memory right at the checkpoint.

Suggest deciding at a clean step boundary instead: all-reduce the total right after `count_loss_token(micro_batches)` in `train_step` and raise there. That's one extra all-reduce per step instead of per micro-batch, avoids a wasted forward, and puts the control flow in the trainer rather than inside a leaf math utility. A dedicated exception type would also be safer than `StopIteration`, which is protocol-overloaded and can be absorbed silently by any intervening generator or `except StopIteration`.

veomni/trainer/text_dpo_trainer.py
```diff
@@ -159,6 +159,12 @@ def __init__(self, args: VeOmniDPOArguments):
             self.base._build_dataset()
             self.base._build_collate_fn()
             self.base._build_dataloader()
+            if args.train.dyn_bsz and args.train.dyn_bsz_runtime == "worker":
+                dynamic_dataset = self.base.train_dataloader.dataset
+                if getattr(dynamic_dataset, "_infinity_padding", False) and not getattr(
+                    dynamic_dataset, "_infinity", False
+                ):
+                    raise ValueError("infinity padding is not supported by TextDPOTrainer.")
```
FoolPlayer: **Blocking**: constraint 24 names both `TextDPOTrainer` and `DiTTrainer` as composed trainers that must each be updated or explicitly rejected, and `DiTTrainer` gets neither. `DiTTrainer.postforward` (`dit_trainer.py:442-449`) computes its own `loss_dict` and never calls `mean_global_loss`, so under `infinity_padding=True` it would train on deep-copied padding batches at full weight until `train_steps` runs out — silently corrupting the model, with no stop signal. The path is reachable: `DiTTrainer._build_dataloader` forwards `dyn_bsz_runtime` (`dit_trainer.py:391`).

Rather than duplicating the guard, consider moving it into `build_native_dataloader` / `DynamicBatchingSizeDataset` behind an explicit opt-in from the trainer, so a future composed trainer fails loudly instead of silently.

docs/usage/basic_modules.md
```diff
@@ -318,6 +318,16 @@ train_dataloader = build_dataloader(
 )
 ```
 
+When worker-side dynamic batching is configured with `infinity=True`, exhausting
```
FoolPlayer: **P1**: this documents behavior that can't currently be enabled. `infinity` / `infinity_padding` exist nowhere outside `build_native_dataloader` and the test's overridden `_build_dataloader`; `BaseTrainer._build_dataloader` forwards `**asdict(args.data.dataloader)` and that dataclass has no such fields. So there's no production entry point, and both the new `postforward` branch and the DPO guard are dead code. Either plumb the two fields through the dataloader argument dataclass, or hold this doc paragraph until the follow-up lands.

veomni/data/dataset.py
```diff
@@ -972,6 +986,12 @@ def __iter__(self):
         else:
             self._just_resumed = False
 
+        # Keep the most recent real micro batch around so infinity_padding has a template
+        # to deep-copy padding batches from after the source is exhausted.
+        last_micro_batch = None
```
FoolPlayer: **P1**: `infinity_padding` isn't resume-safe. `last_micro_batch` is local to `__iter__` and absent from `state_dict()`. On resume, a worker whose restored buffer is empty and whose upstream is already exhausted hits `last_micro_batch is None`, logs a warning and `return`s — it stops instead of padding. Once all workers do that, the run ends at a different step than the pre-checkpoint run, breaking the byte-for-byte resume guarantee this suite exists to protect.

The new test tolerates exactly this: `on_train_end` skips `assert saw_real_batch` when `is_resume_train`, so a resumed run producing zero real batches passes and `compare_global_batch` just compares padding against padding. Persisting the template (or an `_exhausted` flag plus a synthesizable template) would close it.

veomni/data/dataset.py
```diff
@@ -1021,13 +1043,39 @@ def __iter__(self):
...
                 if isinstance(e, StopIteration):
+                    if self._infinity:
+                        if just_restarted:
+                            raise RuntimeError(
+                                "infinity=True but the upstream dataset produced nothing after a restart; "
+                                "refusing to spin forever on an empty dataset."
+                            ) from None
+                        # Restart the upstream iterator without clearing self._buffer.
+                        self._data_iter = iter(self.dataset)
+                        just_restarted = True
```
FoolPlayer: **P2**: this flag is cleared when `next()` *succeeds*, not when a sample is actually buffered. If every upstream sample is discarded as overlong (the `continue` in the length check above), the guard resets, the source exhausts, restarts, and the worker spins at 100% CPU producing nothing — the DataLoader hangs with no error.

The mirror case misfires too: on resume with an exhausted upstream but a non-empty buffer, the first `next()` raises immediately, the restart sets this to `True`, the buffer drains on the following pass, and the next `StopIteration` raises the "refusing to spin forever" `RuntimeError` even though the buffer legitimately had data. Tracking "samples appended since last restart" would be accurate in both directions.

veomni/data/dataset.py
```diff
@@ -1021,13 +1043,39 @@ def __iter__(self):
...
                             logger.warning("dynamic_batching_collate_fn returned None, skip this micro_batch")
+
+                    if self._infinity_padding:
+                        if last_micro_batch is None:
+                            logger.warning(
+                                "infinity_padding is enabled but no micro batch was ever produced, "
+                                "there is nothing to pad from; stopping iteration."
+                            )
+                            return
+                        padding_batch = copy.deepcopy(last_micro_batch)
```
FoolPlayer: **P2**: the `deepcopy` is hoisted outside the `while True`, so every yield returns the identical dict. Pickle memoizes shared references, so all micro-batches within one step alias the same dict and the same CPU tensors even across the worker IPC boundary (`pin_memory=True` happens to break the aliasing by copying; `pin_memory=False` and `num_workers=0` don't). Nothing mutates in place today, so it's latent rather than live — but the `main` path deep-copies per padding batch (`dynamic_batching.py:340`) and matching it is nearly free.

Related: `last_micro_batch = micro_batch` keeps a reference to a batch already handed to the consumer, so the template itself is exposed to downstream mutation.

veomni/data/dataset.py
```diff
@@ -925,6 +936,9 @@ def __init__(
             raise ValueError("force_generate_long_sequence is not supported yet.")
         self.force_generate_long_sequence = force_generate_long_sequence
 
+        self._infinity = infinity
```
FoolPlayer: **P2**: `infinity=True, infinity_padding=True` silently drops the padding mode, and `base.py` / `text_dpo_trainer.py` each re-derive that precedence with `_infinity_padding and not _infinity` by reading two private attributes across a module boundary. Rejecting the combination in `__init__` and exposing a single public property (e.g. `emits_padding_batches`) would keep the rule in one place and avoid drift.

veomni/utils/loss_utils.py
```diff
@@ -73,18 +82,23 @@ def mean_global_loss(
...
-                )
+        elif zero_len_nonzero_loss is None and not torch.allclose(cur_loss, torch.zeros_like(cur_loss)):
+            zero_len_nonzero_loss = (loss_name, cur_loss)
 
         if get_parallel_state().sp_enabled:
             cur_loss = cur_loss / get_parallel_state().sp_size
 
         loss_dict[key] = cur_loss
 
+    if infinity_padding and all_reduced_lens and all(all_reduced_len == 0 for all_reduced_len in all_reduced_lens):
```
FoolPlayer: **P2**: this condition is evaluated per step from the current step's token counts, with no latch. Because the DataLoader round-robins whole batches across workers, a rank can alternate real and padding batches; if two ranks' phases are offset, exactly one rank has real data at every step and the loop never stops until `train_steps`. Not a crash, but it makes the documented guarantee ("stops once all ranks return only padding batches for the same training step") fragile and wastes compute. Latching a per-rank `exhausted` boolean on the first all-padding step and all-reducing that would be robust to phase offsets.

veomni/utils/loss_utils.py
```diff
@@ -31,7 +31,10 @@ def count_loss_token(batches: Union[list[dict[str, torch.Tensor]], dict[str, tor
     }
 
     def _count(obj):
-        if isinstance(obj, dict) and not obj.get("padding_flag", False):
+        if isinstance(obj, dict):
+            if obj.get("padding_flag", False):
```
FoolPlayer, line 35: Worth calling out in the description: this is a real bug fix, not just a refactor. The old body was `if isinstance(obj, dict) and not obj.get("padding_flag", False)`, so a padding dict matched neither that branch nor the `elif isinstance(obj, (list, tuple))` and fell through to `raise TypeError`. That means the pre-existing `dyn_bsz_runtime="main"` padding path (`dynamic_batching.py:341`) crashed in `train_step` whenever the final batch was padded. A regression test would be good, and this probably deserves its own commit.

Adjacent pre-existing bug in this function (not introduced here, but it now feeds the new all-zero detection): the `{token_name}_tokens` line below uses `=` instead of `+=`, so multi-modal token counts are overwritten rather than accumulated across micro-batches.

tests/data/test_dynamic_batching_dataset.py
```diff
@@ -575,11 +590,19 @@ def build_command(shuffle=True, save_by_idx=True, multi_sample_per_iteration=Fal
...
     Returns:
         list: Command arguments for subprocess.run().
     """
     port = find_free_port()
+    # The larger token estimate makes the exhaustion-mode cases request more
+    # steps than the finite test dataset can provide in one pass.
+    train_size = 3000 if infinity or infinity_padding else 2000
+    # Infinite workers can still be active at interpreter teardown, causing the
+    # pin-memory thread to race with process shutdown and occasionally SIGABRT.
+    pin_memory = not (infinity or infinity_padding)
```
FoolPlayer: **P1**: this comment describes a production crash, not a test artifact. `pin_memory` defaults to `True`, so every real `infinity` / `infinity_padding` run hits the same race. `BaseTrainer.train()` only shuts the dataloader iterator down when `use_background_prefetcher` is set, so with never-ending workers the `StatefulDataLoader` iterator is collected at interpreter exit, after CUDA teardown.

Better to fix it on the production side — drop the `VeOmniIter` / underlying iterator reference at the end of each epoch unconditionally so workers and the pin-memory thread are joined before teardown — and then restore `pin_memory=True` here so the test actually covers the risky path.

tests/data/test_dynamic_batching_dataset.py
```diff
@@ -786,14 +824,30 @@ def on_step_begin(self, state: TrainerState, micro_batches: List[Dict[str, Any]]
...
+            )
+
     def on_train_end(self, state: TrainerState, **kwargs) -> None:
+        if self.trainer.infinity:
+            assert self.trainer.saw_real_batch
+            assert not self.trainer.saw_padding_batch
+        elif self.trainer.infinity_padding:
+            if not self.trainer.is_resume_train:
+                assert self.trainer.saw_real_batch
+            assert self.trainer.saw_padding_batch
```
FoolPlayer: **P1**: the termination path is never exercised. `TrainerTest.train_step` calls `next(data_iterator)` and then the two hooks directly — no forward, no `postforward`, no `mean_global_loss` — so the `infinity_padding` stop signal, the `postforward` dataset introspection, and the DPO rejection are all untested. `CheckCallback.on_epoch_end` asserting `curr_step == train_steps - 1` in fact asserts the opposite of production behavior: that nothing stopped the loop.

A unit test on `mean_global_loss(..., infinity_padding=True)` would cover the important cases cheaply: all-zero -> raises; mixed zero/non-zero -> no raise and zero-weighted loss on the padding rank; zero-length with non-zero loss and `infinity_padding=False` -> still `ValueError`.

Separately, the `infinity=True` case has no assertion that a restart ever occurred — `saw_real_batch and not saw_padding_batch` passes even if the dataset never exhausted, so that parametrization can silently degrade into a plain non-exhaustion run if `DATASET_SIZE` or the token math changes. A restart counter would pin it down.

FoolPlayer (commented):
Review notes from a local pass over this diff. Verdict: **risky** — two blocking findings, both about the termination path rather than the batching logic itself. Today's blast radius is limited because the feature isn't reachable from any production config (see the comment on the docs change), but that's a problem in its own right.

Two changes in here are genuine fixes that deserve to be called out in the description, ideally as separate commits: the `count_loss_token` change fixes a pre-existing crash on the `dyn_bsz_runtime="main"` padding path, and the `tests/data/utils.py` `_just_resumed` change fixes a real flag leak.

For the record, PEP 479 is *not* an issue here — I verified `StopIteration` propagates cleanly out through the nested `@contextmanager` layers (`use_parallel_state`, `micro_step_context`) and is caught by `except StopIteration` in `train()`. The `all_reduce` group is also correct: `group=None` is WORLD, so the stop decision is uniform across DP/TP/SP/EP ranks.

Title nit: `[data]` passes CI, but this also touches `veomni/trainer/`, `veomni/utils/`, and `docs/` — `[data, trainer, docs]` would be more accurate.

veomni/arguments/arguments_types.py
```diff
@@ -1331,6 +1331,24 @@ class DataloaderConfig:
         default=None,
         metadata={"help": "Per-worker torch thread count for dataloader subprocesses."},
     )
+    infinity: bool = field(
```
FoolPlayer, line 1394: **P2** (new): these land on the generic `DataloaderConfig`, but they're only honored inside the `dyn_bsz` + `dyn_bsz_runtime="worker"` branch of `build_native_dataloader`. Setting `data.dataloader.infinity=True` with `dyn_bsz=False` or `dyn_bsz_runtime="main"` is silently ignored. The help text says as much, but nothing enforces it — and a silently ignored "never stop training" flag is an unpleasant way to discover the constraint.

`DataloaderConfig.__post_init__` can't see `train.*`, so the check would have to live in the top-level `__post_init__`, or as a `logger.warning_rank0` in `build_native_dataloader` when either flag is set outside the worker path.

FoolPlayer (commented):
Re-reviewed at `f430ded6` (`add args`). The argument plumbing is resolved, and one of my earlier findings needs a correction in your favour. The blocking-level concern from the previous round is unchanged, since `veomni/data/dataset.py`, `veomni/utils/loss_utils.py`, `veomni/trainer/base.py` and `veomni/trainer/text_dpo_trainer.py` are untouched by this commit.

**Fixed** — `infinity` / `infinity_padding` are now on `DataloaderConfig` and reach `build_native_dataloader` through `BaseTrainer._build_dataloader`'s `**asdict(args.data.dataloader)`, so the feature is actually reachable and the `TextDPOTrainer` guard is no longer dead code. Docs now use the fully-qualified `data.dataloader.*` names and state the precedence rule.

**Correction on my `DiTTrainer` blocking** — it isn't reachable, so I'm downgrading it to P2. `DiTTrainer._build_dataloader` enumerates its dataloader kwargs explicitly instead of forwarding `**asdict(...)`, so the new fields never arrive there. And `BaseRLTrainer` is genuinely covered: it *does* forward `**asdict(...)`, but its `postforward` ends with `return super().postforward(...)`, so `mean_global_loss` runs and the stop signal works. Details in the `text_dpo_trainer.py` thread.

**Still open**, all untouched by this commit:

- **blocking** — the mid-step `StopIteration` abort (`loss_utils.py:98` thread): `on_step_begin` without `on_step_end`, latched `set_reshard_after_backward(False)` / `set_requires_all_reduce(False)`, skipped `optimizer.zero_grad()`, and the autograd graph pinned into the collective DCP save.
- **P1** — the termination path still has no test coverage; `TrainerTest.train_step` never runs `postforward`, so `mean_global_loss(..., infinity_padding=True)` is never executed by any test.
- **P1** — `pin_memory=False` is still in the test, and the teardown race it works around is unaddressed on the production side.
- **P1** — `infinity_padding` isn't resume-safe (`last_micro_batch` absent from `state_dict()`).
- **P2** — hoisted `deepcopy`, the `just_restarted` guard, the missing exhaustion latch, and the precedence rule duplicated across `dataset.py` / `base.py` / `text_dpo_trainer.py`.

docs/usage/basic_modules.md
FoolPlayer: Resolved by `f430ded6`. I traced the path end to end: `DataloaderConfig.infinity` / `.infinity_padding` -> `BaseTrainer._build_dataloader`'s `**asdict(args.data.dataloader)` -> `build_native_dataloader`. The `TextDPOTrainer` guard is live now as a result. Thanks.

veomni/trainer/text_dpo_trainer.py
FoolPlayer: Correction on my side — downgrading this from blocking to **P2**, now that the args are plumbed through and I can see which trainers actually receive them.

`DiTTrainer._build_dataloader` (`dit_trainer.py:376-403`) enumerates its dataloader kwargs explicitly instead of forwarding `**asdict(args.data.dataloader)`, so `infinity` / `infinity_padding` never reach it. Not reachable, so not blocking — sorry for the noise.

`BaseRLTrainer` is fine for the opposite reason: it *does* forward `**asdict(...)`, but its `postforward` ends with `return super().postforward(outputs, micro_batch)` (`base_rl_trainer.py:112`), so `mean_global_loss` runs and the stop signal reaches it.

What's left is that DiT is protected only by how it happens to spell its kwargs. If someone switches it to `**dataloader_kwargs` for consistency later, `infinity_padding` silently becomes reachable and DiT would train on padding batches at full weight with no stop signal, since its `postforward` builds `loss_dict` itself. Either add the same explicit `raise` in `DiTTrainer._build_dataloader`, or — cleaner — move the rejection into `build_native_dataloader` / `DynamicBatchingSizeDataset` behind an opt-in, so any trainer that hasn't opted in fails loudly rather than silently.

veomni/arguments/arguments_types.py
LiuzcEECS: logger.warning sounds good

veomni/utils/loss_utils.py
LiuzcEECS: That is right, let me move it to the trainer

LiuzcEECS: Actually I was planning to put it under the dataloader but not sure if it is doable, let me check

veomni/data/dataset.py
LiuzcEECS: sounds good, let's reset the flag only after something is buffered

veomni/data/dataset.py
LiuzcEECS: it should be fine since since nothing mutates in place today

veomni/utils/loss_utils.py
LiuzcEECS: sounds good, let me mention it in the desc and explicitly let @FoolPlayer know it

tests/data/test_dynamic_batching_dataset.py
LiuzcEECS: let me check if "pin-memory thread are joined before teardown" is something doable

veomni/utils/loss_utils.py
LiuzcEECS: "if two ranks' phases are offset, exactly one rank has real data at every step and the loop never stops until train_steps", it might happen for several steps but will not last for long, it is impossible that a rank can alternate real and padding batches for more times than the num of total workers(which is usually 8)

veomni/data/dataset.py
LiuzcEECS: sounds good, I will save the last_micro_batch and test if indeed we resume the same patch

veomni/utils/loss_utils.py
FoolPlayer: I need to walk back part of my own suggestion here, and I don't think the `num_workers` bound holds either. The mechanism turns out to matter, because it changes what the fix should be.

For `dyn_bsz_runtime="worker"`, `dataloader_batch_size == gradient_accumulation_steps` (`arguments_types.py:875`), so one DataLoader batch is exactly one training step's micro-batch list. And with `num_workers > 0` a DataLoader batch is assembled by a *single* worker, served round-robin. So step *i* comes from worker `i % num_workers`, and a step is all-padding iff that one worker is exhausted.

That makes the mixed phase run from the first worker's exhaustion to the last one's, with its length governed by the *imbalance in micro-batch counts across workers* rather than by `num_workers`. Sample counts are balanced within +/-1, but micro-batch counts are token-driven and can diverge: a worker with 10 lists next to one with 20 gives ~20 steps of alternation at `num_workers=2`, not ~2.

The length isn't the real problem, though — **premature termination is**. Worked example, `num_workers=2`, two ranks:

| | worker 0 | worker 1 |
|---|---|---|
| rank A | 10 lists, real on even steps 0-18 | 20 lists, real on odd steps 1-39 |
| rank B | 12 lists, real on even steps 0-22 | 18 lists, real on odd steps 1-35 |

At step 24 (even, so worker 0 on both ranks) both ranks emit only padding, `all_reduced_len == 0` globally, and training stops — while worker 1 still holds real data on both ranks, through steps 39 and 35 respectively. That data is silently dropped.

This is also why my original suggestion was wrong: latching a per-rank boolean on the *first* all-padding step latches at step 24 as well, so it doesn't help. The condition has to be per-worker rather than per-step — a rank is done only once **all `num_workers` of its slots** have gone to padding. Cheapest version is probably to have the padding batch carry the emitting worker id (or an `exhausted` marker), have the trainer accumulate the set of exhausted slots per rank, and all-reduce `len(exhausted_slots) == num_workers` instead of the raw token count.

A test asserting that the number of real micro batches consumed equals the number produced would catch exactly this.

veomni/utils/loss_utils.py
LiuzcEECS: in train_step, we can see
```
        micro_batches: List[Dict[str, Any]] = next(data_iterator)

        self.on_step_begin(micro_batches=micro_batches)

        # Forward and backward for each micro batch
        synchronize()

        total_loss = 0.0
        total_loss_dict = defaultdict(int)

        # token num for fixed_ce_loss in postforward
        self.micro_batches_token_len = count_loss_token(micro_batches)
```

I think we can put `all_reduced_len = all_reduce((micro_batches_token_len[f"{loss_name}_tokens"].item()), op="sum")` in `mean_global_loss` to somewhere in VeOmniIter(which is `data_iterator`) before self.on_step_begin of the above code block and raise a StopIteration just like how it will do when the `data_iterator` is out

veomni/utils/loss_utils.py
LiuzcEECS: correct, I think we need to move the padding logic into VeOmniIter so it will not show the unalignment at the worker level

veomni/utils/loss_utils.py
FoolPlayer: Agreed, and moving it into `VeOmniIter` is better than just relocating it into the trainer — it solves more than the mid-step abort.

All five trainers build one (`base.py:869`, `text_trainer.py:175`, `vlm_trainer.py:348`, `dit_trainer.py:565`, `text_dpo_trainer.py:461`), so the stop condition becomes uniform instead of living in `mean_global_loss`, which `DiTTrainer` never reaches (its `postforward` takes `outputs.loss` directly). That also removes the reason for the explicit `raise ValueError("infinity padding is not supported by TextDPOTrainer.")` guard, and it means raising happens before `on_step_begin`, so no rank ends up with half a step's gradients or an unbalanced `on_step_begin`/`on_step_end` pair. That was the whole of my blocking concern.

One hazard to design around: the vote is a collective that runs *after* the local `next()` has returned. If any rank's underlying iterator raises `StopIteration` locally before reaching the all-reduce, every other rank blocks in the collective until the watchdog fires. Under `infinity_padding=True` the upstream never raises — it pads forever — so the ordering is safe by construction, but only as long as the collective is gated on that mode being active. Worth making the gate explicit rather than relying on the invariant holding, especially since a filter or transform raising `StopIteration` mid-stream would turn a clean crash into a cluster-wide hang.

Keeping the vote's semantics as "every micro batch in this step carries `padding_flag`" also keeps it consistent with what `count_loss_token` does today.

veomni/data/dataset.py
LiuzcEECS: "guaranteeing >=1 real micro-batch per step" does not resolve the problem of `mean_global_loss`,  because `mean_global_loss` is applied on each micro batch.
The solution here is to update the loss to be 0 in mean_global_loss when we see a padding

veomni/utils/loss_utils.py
LiuzcEECS: I end up moving the padding and infinity logic into DistributedDataloader because we might need to remove the VeOmniIter in future.
Now the the padding and infinity logic only works in worker mode and worker mode cannot work with VeOmniIter's background fetcher.
Because the calculation of start_step after resuming is based on the assumption that each epoch runs full steps of data generations, we wont stop the loop now even all ranks are generating padding data but we will mark the padding micro batch with 0 loss in `mean_global_loss`

PR #868, opened 2026-06-29
[model] fix: pass swiglu limit in deepseek v4 fused moe

zhangxin81:
### What does this PR do?

> Passes DeepSeek-V4's `swiglu_limit` into the direct `fused_moe_forward(...)` path so the fused MoE kernel preserves the model's gated SwiGLU clamp semantics. Also adds a regression test that forces the DeepSeek-V4 fused branch and asserts the merged expert weights and `swiglu_limit` are forwarded correctly.

### Checklist Before Starting

- Search for relative PRs/issues and link here: follows up on fused MoE swiglu-limit support.

### Test

> - `pytest tests/models/test_deepseek_v4_fused_moe.py -q`
> - `ruff check veomni/models/transformers/deepseek_v4/deepseek_v4_gpu_patch_gen_config.py tests/models/test_deepseek_v4_fused_moe.py`
> - `ruff format --check veomni/models/transformers/deepseek_v4/deepseek_v4_gpu_patch_gen_config.py tests/models/test_deepseek_v4_fused_moe.py`
> - `make quality`
> - `/veomni-review`: safe

### API and Usage Example

> N/A

### Design & Code Changes

> - Updates the DeepSeek-V4 patchgen config to pass `swiglu_limit=self.limit` to the direct fused MoE call.
> - Regenerates the DeepSeek-V4 generated modeling file and diff via patchgen.
> - Adds a unit test covering the direct fused path argument forwarding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Our guiding principles when building VeOmni are:
 | Model                                                    | Model size                    | Example config File                                                   |
 | -------------------------------------------------------- | ----------------------------- | ----------------------------------------------------------------------|
 | [DeepSeek2.5/3/R1](https://huggingface.co/deepseek-ai)   | 236B/671B                     | [deepseek.yaml](configs/text/deepseek.yaml)                           |
+| [DeepSeek-V4](https://huggingface.co/deepseek-ai)        | Checkpoint-dependent          | [deepseek.yaml](configs/text/deepseek.yaml) template; set V4 path and eager attention (MoE stays `fused_triton` on GPU) |
 | [Llama3-3.3](https://huggingface.co/meta-llama)          | 1B/3B/8B/70B                  | [llama3.yaml](configs/text/llama3.yaml)                               |
 | [Qwen2-3](https://huggingface.co/Qwen)                   | 0.5B/1.5B/3B/7B/14B/32B/72B/  | [qwen2_5.yaml](configs/text/qwen2_5.yaml)                             |
 | [Qwen2-3 VL/QVQ](https://huggingface.co/Qwen)            | 2B/3B/7B/32B/72B              | [qwen3_vl_dense.yaml](configs/multimodal/qwen3_vl/qwen3_vl_dense.yaml)|
--- a/configs/text/deepseek.yaml
+++ b/configs/text/deepseek.yaml
@@ -1,6 +1,10 @@
 model:
+  # Set this to the target DeepSeek checkpoint, e.g. DeepSeek-V3/R1 or DeepSeek-V4.
   # model_path: deepseek-ai/DeepSeek-V3-0324
+  # model_path: <path-or-id-to-DeepSeek-V4-checkpoint>
   ops_implementation:
+    # DeepSeek-V4 attention is eager-only; set this to eager for V4 checkpoints.
+    # Keep moe_implementation as fused_triton on GPU so V4 uses clamp-aware fused MoE.
     attn_implementation: flash_attention_2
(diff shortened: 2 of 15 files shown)


Luosuu:
@zhangxin81 please fix CI

zhangxin81:
### More info for last commit(I think it is very useful):
# Fused MoE Backward: Scattered Row Count

This note documents a generic correctness pitfall in the GPU `fused_triton`
MoE backward path. It was exposed by DeepSeek-V4 hash-MoE routing, but it is not
DeepSeek-V4-specific.

## Summary

The fused MoE forward/backward path expands each original token by its top-k
routes before running grouped GEMMs. Therefore, the scattered activation or
gradient tensor has shape:

```text
[tokens * topk, hidden]
```

Grouped GEMM launch bounds must cover the scattered rows. In the non-EP fused
MoE backward path, using the original token count as `max_M` / `max_K` is only
safe when no expert receives more than `tokens` scattered rows. The correct
bound is the scattered row count after `moe_scatter()`, for example:

```python
grad_fc2_output = moe_scatter(grad_output, scatter_index)
num_scattered_tokens = grad_fc2_output.shape[0]
```

Then pass `num_scattered_tokens` to the downstream grouped GEMM calls.

## Why this can be hidden for other MoE models

Consider `tokens = 3` and `topk = 2`.

If each token routes to two different experts:

```text
token0 -> expert0, expert1
token1 -> expert0, expert1
token2 -> expert0, expert1
```

After grouping by expert:

```text
expert0: 3 rows
expert1: 3 rows
```

Each expert group has at most `tokens` rows, so an old launch bound like
`max_M = tokens` may still cover every group. The code is still conceptually
wrong, but this routing pattern may not expose the bug.

## Why duplicate top-k experts expose the bug

DeepSeek-V4 hash-MoE can route multiple top-k slots of the same token to the
same expert. For example:

```text
token0 -> expert0, expert0
token1 -> expert0, expert0
token2 -> expert0, expert0
```

After grouping by expert:

```text
expert0: 6 rows
```

The expert group now needs `tokens * topk` rows. If backward grouped GEMM is
launched with `max_M = tokens`, only the first half is computed:

```text
computed:
  token0-slot0
  token0-slot1
  token1-slot0

not computed / stale:
  token1-slot1
  token2-slot0
  token2-slot1
```

Those stale rows can still be consumed by later gather or gradient operations,
which may introduce arbitrary values, `inf`, or `nan`. A visible symptom is a
non-finite gradient norm, such as `gnorm = nan`.

## Key takeaway

This is a generic fused MoE backward correctness issue:

> DeepSeek-V4 did not introduce a model-specific issue; it exposed a latent
> generic bug in the fused MoE backward kernel.

Any model that uses the GPU `fused_triton` non-EP MoE path with `topk > 1` can
hit this if routing allows one expert group to receive more than `tokens`
scattered rows. Duplicate top-k expert assignments make the failure easy to
reproduce, but the invariant is broader: grouped GEMM bounds must match the
scattered tensor row count, not the original token count.

## Regression test pattern

A focused regression test should construct duplicate top-k routes explicitly:

```text
selected_experts = zeros([tokens, topk])
routing_weights = full([tokens, topk], weight)
```

Then compare split-fc1 fused, merged-fc1 fused, and eager MoE backward results,
and assert all produced gradients are finite.

FoolPlayer (approved):
Reviewed the actual PR diff against VeOmni constraints. The changes look safe to me.

FoolPlayer (approved):

PR #875, opened 2026-07-02

YLlllllllllll:
## What & Why

```
KeyError: 'betas'
```

## Fix

- `filter_empty_param_groups()` — drop empty param groups in `build_optimizer`, so the saved/loaded group layout is consistent (no empty group to lose hyperparams).
- `restore_optimizer_param_group_defaults()` — re-apply `optimizer.defaults` to every param group after `set_optimizer_state_dict`, as a safety net.
- Call the restore in `OptimizerState.load_state_dict` (`dcp_checkpointer.py`).

## Tests / CI

- New `tests/optim/test_optimizer_vlm_param_groups.py`:
  - empty-vit + backbone groups → filtered correctly,
  - `restore_optimizer_param_group_defaults` repairs a group missing `betas`.
- Wired into `.github/workflows/gpu_unit_tests.yml`. The existing DCP save/load tests only exercise the **default** optimizer and never the VLM two-param-group path, which is why this slipped through.

## Validation

Verified end-to-end on Qwen3.5-9B + ShareGPT4V (`freeze_vit=true`, sp2, DCP), same pod A/B, only variable = this patch:

### DIFF
--- a/.github/workflows/gpu_unit_tests.yml
+++ b/.github/workflows/gpu_unit_tests.yml
@@ -172,6 +172,7 @@ jobs:
           uv run --frozen pytest -s -x tests/optim/test_muon_unit.py
           uv run --frozen pytest -s -x tests/optim/test_muon_fsdp2_parity.py
           uv run --frozen pytest -s -x tests/optim/test_muon_fsdp2_smoke.py
+          uv run --frozen pytest -s -x tests/optim/test_optimizer_vlm_param_groups.py
       - name: Run LoRA e2e test
         # peft ships under the `gpu` extra (rolled in from the dropped `lora`
         # extra), so the earlier `uv sync --extra gpu --group dev` step is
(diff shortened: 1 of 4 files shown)


YLlllllllllll:
Thanks for the review — all four points addressed in the updated commit:

1. **Generators/iterators as `params`**: `filter_empty_param_groups` now materializes `params` into a list, so empty generators are correctly dropped (not `bool(generator) == True`) and non-empty ones are not exhausted before the optimizer consumes them.
2. **`MultiOptimizer` support**: `restore_optimizer_param_group_defaults` now recurses into `optimizers_dict` sub-optimizers (which have no top-level `defaults`/`param_groups`).
3. **Auto-built groups not filtered**: filtering is now also applied after the automatic decay/no-decay group construction (covers the all-frozen-decay case that yields an empty decay group).
4. **Early-return bypass**: `restore_optimizer_param_group_defaults(self.optimizer)` is now also called on the `should_extra_parallel_aware` path before `return`.

Added regression tests for the generator-materialization and MultiOptimizer-recursion cases.

YLlllllllllll:
recheck

YLlllllllllll:
recheck

PR #876, opened 2026-07-02

YLlllllllllll:
## What & Why

```
KeyError: 'betas'
```

## Fix

- **`filter_empty_param_groups()`** — drop empty param groups in `build_optimizer` so the saved/loaded group layout stays consistent. `params` may be a generator/iterator, so it is **materialized to a list** (an empty generator is `True` under `bool()` and would otherwise slip through, and it must not be exhausted before the optimizer consumes it). Applied to **both** caller-provided groups (VLM vit/backbone) **and** the auto-built decay/no-decay groups (an all-frozen-decay model yields an empty decay group with the same failure mode).
- **`restore_optimizer_param_group_defaults()`** — re-apply `optimizer.defaults` to every param group after `set_optimizer_state_dict`, as a safety net. It **recurses into `MultiOptimizer` sub-optimizers** (ExtraParallel+FSDP2 / Muon) via `optimizers_dict`, which have no top-level `defaults`/`param_groups`.
- Call the restore in `OptimizerState.load_state_dict` for **both** the single-optimizer path and the `should_extra_parallel_aware` early-return path.

## Tests / CI

New `tests/optim/test_optimizer_vlm_param_groups.py`:
- empty vit + backbone groups → filtered correctly,
- **generator** `params` → empty dropped, non-empty materialized (not exhausted),
- `restore_optimizer_param_group_defaults` repairs a group missing `betas`,
- restore **recurses** into a `MultiOptimizer`-like container.

Wired into `.github/workflows/gpu_unit_tests.yml`. The existing DCP save/load tests only exercise the **default** optimizer and never the VLM two-param-group path, which is why this slipped through.

## Validation

Verified end-to-end on Qwen3.5-9B + ShareGPT4V (`freeze_vit=true`, sp2, DCP), same-pod A/B, only variable = this patch:

### DIFF
--- a/.github/workflows/gpu_unit_tests.yml
+++ b/.github/workflows/gpu_unit_tests.yml
@@ -172,6 +172,7 @@ jobs:
           uv run --frozen pytest -s -x tests/optim/test_muon_unit.py
           uv run --frozen pytest -s -x tests/optim/test_muon_fsdp2_parity.py
           uv run --frozen pytest -s -x tests/optim/test_muon_fsdp2_smoke.py
+          uv run --frozen pytest -s -x tests/optim/test_optimizer_vlm_param_groups.py
       - name: Run LoRA e2e test
         # peft ships under the `gpu` extra (rolled in from the dropped `lora`
         # extra), so the earlier `uv sync --extra gpu --group dev` step is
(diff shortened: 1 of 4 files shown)


YLlllllllllll:
Both addressed in the latest push:

1. `filter_empty_param_groups` now also drops `requires_grad=False` params, so a group containing only frozen params (no optimizer state) is treated as empty — added a regression test `test_filter_empty_param_groups_drops_frozen_only_group`.
2. `restore_optimizer_param_group_defaults` now retrieves `param_groups` via `getattr` and returns early if either `defaults` or `param_groups` is missing, guarding against non-standard/mock optimizers.

PR #878, opened 2026-07-02

YLlllllllllll:
## What & Why

```
KeyError: 'betas'
```

## Fix

- **`filter_empty_param_groups()`** — drop empty param groups in `build_optimizer` so the saved/loaded group layout stays consistent. Robustness details:
  - `params` may be a generator/iterator, so it is **materialized to a list** (an empty generator is `True` under `bool()` and would otherwise slip through, and it must not be exhausted before the optimizer consumes it).
  - **frozen params (`requires_grad=False`) are dropped**, so a group containing only frozen params (which never gets optimizer state) is treated as empty and cannot trigger the same resume `KeyError`.
  - applied to **both** caller-provided groups (VLM vit/backbone) **and** the auto-built decay/no-decay groups (an all-frozen-decay model yields an empty decay group with the same failure mode).
- **`restore_optimizer_param_group_defaults()`** — re-apply `optimizer.defaults` to every param group after `set_optimizer_state_dict`, as a safety net.
  - **recurses into `MultiOptimizer` sub-optimizers** (ExtraParallel+FSDP2 / Muon) via `optimizers_dict`, which have no top-level `defaults`/`param_groups`.
  - retrieves both `defaults` and `param_groups` via `getattr` and returns early if either is missing (guards against non-standard/mock optimizers).
- Call the restore in `OptimizerState.load_state_dict` for **both** the single-optimizer path and the `should_extra_parallel_aware` early-return path.

## Tests / CI

New `tests/optim/test_optimizer_vlm_param_groups.py`:
- empty vit + backbone groups → filtered correctly,
- group of only **frozen** (`requires_grad=False`) params → dropped,
- **generator** `params` → empty dropped, non-empty materialized (not exhausted),
- `restore_optimizer_param_group_defaults` repairs a group missing `betas`,
- restore **recurses** into a `MultiOptimizer`-like container.

Wired into `.github/workflows/gpu_unit_tests.yml`. The existing DCP save/load tests only exercise the **default** optimizer and never the VLM two-param-group path, which is why this slipped through.

## Validation

Verified end-to-end on Qwen3.5-9B + ShareGPT4V (`freeze_vit=true`, sp2, DCP), same-pod A/B, only variable = this patch:


Luosuu (approved):

Coach257 (changes requested):
**Root cause is an unconditionally-created empty param group, not the loader**

The empty group originates in `VLMTrainer._build_optimizer` (`vlm_trainer.py:229-240`). It already filters `requires_grad`, so under `freeze_vit` `vit_params == []`, yet the vit group is still appended unconditionally:

```python
vit_params, other_params = [], []
for name, param in self.base.model.named_parameters():
    if param.requires_grad:
        (vit_params if "visual" in name else other_params).append(param)

param_groups = [
    {"params": vit_params, "lr": args.train.vit_lr},   # empty when freeze_vit
    {"params": other_params, "lr": args.train.optimizer.lr},
]
```

Guarding the append is the actual root fix (and it also covers `VLMRLTrainer`, which inherits this method):

```python
param_groups = []
if vit_params:
    param_groups.append({"params": vit_params, "lr": args.train.vit_lr})
if other_params:
    param_groups.append({"params": other_params, "lr": args.train.optimizer.lr})
```

YLlllllllllll (commented):
Addressed — root cause fixed in `VLMTrainer._build_optimizer`: only append vit/backbone groups when they have trainable params (covers `VLMRLTrainer` via inheritance). Kept `filter_empty_param_groups` + DCP `restore_optimizer_param_group_defaults` as defense in depth. Added regression test `test_vlm_trainer_style_param_groups_skip_empty_vit`.

YLlllllllllll:
Fixed per @Coach257 — `VLMTrainer._build_optimizer` now skips empty vit/backbone groups. Defense-in-depth filter/restore kept. Ready for re-review.

Coach257 (approved):

PR #810, opened 2026-06-01
[dist, trainer] feat: add CUDA graph compile support

wang-zerui:
### What does this PR do?

Adds CUDA Graph friendly per-block `torch.compile` support for FSDP2 training, addressing #401.

The implementation compiles leaf FSDP target module forwards in place so module identity and FSDP2 hooks are preserved, keeps FSDP collectives outside compiled regions, enforces padded static packed lengths for the compile path, and marks training-step boundaries for PyTorch CUDA Graph Trees when available.

### Checklist Before Starting

- Search for relative PRs/issues and link here: #401
- PR title follows `[{modules}] {type}: {description}` format (enforced by [check_pr_title.yml](.github/workflows/check_pr_title.yml))
  - **Allowed modules:** `agent`, `ci`, `ckpt`, `config`, `data`, `dist`, `docker`, `docs`, `logging`, `lora`, `misc`, `model`, `omni`, `optim`, `ops`, `parallel`, `perf`, `release`, `task`, `trainer`
  - **Allowed types:** `feat`, `fix`, `refactor`, `chore`, `test`
  - Breaking changes: prepend `[BREAKING]` — e.g. `[BREAKING][parallel, model] feat: dynamic batching`

### Test

Official VeOmni GPU image:

- `PYTHONPATH=/workspace/VeOmni /app/.venv/bin/python -m pytest tests/distributed/test_torch_compile.py -q` (10 passed)
- `PATH=/app/.venv/bin:$PATH make quality`

Additional manual GPU probes on a 2x A100 SXM4 official VeOmni image: FSDP2 toy model with 3 per-layer `torch.compile(mode="reduce-overhead")` blocks, `torch.compiler.cudagraph_mark_step_begin()`, grad norm, loss allreduce, and 5 optimizer steps completed successfully (`compiled=3`). A gradient-accumulation variant with 2 micro-batches per optimizer step also completed 5 optimizer steps / 10 compiled forward-backward passes.

### API and Usage Example

New training arguments:

- `train.enable_compile=True`
- `train.compile_backend`, `train.compile_mode`, `train.compile_fullgraph`, `train.compile_dynamic`

`train.enable_compile=True` requires `train.dyn_bsz=True` and `train.pad_to_length=True` so packed inputs have static shapes for CUDA Graph capture.

### Design & Code Changes

- Add `veomni.distributed.torch_compile` helpers to select leaf target modules, compile module forwards, and mark CUDA graph step boundaries.
[...]

PR #881, opened 2026-07-02
[dist, trainer] feat: add FSDP2 torch compile support

zhangxin81:
### What does this PR do?

Follow-up to #810, based on @wang-zerui's CUDA graph compile support work.

This PR adds CUDA Graph friendly per-block `torch.compile` support for FSDP2 text training. It compiles FSDP2 leaf target module forwards in place before `fully_shard()`, marks CUDA graph step boundaries during training, and keeps compatibility aliases for the original compile options from #810.

### Checklist Before Starting

- Search for relative PRs/issues and link here: #810, #401

### Test

- `python -m pytest tests/distributed/test_torch_compile.py -q`
  - `14 passed in 31.37s`
- `make quality`
  - `ruff check tasks tests veomni docs`: passed
  - `ruff format --check tasks tests veomni docs`: passed
- GPU validation summary is included in `tests/pr810_compile/REPORT.md`:
  - 2-GPU FSDP2 baseline and compile-enabled toy text training completed.
  - Compile-enabled run compiled the two toy decoder layer forwards.
  - Profiling showed `cudaGraphLaunch` activity only in the compile run.
  - Loss and memory behavior stayed aligned with the non-compile baseline.
- Additional 2-GPU H100 FSDP2 end-to-end steady-state benchmark:
  - Qwen3-toy text training, synthetic plaintext data, sequence length 512, `global_batch=2`, `micro_batch=1`.
  - Timed 80 training steps with GPU events and excluded the first 30 compile warmup steps.
  - `mode=reduce-overhead` improved p50 step time from 65.20 ms to 60.20 ms (**-7.67%**) and throughput from 15.34 it/s to 16.61 it/s (**+8.31%**).
  - One-time compile warmup overhead was about 14.3 s, amortized after about 2,860 steps.
  - Loss curve and peak VRAM stayed at parity with the non-compile baseline.

### Performance Comparison

Performance details are summarized in `tests/pr810_compile/REPORT.md`, with an additional end-to-end steady-state benchmark from the 2-GPU H100 validation run. Key results:

#### 1-GPU API micro-benchmark

| Metric | Baseline | Compile | Delta |
| --- | ---: | ---: | ---: |
| Active-window wall time | 60.27 ms | 40.50 ms | **-32.8%** |
| GPU kernel launch count | 1380 | 690 | **-50.0%** |
| Host-side CUDA launch count | 1242 | 408 | **-67.1%** |
| `cudaGraphLaunch` count | 0 | 12 | **+12** |
| Total GPU busy time | 35.12 ms | 31.88 ms | -9.2% |

This shows the intended CUDA Graph replay effect clearly: compile does not materially change the underlying math kernels, but it significantly reduces host-side launch overhead.

#### 2-GPU H100 FSDP2 end-to-end steady-state benchmark

This benchmark uses Qwen3-toy text training with FSDP2 on 2× H100. Timing is based on GPU events over 80 training steps, with the first 30 steps excluded to avoid Dynamo tracing, Inductor codegen, recompilation, and initial CUDA Graph capture overhead. The table below reports the remaining 50 post-warmup steps.

| Metric | Baseline | Compile | Delta |
| --- | ---: | ---: | ---: |
| Mean ms/step | 65.62 | 60.81 | **-7.32%** |
| p50 ms/step | 65.20 | 60.20 | **-7.67%** |
| p90 ms/step | 69.12 | 63.25 | **-8.49%** |
| End-to-end throughput (`1 / p50`) | 15.34 it/s | 16.61 it/s | **+8.31%** |

The compile-enabled run adds about 14.3 s of one-time warmup overhead in the first 30 steps, mainly from Dynamo tracing, Inductor codegen, and CUDA Graph capture. With a steady-state saving of about 5.0 ms/step, the warmup cost is amortized after about 2,860 steps. For long SFT or pretraining runs, this cost should be negligible.

#### 2-GPU FSDP2 profile signal

| Metric | Baseline | Compile | Delta |
| --- | ---: | ---: | ---: |
| GPU kernel launch count | 1908 | 1219 | **-36%** |
| Host-side CUDA launch count | 1722 | 918 | **-47%** |
| `cudaGraphLaunch` count | 0 | 12 | **+12** |
| Peak VRAM | 18.16 GB | 18.16 GB | parity |
| Matched-step loss curve | 12.83 → 11.53 | 12.83 → 11.53 | parity |

The separate narrow 2-GPU profile window confirms the production signal: the real FSDP2 path produced CUDA Graph launches while preserving loss and memory parity. That profile window also captured compile/CUDA Graph warmup and FSDP2 communication effects, so the steady-state benchmark above is the throughput claim.

### API and Usage Example

```yaml
train:
  dyn_bsz: true
  pad_to_length: true
  torch_compile:
    enable: true
    mode: reduce-overhead
    fullgraph: false
    dynamic: false
```

Deprecated aliases remain supported and are mapped to `train.torch_compile.*` with warnings:

- `train.enable_compile`
- `train.compile_backend`
- `train.compile_mode`
- `train.compile_fullgraph`
- `train.compile_dynamic`

### Design & Code Changes

- Add `TorchCompileConfig` under `train.torch_compile.*`.
- Add `veomni.distributed.torch_compile` helpers for:
  - selecting leaf compile target modules,
  - compiling module forwards without replacing module identity,
  - marking CUDA graph step boundaries when available.
- Wire compile config through `BaseTrainer.build_model()` into FSDP2 parallelization.
- Compile only CUDA + FSDP2 leaf target module forwards and reject unsupported DDP, non-FSDP, non-CUDA, and ExtraParallel cases.
- Guard the feature to text data paths requiring `dyn_bsz=True` and `pad_to_length=True` for static packed shapes.
- Add unit tests and usage docs.

- [x] Added/updated documentation
- [x] Added tests to CI workflow (or explained why not feasible)


veomni/arguments/arguments_types.py
```diff
@@ -432,6 +432,32 @@ class CheckpointConfig:
...
+@dataclass
+class TorchCompileConfig:
+    """train.torch_compile.* — Per-block torch.compile options."""
+
+    enable: bool = field(
+        default=False,
+        metadata={"help": "Enable per-block torch.compile for CUDA Graph friendly text training."},
+    )
+    backend: Optional[str] = field(
+        default=None,
```
Luosuu: default cudagraphs

veomni/arguments/arguments_types.py
```diff
@@ -432,6 +432,32 @@ class CheckpointConfig:
...
+        metadata={"help": "Backend passed to torch.compile. None uses PyTorch's default backend."},
+    )
+    mode: Optional[str] = field(
+        default="reduce-overhead",
+        metadata={"help": "Mode passed to torch.compile. 'reduce-overhead' enables CUDA Graphs when possible."},
+    )
+    fullgraph: bool = field(
+        default=False,
```
Luosuu: true

veomni/arguments/arguments_types.py
```diff
@@ -555,20 +574,32 @@ class TrainingArguments:
...
-            raise ValueError(
-                f"dyn_bsz_physical_overflow_ratio must be >= 1.0, got {self.dyn_bsz_physical_overflow_ratio}."
-            )
-
         self._train_steps = -1
         self.local_rank = int(os.getenv("LOCAL_RANK", 0))
         self.global_rank = int(os.getenv("RANK", 0))
         self.world_size = int(os.getenv("WORLD_SIZE", 1))
 
+        deprecated_compile_fields = {
```
Luosuu: no need to keep compat with deprecated args

veomni/distributed/torch_compile.py
```diff
@@ -0,0 +1,138 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
Luosuu, line 1: let's keep this simple first, we only compile decoder blocker of each model, which means other parts are not included (like ViT, LM head).

Luosuu (commented):
overall LGTM!

FoolPlayer:
**Review: needs-attention.** The feature is opt-in, well-gated, and doesn't touch default paths. Two items should be resolved before merge.

## High

1. **Default `mode="reduce-overhead"` (CUDA Graphs) vs FSDP2 resharding.** CUDA graphs capture params by static address, but `reshard_after_forward=True` (the default) frees/re-all-gathers per-block params each step, often at a new address → cudagraph-trees re-records every iteration (no speedup + memory growth), worst case silent wrong results. The committed report only covers a 2-layer toy at `micro_batch=1`, which doesn't generalize. Suggest defaulting to `None`/`"default"` (inductor, no cudagraphs) and making `reduce-overhead` an explicit opt-in that requires `reshard_after_forward=False`; validate on a multi-layer model with resharding.

2. **`mark_compile_step_begin()` granularity under gradient accumulation.** It is called once per optimizer step, but compiled forwards run once per micro-batch (× gradient-checkpoint recompute). With `num_micro_steps > 1` (the common path) this can raise `accessing tensor output of CUDAGraphs that has been overwritten by a subsequent run`, or grow the graph memory pool across micro-steps. This path is untested (benchmark used `micro_batch=1`). Suggest moving the call inside the micro-batch loop (before each `forward_backward_step`), or asserting `gradient_accumulation_steps == 1` when compile is enabled.

## Medium

3. **Text-only guard is opt-out, with no runtime re-check.** `DataArguments.supports_torch_compile` defaults to `True`, so any future multimodal args that forget `= False` (and use a text `data_type`) would slip through both arg-level checks; `parallelize_model_fsdp2` does not re-check data/trainer type. Current VLM/DiT/DPO/Omni are safe (they inherit `VeOmniArguments.__post_init__`), but consider flipping to opt-in (base `False`, text trainers set `True`) and/or enforcing at the parallelize layer.

## Low / nits

4. `TorchCompileConfig` (arguments) and `distributed.torch_compile.CompileConfig` are duplicated and bridged by field-name reflection in `base.py` — reuse one class, or add a field-parity test.
5. `select_leaf_compile_modules` compiles the root when `""` is the only target (pulls the whole model's FSDP collectives into the compiled region) — skip/raise instead.
6. Deprecated-alias merge (`value is not None and value is not False`) can't express a `False`/`None` override — harmless today, worth a doc note ("new `torch_compile.*` wins; aliases only set non-default values").
7. `mark_compile_step_begin` added to `BaseTrainer.train_step` is effectively dead (Text/DPO trainers use their own `train_step` via composition); `tests/pr810_compile/REPORT.md` is a doc rather than a test and is named `pr810` under PR #881.

## Verified OK

Compile-before-`fully_shard` keeps FSDP all-gather/reshard outside the compiled region; EP is explicitly rejected; all non-FSDP2 paths in `build_parallelize_model` raise (no silent no-op); NPU is excluded via `get_device_type()`; PR title is valid; comments are English; no unused imports.

zhangxin81:
# PR #881 — Response to Review Comments

Thanks for the review! Below is a per-comment breakdown of what changed
in the follow-up commit, plus an explanation for the one place where I
did **not** follow the suggestion literally, together with the GPU
evidence that motivated it.

---

## What changed in this revision

### 1. `TorchCompileConfig.backend` — "default cudagraphs"

**Not adopted as-is. Kept `"inductor"` as the default; documented
`"cudagraphs"` as a first-class supported value.**

See the "Why the default is `inductor`, not `cudagraphs`" section below
for the rationale and the H100 evidence.

### 2. `TorchCompileConfig.fullgraph` — "true"

**Adopted.** Default changed from `False` → `True`.

```python
fullgraph: bool = field(
    default=True,
    metadata={"help": "Whether to pass fullgraph=True to torch.compile."},
)
```

### 3. `TrainingArguments` — "no need to keep compat with deprecated args"

**Adopted.** All five deprecated aliases and their migration block are
removed:

- `train.enable_compile`
- `train.compile_backend`
- `train.compile_mode`
- `train.compile_fullgraph`
- `train.compile_dynamic`

...along with the `deprecated_compile_fields` handler in
`TrainingArguments.__post_init__`. `docs/usage/arguments.md` and
`tests/pr810_compile/REPORT.md` are updated accordingly, and the legacy-
alias unit tests are removed. Users who want per-block compile now pass
`train.torch_compile.*` directly.

### 4. `veomni/distributed/torch_compile.py` — "keep this simple; only compile decoder blocks"

**Adopted.** The file is simplified to a single entry point
`compile_decoder_blocks(model, cfg)` that walks `model.named_modules()`
and compiles the `forward` method of every module whose class name ends
in `DecoderLayer` (in place, preserving module identity so FSDP2 hooks
stay outside the compiled region). ViT / VisionBlock / LM head /
embeddings are no longer touched.

- Removed: `select_leaf_compile_modules`, `compile_module_forwards`,
  the seen-set de-dup, the parent/leaf FQN bookkeeping.
- Call site in `veomni/distributed/torch_parallelize.py` reduces to
  `compile_decoder_blocks(model, compile_config)`.

Selection rule reference: HF modeling files consistently use the
`*DecoderLayer` suffix — e.g. `Qwen3DecoderLayer`, `Qwen2DecoderLayer`,
`Qwen3MoeDecoderLayer`, `Qwen2_5_VLDecoderLayer`, `LlamaDecoderLayer`,
`DeepseekV3DecoderLayer`, `SeedOssDecoderLayer`. Vision / audio /
diffusion blocks use `*VisionBlock` / `*Block` / `*EncoderLayer` and are
correctly excluded.

### 5. New guardrail — `backend="cudagraphs"` + `mode="reduce-overhead"`

**Added.** `torch.compile`'s `cudagraphs` backend does **not** accept a
`mode=` kwarg (it's an inductor-only knob). This combination previously
survived config validation and only exploded at first forward with a
confusing dynamo stack trace:

```
torch._dynamo.exc.BackendCompilerFailed: backend='cudagraphs' raised:
TypeError: CudagraphsBackend.__call__() got an unexpected keyword
argument 'mode'
```

`compile_decoder_blocks` now raises early with an actionable message:

```python
if compile_config.mode is not None and compile_config.backend == "cudagraphs":
    raise ValueError(
        "train.torch_compile.mode is not accepted by the 'cudagraphs' backend. "
        "Leave mode=None with backend='cudagraphs', or switch backend to 'inductor'."
    )
```

Covered by two new unit tests (`rejects_mode_with_cudagraphs_backend`,
`accepts_cudagraphs_backend_without_mode`).

---

## Why the default is `inductor`, not `cudagraphs`

I want to flag this one explicitly because it does not follow the
review comment literally.

### The two are not interchangeable

`torch.compile` treats `backend="cudagraphs"` and
`backend="inductor", mode="reduce-overhead"` as **different paths**,
not aliases:

| Path                                            | Dynamo trace | Inductor lowering / codegen | CUDA Graph capture |
| ----------------------------------------------- | :----------: | :-------------------------: | :----------------: |
| `backend="inductor", mode="reduce-overhead"`    |      ✅      |             ✅              |         ✅         |
| `backend="cudagraphs"`                          |      ✅      |             ❌              |         ✅         |

- `inductor + reduce-overhead` = **fusion + Triton codegen + CUDA Graph
  replay**. Speedup comes from two independent sources: kernel-level
  fusion (real GPU-time reduction) *and* host-side launch overhead
  elimination.
- `cudagraphs` = **CUDA Graph replay only**. It hands the Dynamo
  `GraphModule` straight to CUDA Graph capture and skips Inductor
  entirely. Speedup only comes from launch overhead elimination; the
  underlying math kernels are the same as eager.

Concretely, the `cudagraphs` backend does not accept `mode=`,
`options=`, or Inductor-side flags at all — its call signature is
different. So switching the default to `cudagraphs` is not a one-token
change; it forces `mode=None` for every downstream user.

### The measured evidence on 1× H100 (this PR's benchmark)

From the 1-GPU micro-benchmark in `tests/pr810_compile/REPORT.md`,
comparing eager vs the compile path used here:

| Metric                             | Baseline | Compile | Δ            |
| ---------------------------------- | -------: | ------: | ------------ |
| Active-window wall time            |  60.2 ms | 40.5 ms | **-32.8%**   |
| GPU kernel launch count            |     1380 |     690 | **-50.0%**   |
| Host-side CUDA launch count        |     1242 |     408 | **-67.1%**   |
| `cudaGraphLaunch` count            |        0 |      12 | **+12**      |
| Total GPU busy time                |  35.1 ms | 31.8 ms | **-9.2%**    |

The bulk of the win is **wall-time reduction from fewer host launches
+ CUDA Graph replay**, but there is *also* a 9.2% drop in "GPU busy
time" — i.e. Inductor is producing kernels that spend less time on the
GPU than the eager equivalent. That component would go to zero if we
switched to `cudagraphs`, because Inductor is no longer in the pipeline
and no fusion / codegen happens.

Steady-state 2× H100 FSDP2 also holds up: `-7.32% mean ms/step`,
`+8.31% throughput` at parity loss / VRAM.

### End-to-end sanity check on H100

I re-ran the smoke on a live H100 (SM 9.0) with the defaults from this
revision:

```
TorchCompileConfig(enable=False, backend='inductor',
                   mode='reduce-overhead', fullgraph=True, dynamic=False)
```

- 4 decoder blocks compiled, ViT / LM head / embed correctly skipped.
- 3 forward + backward steps with `cudagraph_mark_step_begin()` between
  them completed without errors.
- Unit tests: **14/14 pass** on-GPU (`pytest tests/distributed/test_torch_compile.py -v`).
- `ruff check` + `ruff format --check` clean.

For contrast, forcing `backend="cudagraphs" + mode="reduce-overhead"`
on the same box reproduces the `TypeError` above at first forward. This
is what motivated the new early-`ValueError` guard.

### Recommendation

- Keep **`inductor` + `reduce-overhead`** as the default. It's the
  combination that (a) matches the benchmarked numbers already in the
  PR description, (b) captures both fusion and CUDA Graph wins, and
  (c) is what mainline PyTorch documents as the recommended CUDA-Graph
  path.
- **Fully support `cudagraphs`** as a first-class opt-in via
  `train.torch_compile.backend="cudagraphs"` for workloads that
  already use externally-fused kernels (Liger norm/RoPE, FlashAttention,
  fused SwiGLU) and want to skip Inductor's ~14 s warmup. Users pick
  their tradeoff explicitly.

Happy to flip the default if there's a specific workload where
`cudagraphs`-only outperforms `inductor + reduce-overhead` on H100 — if
you can share that, I'll re-benchmark and follow up.

---

## CI / verification

- All 17 pre-review CI check runs on `ef4f089` were `success` (17/17).
- 14 unit tests in `tests/distributed/test_torch_compile.py` — passing.
- `ruff check` + `ruff format --check` — clean.
- H100 (SM 9.0) end-to-end smoke — compiled 4 decoder blocks, 3-step
  fwd+bwd loop with `cudagraph_mark_step_begin()` — passing.

veomni/distributed/torch_compile.py
```diff
@@ -0,0 +1,109 @@
...
+    """Runtime options for compiling FSDP2 decoder blocks."""
+
+    enable: bool = False
+    backend: Optional[str] = "inductor"
+    mode: Optional[str] = "reduce-overhead"
+    fullgraph: bool = True
+    dynamic: bool = False
+
+
+def _is_decoder_block(module: nn.Module) -> bool:
```
Luosuu: typically the _no_split_modules attribute of the model class is the decoder layer name

veomni/distributed/torch_compile.py
```diff
@@ -0,0 +1,109 @@
...
+
+logger = logging.get_logger(__name__)
+
+
+@dataclass
+class CompileConfig:
+    """Runtime options for compiling FSDP2 decoder blocks."""
+
+    enable: bool = False
+    backend: Optional[str] = "inductor"
```
Luosuu, line 34: cudagraphs

Luosuu (approved):

zhangxin81:
Follow-up on the `torch.compile` + FSDP2 CUDA Graphs concern:

I checked TorchTitan's main training path and aligned VeOmni with that behavior.

TorchTitan does **not** use `torch.compile(mode="reduce-overhead")` by default in its main path. Its `CompileConfig` only exposes `backend` (default `"inductor"`), and `torchtitan/distributed/compile.py` compiles each TransformerBlock with:

```python
transformer_block.compile(backend=backend, fullgraph=True)
```

So the TorchTitan main path is:

- compile each transformer block before FSDP2 wrapping
- use `backend="inductor"`
- do not pass `mode`, so PyTorch uses the inductor default (`mode=None`)
- keep FSDP2 `reshard_after_forward` as a separate parallelism policy

That means TorchTitan avoids the problematic default combination by not enabling CUDA Graph replay through `reduce-overhead` in the first place. Its experimental `graph_trainer` has a custom cudagraph pass and wrapper, but that is a separate graph-training path rather than the default model compile path.

I updated VeOmni accordingly:

- `train.torch_compile.mode` now defaults to `None`, matching TorchTitan's main-path behavior of using inductor without CUDA Graph replay by default.
- `mode="reduce-overhead"` remains available as an explicit opt-in.
- CUDA Graph modes (`mode="reduce-overhead"` or `backend="cudagraphs"`) now require `train.accelerator.fsdp_config.reshard_after_forward=False`; otherwise VeOmni raises early with an actionable error.
- `cudagraph_mark_step_begin()` is only called when CUDA Graphs are actually enabled, and it is called per micro-batch before forward/backward so gradient accumulation has correct graph step boundaries.
- Decoder-block selection now follows the model class `_no_split_modules` decoder-layer entries, matching the reviewer suggestion.

Validation run locally in the PR worktree:

```text
PYTHONPATH=/Users/bytedance/Documents/VeOmni-pr881-review \
  /Users/bytedance/Documents/VeOmni/.venv/bin/python -m pytest tests/distributed/test_torch_compile.py -q
# 19 passed

PATH=/Users/bytedance/Documents/VeOmni/.venv/bin:$PATH make quality
# ruff check + ruff format --check passed

python3 tests/special_sanity/check_device_api_usage.py -d veomni
python3 tests/special_sanity/check_device_api_usage.py -d tests
# passed
```

Luosuu (approved):

zhangxin81:
# PR #881 compile_default Performance Analysis

## Setup

- GPU: 1 x H100 80GB HBM3 via `CUDA_VISIBLE_DEVICES=0`
- Torch: `2.7.1`
- CUDA build: `12.8`
- dtype: bf16
- Test target: eager vs `compile_default`

`compile_default` means:

```python
torch.compile(
    backend="inductor",
    mode=None,
    fullgraph=True,
    dynamic=False,
)
```

applied to decoder layer forwards, matching the current PR default.

## Method

I used a synthetic Qwen3-style decoder stack to isolate decoder compute:

- 2 decoder layers
- RMSNorm + Q/K/V/O projections
- causal SDPA
- SwiGLU MLP
- residual connections
- forward + backward
- median of 10 measured steps after 4 warmup steps
- first real step is recorded separately to capture Dynamo/Inductor compile cost

Shapes:

| case | hidden | intermediate | heads | kv heads | configured layers |
| --- | ---: | ---: | ---: | ---: | ---: |
| `tiny_0p15b_shape` | 768 | 2048 | 12 | 4 | 8 |
| `small_0p6b_shape` | 1024 | 3072 | 16 | 8 | 12 |
| `mid_1p7b_shape` | 2048 | 6144 | 16 | 8 | 12 |
| `large_4b_shape` | 2560 | 7680 | 20 | 4 | 16 |

Matrix:

- `seq_len`: 512, 1024, 2048
- `batch_size`: 1, 2, 4
- 36 comparable eager / compile_default pairs

## Overall Results

- Completed comparable shapes: 36
- `compile_default` steady-state wins: 36 / 36
- Geomean `compile_default / eager` median step ratio: `0.764x`
- Mean steady-state delta: `-23.5%`
- Mean peak memory delta: `-13.4%`

Important: this is steady-state only. First-step compile cost remains large.

## By Model Scale

| model | n | wins | geomean compile/eager | mean delta | mean peak mem delta | mean amortization steps |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `tiny_0p15b_shape` | 9 | 9 | 0.755 | -24.4% | -8.9% | 5204 |
| `small_0p6b_shape` | 9 | 9 | 0.747 | -25.2% | -12.2% | 2308 |
| `mid_1p7b_shape` | 9 | 9 | 0.768 | -23.1% | -16.8% | 2317 |
| `large_4b_shape` | 9 | 9 | 0.786 | -21.2% | -15.8% | 4651 |

Takeaway:

- The best relative result was the small / mid shapes.
- The large shape still wins in steady state, but the relative gain is smaller because larger GEMMs / SDPA already keep the GPU busier and leave less Python / elementwise overhead to remove.
- Tiny also wins, but first-step compile cost is harder to amortize because each steady-state step is short.

## By Sequence Length

| seq_len | n | wins | geomean compile/eager | mean delta | mean peak mem delta | mean amortization steps |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 512 | 12 | 12 | 0.770 | -22.9% | -2.9% | 6519 |
| 1024 | 12 | 12 | 0.762 | -23.7% | -15.2% | 2442 |
| 2048 | 12 | 12 | 0.760 | -23.8% | -22.1% | 1899 |

Takeaway:

- Steady-state speedup is similar from 512 to 2048 in this decoder-only benchmark.
- Amortization gets much better at longer sequence lengths because each step saves more absolute time.
- Peak memory improvement is also larger at longer sequence lengths.

## By Batch Size

| batch_size | n | wins | geomean compile/eager | mean delta | mean peak mem delta | mean amortization steps |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 12 | 12 | 0.782 | -21.6% | -4.9% | 6423 |
| 2 | 12 | 12 | 0.769 | -23.1% | -14.7% | 2582 |
| 4 | 12 | 12 | 0.742 | -25.7% | -20.6% | 1855 |

Takeaway:

- Larger batch improves the value of `compile_default`.
- Batch 1 still wins in steady state, but compile cost is hardest to amortize.
- Batch 4 has the best mean speedup and memory reduction.

## Best Shapes

| model | seq | batch | eager ms | compile ms | delta | amortization steps |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `small_0p6b_shape` | 2048 | 4 | 7.028 | 4.766 | -32.2% | 1104 |
| `tiny_0p15b_shape` | 2048 | 4 | 5.224 | 3.664 | -29.9% | 1928 |
| `mid_1p7b_shape` | 1024 | 4 | 7.947 | 5.703 | -28.2% | 984 |
| `tiny_0p15b_shape` | 1024 | 1 | 4.110 | 2.964 | -27.9% | 1970 |
| `small_0p6b_shape` | 1024 | 1 | 4.134 | 2.987 | -27.8% | 1788 |

## Weakest Shapes

These still improved in steady state, but had the weakest deltas or worst amortization.

| model | seq | batch | eager ms | compile ms | delta | amortization steps |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `large_4b_shape` | 1024 | 1 | 3.770 | 3.339 | -11.4% | 6127 |
| `mid_1p7b_shape` | 2048 | 1 | 4.745 | 4.039 | -14.9% | 3931 |
| `tiny_0p15b_shape` | 512 | 1 | 3.522 | 2.994 | -15.0% | 27327 |
| `large_4b_shape` | 2048 | 1 | 6.058 | 5.144 | -15.1% | 3018 |
| `large_4b_shape` | 512 | 2 | 3.730 | 3.065 | -17.8% | 5691 |

## Default Compile Path

The current default is intentionally the safe compile path:

- `backend="inductor"`
- `mode=None`
- safe with default `reshard_after_forward=True`
- no CUDA Graph replay

The previous `mode="reduce-overhead"` path should be treated as a separate opt-in performance mode, not as the default behavior.

## Practical Conclusion

For repeated static-shape decoder compute, `compile_default` is beneficial in steady state on H100:

- about 24% geomean faster in this matrix
- lower peak allocated memory in most larger-token shapes

But it is not a free default performance win:

- first-step compile cost is large
- small / short jobs may not amortize it
- shape stability is required; repeated recompiles can erase the steady-state gain

For PR messaging, the safest wording is:

> `compile_default` improves steady-state decoder-block compute for repeated static shapes, while keeping CUDA Graph replay disabled by default. The main tradeoff is compile-time amortization: long, static-shape runs benefit; short or shape-changing runs may not. CUDA Graph replay remains an explicit opt-in path via `mode="reduce-overhead"` with `reshard_after_forward=False`.

zhangxin81:
# PR #881 compile_default Fusion Analysis

## Summary

`compile_default` in PR #881 means:

```python
torch.compile(
    backend="inductor",
    mode=None,
    fullgraph=True,
    dynamic=False,
)
```

It does **not** enable CUDA Graph replay, because `mode=None`. The measured steady-state speedup therefore mainly comes from Inductor lowering and Triton fusion inside each decoder block.

In the benchmark artifacts, the generated Triton / Inductor cache contains representative fused kernels such as:

```text
triton_red_fused__to_copy_add_mean_mul_pow_rsqrt_*
triton_per_fused__to_copy_add_mean_mul_pow_rsqrt_*
triton_red_fused__to_copy_add_div_mul_pow_sum_*
triton_poi_fused_add_fill_mul_sigmoid_silu_sub_*
triton_poi_fused_mul_silu_*
triton_poi_fused_add_*
triton_poi_fused_clone_*
```

These names show that the wins come from fusing normalization, activation, residual, cast/copy, and backward pointwise/reduction chains around the main GEMM and attention kernels.

## Fused Operator Groups

### 1. RMSNorm Forward

Qwen3-style RMSNorm is roughly:

```python
x_fp32 = x.float()
variance = x_fp32.pow(2).mean(-1, keepdim=True)
x_norm = x_fp32 * torch.rsqrt(variance + eps)
out = weight * x_norm
out = out.to(bf16)
```

Representative generated kernels:

```text
triton_red_fused__to_copy_add_mean_mul_pow_rsqrt_*
triton_per_fused__to_copy_add_mean_mul_pow_rsqrt_*
```

These correspond to fusion across:

- dtype cast / `_to_copy`
- square / `pow`
- mean reduction
- epsilon add
- `rsqrt`
- scale by RMSNorm weight
- cast back to bf16

In eager execution, these operations often materialize several intermediate tensors and launch multiple small kernels. Inductor fuses them into one or a small number of Triton kernels, reducing launch overhead and HBM traffic.

### 2. RMSNorm Backward

Representative generated kernels:

```text
triton_red_fused__to_copy_add_div_mul_pow_sum_*
triton_per_fused__to_copy_add_div_mul_pow_sum_*
triton_red_fused__to_copy_add_mul_sum_*
```

These map to RMSNorm backward-style reduction and pointwise chains:

- dtype casts
- square / `pow`
- `sum`
- division
- multiplication
- addition

RMSNorm backward contains reductions plus pointwise transforms. Inductor can combine parts of these chains into Triton reduction kernels and pointwise kernels, reducing separate reduction/elementwise launches.

### 3. SwiGLU / MLP Activation

Qwen3 MLP uses a SwiGLU-style block:

```python
gate = gate_proj(x)
up = up_proj(x)
hidden = silu(gate) * up
out = down_proj(hidden)
```

`SiLU(x)` is effectively:

```python
x * sigmoid(x)
```

Representative generated kernels:

```text
triton_poi_fused_add_fill_mul_sigmoid_silu_sub_*
triton_poi_fused_mul_silu_*
```

These indicate fusion across:

- `sigmoid`
- `silu`
- multiplication with the up projection
- related pointwise operations from autograd
- some mask/fill/sub/add operations introduced by the graph or backward path

The GEMMs themselves still run as external high-performance matrix kernels, but Inductor fuses the activation and surrounding pointwise work.

### 4. Residual Add and Layout / Copy Helpers

Decoder blocks include residual paths such as:

```python
hidden_states = residual + attn_out
hidden_states = residual + mlp_out
```

Representative generated kernels:

```text
triton_poi_fused_add_*
triton_poi_fused_clone_*
```

These correspond to:

- residual additions
- casts
- contiguous / clone-like layout materialization
- adjacent pointwise operations that can be combined or reordered

This reduces standalone add/copy kernels and can also reduce temporary buffer pressure.

## What Is Not Fused

`compile_default` does not fuse the whole decoder block into one monolithic kernel.

The following usually remain external optimized kernels:

- Q/K/V/O projection GEMMs
- MLP gate/up/down projection GEMMs
- SDPA / attention kernels
- large matrix multiplications handled by cuBLAS or equivalent extern kernels

Therefore, the speedup is not from turning attention and all GEMMs into a single fused kernel. The improvement comes mostly from fusing the non-GEMM work around those kernels.

## Why This Improves Performance

The performance improvement comes from:

- fewer small CUDA kernel launches
- less intermediate tensor materialization
- fewer HBM write/read round trips
- fused RMSNorm forward and backward chains
- fused SwiGLU activation and backward pointwise chains
- fused residual add, cast, and copy/layout helpers
- static-shape specialization with `dynamic=False`

This matches the matrix benchmark result: `compile_default` improved steady-state decoder-block compute across the tested static shapes.

## Why The Gain Varies By Shape

For larger model or token shapes, a larger fraction of time is spent in GEMM and attention kernels that are already highly optimized. Inductor still improves the surrounding pointwise/reduction work, but the relative share of optimizable work is smaller.

For smaller shapes, pointwise and launch overhead are a larger fraction of total time, so fusion can help. However, the absolute time saved per step is smaller, making the first-step compile cost harder to amortize.

## Practical Interpretation

`compile_default` is beneficial when the workload repeatedly runs the same static decoder-block shapes. It improves steady-state decoder compute by fusing RMSNorm, SwiGLU, residual, cast/copy, and backward pointwise/reduction chains.

It should not be described as a CUDA Graph optimization path. CUDA Graph replay remains a separate explicit opt-in via `mode="reduce-overhead"` with the corresponding FSDP2 guardrails.

PR #882, opened 2026-07-03
[trainer, model, data] feat: integrate DeepSpec draft-model training

zhangxin81:
### What does this PR do?

Integrate **DeepSpec** speculative-decoding **draft-model training** into VeOmni, so draft models (DSpark / DFlash / Eagle3) train through VeOmni's distributed engine instead of DeepSpec's original single-node `mp.spawn` + `FSDP(NO_SHARD)` loop.

The training loop remains decoupled from the target model: DeepSpec builds the target hidden-state cache offline, and VeOmni trains only the draft `PreTrainedModel` over that tensor cache. This PR reuses DeepSpec's modeling, loss, and dataset code, and adapts only the interfaces VeOmni needs.

This update also routes the Qwen3 DSpark draft model through a patchgen-generated VeOmni modeling file so DSpark's imported Qwen3 `RMSNorm` and `MLP` can dispatch through `OpSlot("rms_norm", "standard")` and `OpSlot("swiglu_mlp", "standard")` for Liger-backed runs. RoPE and DeepSpec's custom DSpark loss stay eager because their shape/loss contracts differ from VeOmni's existing generic slots.

### Checklist Before Starting

- Related: DeepSpec repo for draft models and target-cache tooling. No existing VeOmni issue/PR was found for this integration.
- PR title follows `[{modules}] {type}: {description}`: `[trainer, model, data] feat: integrate DeepSpec draft-model training`.

### Test

Validation completed:

| Area | Result |
| --- | --- |
| Static checks | Passed `compileall` for the DeepSpec integration, model registry, trainer, task, and prep script paths. |
| Import safety | Passed: VeOmni imports without requiring `DEEPSPEC_PATH`; missing DeepSpec checkout reports an actionable error. |
| Adapter contracts | Passed fake-DeepSpec checks for path resolution, cache dataset adapter, cache collator, draft config, and algorithm inference. |
| 1xH100 DSpark smoke | Passed 3 optimizer steps, DCP checkpoint writes, and HF checkpoint export. |
| 2xH100 FSDP2 DSpark smoke | Passed 3 optimizer steps with `init_device=meta`, DCP checkpoint writes, and HF checkpoint export. |
| Checkpoint resume | Passed resume from `global_step_2` and saved `global_step_3`. |
| Parallelism guard | Passed: rejects HSDP/non-pure-FSDP settings that would mis-scale DeepSpec's world-size-scaled loss. |
| 2xH100 medium benchmark | Passed 80 optimizer steps and DCP checkpoint save. |
| 8xH100 production-shape validation | Passed 20-step and 120-step Qwen3-4B-shape DSpark runs. |
| Liger smoke | Passed 2xH100 smoke; logs confirm patched DSpark model and `rms_norm`/`swiglu_mlp` Liger dispatch. |
| Liger production-shape benchmark | Passed 120 optimizer steps on 8xH100. |

CPU pytest regressions were not clean on the CPU-only master environment for environment reasons, not PR-specific failures: missing `einops`, dataloader `pin_memory` requiring an accelerator, and checkpoint tests requiring CUDA/distributed support.

Performance summary from the synthetic production-shape benchmark:

| Path | Post-warmup throughput | Peak GPU memory |
| --- | ---: | ---: |
| Native DeepSpec `NO_SHARD` | ~5.0 steps/s | 42.21 GiB max from `nvidia-smi` sampling |
| VeOmni FSDP2, eager draft ops | ~5.45 steps/s | 18.53 GiB from VeOmni logger |
| VeOmni FSDP2 + Liger RMSNorm/SwiGLU | ~6.00 steps/s | 18.45 GiB from VeOmni logger |

Interpretation:

- VeOmni FSDP2 gives the main memory scalability benefit by sharding parameters, gradients, and optimizer state instead of replicating them with `NO_SHARD`.
- The Liger patch improves stable post-warmup training throughput by about `6.00 / 5.45 ~= 1.10x` over the eager VeOmni path, and about `6.00 / 5.00 ~= 1.20x` over the native DeepSpec estimate in this synthetic benchmark.
- Native DeepSpec completed training/checkpointing in the benchmark runs but repeatedly hung during distributed teardown, so native wrapper wall time is not comparable.

### API and Usage Example

New task entry, prep script, launcher, configs, and docs. No existing public VeOmni API is changed.

```bash
# 1. Build the draft init checkpoint. This uses the target model once, offline.
python scripts/deepspec/prepare_draft_init.py \
    --algorithm dspark --arch qwen3 \
    --target_model_name_or_path Qwen/Qwen3-4B \
    --output_dir ~/deepspec_init/dspark_qwen3_4b \
    --block_size 7 --num_draft_layers 5 --target_layer_ids 1 9 17 25 33 \
    --mask_token_id 151669 --num_anchors 512 \
    --markov_rank 256 --markov_head_type vanilla \
    --confidence_head_alpha 1.0 --confidence_head_with_markov \
    --loss_decay_gamma 4.0 --ce_loss_alpha 0.1 --l1_loss_alpha 0.9

# 2. Point configs/deepspec/dspark_qwen3_4b.yaml at the init checkpoint and target cache.

# 3. Launch. The launcher auto-detects a sibling DeepSpec/ checkout.
bash scripts/deepspec/train_draft.sh configs/deepspec/dspark_qwen3_4b.yaml
```

### Design & Code Changes

Full integration write-up: `docs/design/deepspec_draft_integration.md`.

Correctness anchor: DeepSpec scales its loss by `world_size` to cancel DDP/FSDP gradient averaging. Under pure FSDP data parallelism, where the FSDP shard group spans the full world and all other parallel dimensions are size 1, FSDP2's gradient averaging cancels DeepSpec's `x world_size` scaling exactly. `DraftModelTrainer._validate_parallelism` enforces this by rejecting SP/TP/EP/PP/CP/HSDP settings for now.

Change list:

- `veomni/integrations/deepspec/`: locate and lazily import a DeepSpec checkout via `DEEPSPEC_PATH` or a sibling `DeepSpec/` checkout.
- `veomni/models/transformers/deepspec_draft/`: register `model_type=deepspec_draft`, `DeepSpecDraftConfig`, and DeepSpec draft architectures.
- `veomni/models/transformers/deepspec_draft/dspark_qwen3_gpu_patch_gen_config.py`: patchgen config for Qwen3 DSpark draft RMSNorm/SwiGLU OpSlot routing.
- `veomni/models/transformers/deepspec_draft/generated/`: generated Qwen3 DSpark GPU modeling file and generated diff.
- `veomni/data/deepspec/`: adapt DeepSpec `CacheDataset` and `CacheCollator` to VeOmni's mapping dataloader contract.
- `veomni/trainer/deepspec/draft_trainer.py`: `DraftModelTrainer(BaseTrainer)` for model build, frozen target embeds/lm_head, dataset/collator setup, DeepSpec loss, grad accumulation, metrics, and checkpoint-compatible training.
- `tasks/train_deepspec_draft.py`, `scripts/deepspec/*`, `configs/deepspec/*`: task entry, init-checkpoint prep, launcher, and sample configs.

Current limits:

- Only pure FSDP data parallelism is enabled for DeepSpec draft training in this PR. SP/TP/EP/PP/CP/HSDP are intentionally rejected until their loss-scaling contracts are designed and tested.
- Production-shape benchmarks use synthetic but shape-correct Qwen3-4B target/cache tensors. They validate integration behavior, memory scale, and compute-path throughput, but they are not final model-quality metrics.
- Eagle3 long GPU testing was not run in this pass; DSpark was prioritized because it covers the prep script, cache adapter, loss bridge, FSDP2, DCP, resume, guard, and Liger patch path.

- [x] Applied pre-commit checks locally where available: `make quality`.
- [x] Added/updated documentation (`docs/design/deepspec_draft_integration.md`, `configs/deepspec/README.md`).
- [x] No `tasks/` scripts were moved/renamed; a new task entry was added.
- [ ] Added tests to CI workflow: not added in this PR because meaningful end-to-end validation requires DeepSpec checkout plus GPU target-cache artifacts.


veomni/trainer/deepspec/draft_trainer.py
```diff
@@ -0,0 +1,385 @@
...
+        self.on_step_end(loss=total_loss, loss_dict=total_loss_dict, grad_norm=grad_norm)
+
+    def _flush_deepspec_metrics(self) -> Dict[str, float]:
+        """Flush DeepSpec's distributed metric accumulator (all ranks call it).
+
+        ``deepspec.utils.metrics.flush`` performs collective all-reduces and
+        asserts a consistent metric schema across ranks, so every rank must call
+        it exactly once per step.
+        """
+        try:
```
FoolPlayer, line 376: Maybe it's better to pin a commit for Deepspec and add it to pyproject.toml

veomni/trainer/deepspec/draft_trainer.py
```diff
@@ -0,0 +1,385 @@
...
+        from the frozen target). VeOmni's ``build_optimizer`` only optimizes
+        ``requires_grad`` params, so flipping ``requires_grad`` here is enough to
+        keep them out of the optimizer and out of grad-norm clipping.
+        """
+        model = self.model
+        if hasattr(model, "set_embedding_head_trainable"):
+            model.set_embedding_head_trainable(False)
+        else:  # defensive: freeze by name if the helper is absent
+            for name, param in model.named_parameters():
+                if name.startswith("embed_tokens") or name.startswith("lm_head"):
```
FoolPlayer: sub string match like `"embed_tokens" in name or "lm_head" in name` maybe better

veomni/trainer/deepspec/draft_trainer.py
```diff
@@ -0,0 +1,385 @@
...
+        if getattr(self, "LOG_SAMPLE", False):
+            helper.print_example(example=micro_batch, rank=self.args.train.local_rank)
+            self.LOG_SAMPLE = False
+        return micro_batch
+
+    def forward_backward_step(self, micro_batch: Dict[str, torch.Tensor]):
+        micro_batch = self.preforward(micro_batch)
+
+        with self.model_fwd_context:
+            loss = self._compute_deepspec_loss(micro_batch)
```
FoolPlayer, line 257: We need get the loss reduce on global tokens, it seems _compute_deepspec_loss does not return the global loss.

FoolPlayer (approved):

zhangxin81:
@FoolPlayer thanks for the review — addressed in the latest commit. Summary of the three comments:

**1. Freeze `embed_tokens` / `lm_head` by substring match** — done. The fallback now uses `"embed_tokens" in name or "lm_head" in name` so nested HF param names (`model.embed_tokens.weight`) are matched.

**2. Loss should reduce over global tokens** — done. `train_step` now reports the true token-weighted global loss via a mean all-reduce over the FSDP group, which exactly inverts DeepSpec's `×world_size` backward scaling. Previously the reported `loss` came from DeepSpec's per-rank `reduction="mean"` metric (a mean-of-per-rank-means that mis-weights ranks with fewer tokens); it is now overridden with the global value. Gradients are unchanged.

**3. Pin DeepSpec and add it to `pyproject.toml`** — done, in two steps, because DeepSpec had no build backend (no `setup.py`/`pyproject.toml`), so it could not be added as a pip/uv dependency as-is:

- Opened **deepseek-ai/DeepSpec#54** to add DeepSpec's own `pyproject.toml` (setuptools backend, packages `deepspec*` only). This makes it pip/uv-installable.
- VeOmni now declares DeepSpec as an **opt-in `deepspec` extra**: a git source pinned by commit in `[tool.uv.sources]`, plus an empty `[[tool.uv.dependency-metadata]]` so DeepSpec's own `torch==2.9.1` / `transformers==5.10.2` pins don't conflict with VeOmni's stack (VeOmni already provides every runtime dep the imported `deepspec.modeling/data/utils` modules need). `uv lock` / `uv lock --check` pass; torch/transformers are not downgraded; the extra is gated so default `--extra gpu` installs and CI (`--locked`/`--frozen`) are unaffected. `DEEPSPEC_PATH` / sibling checkout remains as a dev fallback.

> **Note on the pin (temporary):** the `rev` currently points at the fork commit that carries the #54 packaging, since a working dependency must reference a commit that actually contains `pyproject.toml`. **Once deepseek-ai/DeepSpec#54 merges, I'll repoint `[tool.uv.sources].deepspec` to the upstream commit and re-lock.** This is flagged in the `pyproject.toml` comment and the design doc.

FoolPlayer:
## Re-review after the "address review feedback" update (`f333790`)

Thanks for the fast turnaround. I re-checked the update against the earlier feedback and verified the changed logic (dependency wiring, loss-reduction math, freeze fallback) plus the framework APIs it now relies on.

### Resolved (verified)

**1. DeepSpec is now a reproducible, git-pinned `deepspec` extra** (was a `sys.path` soft-dependency).
- Opt-in via `uv sync --extra gpu --extra deepspec`; doesn't touch the default install.
- Pinned by git commit → reproducible, no runtime env var required.
- `[[tool.uv.dependency-metadata]]` with empty `requires-dist` prevents DeepSpec's own `torch==2.9.1 / transformers==5.10.2` pins from conflicting with VeOmni's stack — the same escape hatch already used for `flash-qla`.
- `ensure_deepspec_importable()` now prefers an already-installed `deepspec` and only falls back to `$DEEPSPEC_PATH` / a sibling checkout for local dev. `uv.md` + design doc updated to match.

**2. Headline `loss` is now reduced over global tokens.**
- `global_loss = all_reduce(total_loss, op="mean", group=fsdp_group)`, used for both the logged `loss` and `on_step_end`.
- Math checks out: DeepSpec's backward loss carries a `world_size` factor to cancel FSDP gradient averaging; a mean all-reduce over the fsdp group (== world under the pure-FSDP guard) cancels it exactly, giving the true token-weighted mean instead of a mean-of-per-rank-means. Reducing over the fsdp group (not the default world group) also keeps it idempotent w.r.t. VeOmni's metric callback. `all_reduce` accepts a Python float and `parallel_state.fsdp_group` exists, so the call is valid.

**3. Freeze fallback bug fixed.**
- `name.startswith(...)` → substring match (`"embed_tokens" in name` / `"lm_head" in name`) is a genuine fix: under HF's nested naming the params are `model.embed_tokens.weight` / `lm_head.weight`, so the old prefix match froze nothing in the fallback path.

### Follow-ups (non-blocking except the first)

**1. The git source points to a personal fork (`zhangxin81/DeepSpec`), pending upstream `deepseek-ai/DeepSpec#54`.** Commit-pinning protects against force-push, but a personal fork can be deleted outright — a supply-chain risk for `main`. Recommend repointing to the upstream commit (or an org-controlled mirror) before merge; there's already a TODO to do so once #54 lands.

**2. Grad-accum weighting (minor).** The reported `global_loss` is a plain mean over micro-batches of each micro-batch's global token-weighted mean, not a single token-weighted mean across the whole accumulated batch. Negligible when per-micro-batch token counts are similar, and it matches DeepSpec's per-micro-batch semantics. The new comment's phrasing "world_size * global_token_loss on every rank" is slightly loose (the per-rank value differs; only the reduced value equals the global loss).

**3. `_flush_deepspec_metrics()` wraps a collective in try/except.** `deepspec.utils.metrics.flush()` runs internal all-reduces; if it raises on some ranks but not others, peers can hang. Low probability (the design requires every rank to call it once per step), but the guard can't un-hang peers.

**4. No CI/unit coverage** for the adapter path — inherent to an opt-in extra + external target cache, so mostly a documentation matter.

### Verdict

High-quality update — the three substantive issues (non-reproducible dependency, non-global loss logging, no-op freeze fallback) are genuinely resolved, and the loss fix is mathematically sound. The only item worth resolving before merge is the personal-fork git source; the rest are fine as follow-ups.

zhangxin81:
## End-to-end equivalence validation: Native DeepSpec vs VeOmni FSDP2

Following the review, I ran a full **before/after** experiment to empirically confirm that the FSDP2 draft-training path reproduces the reference Native DeepSpec trainer — both in training dynamics and in downstream speculative-decoding acceptance. This is the equivalence evidence for the adapter path called out as the (non-blocking) "no CI coverage" follow-up.

**Setup (only the training engine varies).** 8×H100-80GB, torch 2.11.0, `torch.compile` off, same seed.

| | Before (baseline) | After (this PR) |
|---|---|---|
| Trainer | Native DeepSpec (`train.py`, `mp.spawn` + FSDP `NO_SHARD`) | VeOmni `tasks/train_deepspec_draft.py` (FSDP2) |
| Draft model | DSpark, Qwen3-4B target, `block_size=7`, `target_layer_ids=[1,9,17,25,33]` | identical |
| Target-hidden cache | shared (`num_samples=4914`, `hidden=2560`, `seqlen=1024`) | identical |
| Draft init ckpt | shared (1393.1M params) | identical |
| Global batch / steps | 256 / 2000 | 256 / 2000 |

Both consume the **same** target cache and draft-init checkpoint, so the only variable is the engine.

### 1. Training converges equivalently

| Metric | Native (before) | VeOmni PR#882 (after) |
|---|---|---|
| Final training loss | **0.2481** | **0.26** |
| Final train `accept_rate@0` / `@6` | — | 0.95 / 0.94 |
| Exit code | 0 | 0 |

Both start near loss ~4.1 and converge to ~0.25 on the same trajectory. This confirms the gradient-scaling-equivalence argument in the design doc holds in practice: under the pure-FSDP guard, FSDP2's `AVG` reduce-scatter exactly cancels DeepSpec's `×world_size` backward factor, so the FSDP2 run reproduces the replicated `NO_SHARD` gradient — now with real param/grad/optimizer sharding.

### 2. Downstream acceptance rate — parity (marginally ahead)

Evaluated both final draft checkpoints with DeepSpec's `Qwen3DSparkEvaluator` (target `Qwen/Qwen3-4B`, greedy `temperature=0`, `max_new_tokens=256`) on a fixed offline subset — gsm8k(100) + humaneval(100) + mt-bench(80) + alpaca(100), 380 samples, identical seed/subset for both.

| Metric | Native | VeOmni | Δ |
|---|---|---|---|
| **overall_acceptance_length** | **1.8887** | **1.9243** | **+0.0356 (+1.89%)** |
| overall_verify_rate | 0.2364 | 0.2409 | +0.0045 |

Per-dataset acceptance length, VeOmni is equal-or-higher on all four:

| Dataset (n) | Native AL | VeOmni AL | Δ AL |
|---|---|---|---|
| gsm8k (100) | 2.0687 | 2.0956 | +0.0269 |
| humaneval (100) | 2.1262 | 2.1860 | +0.0598 |
| mt-bench (80) | 1.6824 | 1.6908 | +0.0084 |
| alpaca (100) | 1.6363 | 1.6782 | +0.0419 |

Per-position `accept_rate@k` curves overlap tightly (block size 7). I read the +1.9% as **within run-to-run variance for a 2k-step run → parity**, not a robust improvement. The load-bearing conclusion is that the FSDP2 path is **correct and quality-neutral** relative to the reference trainer.

### 3. Kernel note (honest framing of the "after" run)

The two engines are *not* running identical kernels, and I want to be explicit about it:

- **Native (before):** stock HF Qwen3 — eager `Qwen3RMSNorm`, eager SwiGLU, HF cross-entropy; `flex_attention` for the draft blocks.
- **VeOmni (after):** the patchgen modeling binds `OpSlot`s to `OpsImplementationConfig`'s fused defaults. The YAML only overrode `attn_implementation`, so the omitted op fields kept their GPU defaults → the run actually used **Liger fused RMSNorm + fused SwiGLU + Liger CE** (plus `flex_attention`), via `apply_ops_config(args.model.ops_implementation)` in `draft_trainer.py`.

So this is genuinely **eager-native vs fused-VeOmni**. Steady-state per-step time was ~4.25 s/it (VeOmni) vs ~4.32 s/step (native) — a small **~1.6%** compute edge from the fused elementwise kernels. It's *not* the 1.3–1.8× that fused norm/MLP give on norm/MLP-bound models, and that's expected: the step is dominated by `flex_attention` over the block-7 draft structure and the cached-hidden-state pipeline, and the draft is tiny (1.4B, seqlen 1024), so the fusible fraction is small. Parity-with-a-slight-edge is the correct read for this workload.

(Wall-clock aside: VeOmni's DCP checkpointing fired every epoch during this run — a save-cadence artifact of the config, unrelated to per-step compute.)

### 4. One concrete follow-up surfaced during eval

VeOmni's HF export writes `model_type=deepspec_draft`, which `AutoConfig` (used by DeepSpec's `eval.py`) doesn't recognize. For this comparison I patched the exported `config.json` `model_type` → `qwen3` (the architecture is identical to the native DSpark export) and it loaded fine. The design doc already notes you load the exported weights into the matching DeepSpec draft class for eval; worth a one-line note in the README's "Checkpoints" section so downstream eval doesn't trip on it.

### Summary

- Training loss equivalent (0.2481 vs 0.26); same convergence trajectory.
- Downstream acceptance length **1.9243 vs 1.8887 (+1.9%, ≈parity)**, equal-or-higher on all 4 datasets.
- FSDP2 path is correct and quality-neutral vs the reference trainer, with a small compute edge from the fused-op defaults.

zhangxin81:
### Fix: patchgen drift check no longer requires the optional `deepspec` dependency

The `Check Patchgen Drift` CI job was failing (it actually pre-dates the upstream merge — it started with the first DeepSpec draft commit):

```
patchgen.codegen.CodegenError: Cannot import module
'deepspec.modeling.dspark.qwen3.modeling': No module named 'deepspec'
```

**Root cause.** `patchgen --check` discovers *every* `*_patch_gen_config.py` under `veomni/models/transformers/` and imports each config's `source_module` to regenerate and diff it. The DeepSpec draft config points at `deepspec.modeling.dspark.qwen3.modeling`, which is only installed via the optional `deepspec` extra. The CI job runs `uv run --extra gpu --dev patchgen --check` (no `deepspec`), so that one import raised `CodegenError` and aborted the whole job.

**Why a workflow-only fix doesn't work.** The set of checked configs is fixed by discovery, not by the CI command — the DeepSpec config file is always on disk and always gets imported. So no amount of path-filtering or `--extra` toggling in the YAML avoids the import; the tolerance has to live in patchgen itself.

**Fix (18 lines, 3 files).** Add an `optional: bool = False` field to `PatchConfig`. When set, `check_config` skips the config if its `source_module` can't be imported, instead of aborting:

```python
try:
    generator.load_source()
except CodegenError:
    if config.optional:
        print(f"  SKIP  {config.source_module} (optional source not importable)")
        return True
    raise
```

The DeepSpec draft config is marked `optional=True`. This turns `--extra deepspec` into a pure opt-in switch:

| Command | `deepspec_draft` |
|---|---|
| `uv run --extra gpu --dev patchgen --check` (CI default) | **SKIP** → job green |
| `uv run --extra deepspec --dev patchgen --check` | drift is verified |

**Why not just add `--extra deepspec` to CI?** That would make every PR pull the `deepspec` git source into the CI environment. Keeping the default job dependency-free avoids that and keeps the drift gate fast; maintainers touching the DeepSpec draft can verify drift locally with `--extra deepspec` (documented in a comment on the config). If we later want automated coverage, a `workflow_dispatch`/label-gated job is a cleaner place for it than the default gate.

Verified locally (torch present, `deepspec` absent — same shape as CI): 29 configs discovered, `deepspec_draft` **SKIP**, the other 28 **OK**, `ruff check` clean.

PR #889, opened 2026-07-07
[dist] feat: per-rank ExtraParallel-slice streaming weight loader + FSDP2 build fixes

Coach257:
### Summary
Adds an opt-in **faster, lower-memory** weight-loading path for large MoE checkpoints, plus two general FSDP2 build robustness fixes. Gated/off-by-default; existing loading paths are unchanged unless the new kwarg is set.

### 1. `ep_sharded_stream_load` — per-rank ExtraParallel-slice streaming loader
New `load_model_weights_ep_sharded` (`veomni/models/module_utils.py`), wired into `parallelize_model_fsdp2`'s every-rank-reads path behind `ep_sharded_stream_load=True`.

For params the model's parallel plan marks **ExtraParallel-sharded** (`Shard(0)` — MoE experts, embed parallel), **each rank reads only its `[E/ep, …]` dim-0 slice** straight from the safetensors shard via `safe_open(...).get_slice()[start:end]`. Dense params/buffers are read whole and dispatched exactly as `load_model_weights`; a sliced expert is dispatched with `parallel_plan=None` (not sliced twice), then FSDP's `dtensor_factory` shards it over the `ep_fsdp` sub-mesh.

**Why it's faster (vs the two existing paths):**
- **vs every-rank-full-read (`load_model_weights`):** today every one of the N ranks reads the *entire* expert set from disk/HDFS and then throws away `(ep-1)/ep` of it in host memory. Streaming makes the `ep` ranks read **disjoint `1/ep` slices in parallel**, so aggregate expert bytes read drop from `N × full` to ≈ the full set read once — and the full `[E,…]` tensor is never materialized, so peak host RAM drops ~`ep`-fold (also avoids the pinned-Shmem offload pressure that can OOM the cgroup).
- **vs rank0-broadcast (`rank0_load_and_broadcast_weights`):** no serialization through a single rank0 reader + broadcast; reads are fanned out across ranks.
- **Measured (downstream MoT, ~1.5 TiB, EP8+SP4+FSDP8):** `broadcast_model_weights_from_rank0=True` ≈ **1471.9 s** vs `ep_sharded_stream_load` ≈ **535.8 s** (~**2.75×** faster). (Numbers from our downstream branch; not re-measured against `main`.)

**Safe fallback:** raises `NotImplementedError` (→ caller falls back to `load_model_weights`) for unsupported cases: PEFT, no ExtraParallel plan, non-`Shard(0)` ExtraParallel, or a converter that needs the whole tensor set (see caveat below).

> ⚠️ **MoE checkpoints that need load-time key fusion must be pre-converted first.** Many HF MoE checkpoints store per-expert weights as separate keys (`experts.0.gate_proj.weight`, …) that VeOmni fuses into a single `[E, …]` tensor via a checkpoint-tensor converter at load time. Fusion needs the whole tensor set, which streaming can't provide, so such models **bail to `load_model_weights`**. To get the streaming speedup, first re-save the checkpoint once in the already-fused `[E, …]` layout the model expects natively (so no fusion converter is needed at load); then each rank can `get_slice()` its `[E/ep, …]` directly. Un-converted (fusion-required) checkpoints can only use the original loader.

### 2. Streamable checkpoint-tensor converters
Adds an **optional** `CheckpointTensorConverter.is_dim0_zero_pad(name) -> bool` capability (+ `checkpoint_converter_is_dim0_zero_pad` helper). A converter whose *only* transform is appending zero rows on dim-0 (which commutes with `Shard(0)`) stays streamable: the loader reads the real-row slice and zero-fills the tail.
- Documented as valid **only for semantically-inert padded rows** (e.g. a vocab/embedding table padded past real vocab). Explicitly **not** valid for MoE experts (a zero-padded "expert" is routable; its zero router logit can win top-k → silent corruption). VeOmni EP requires `num_experts % ep == 0` instead.

### 3. FSDP2 build fixes (`veomni/distributed/torch_parallelize.py`)
- **Wrap ALL ExtraParallel modules under a layer, not just the first.** A layer may hold >1 such module (e.g. multiple experts modules per decoder layer); each must be `fully_shard`'d over the `ep_fsdp` mesh, else the un-wrapped ones get re-sharded over the FSDP mesh and their already-EP-sliced params are corrupted. Also matches `layer_fqn + "."` so `layers.1` doesn't swallow `layers.10`.
- **Tolerate a globally-enabled ExtraParallel not declared by the model's plan** (a text/over-encoder has no experts): treat it as a no-op instead of crashing in `get_extra_parallel_fsdp_no_shard_info`.
- **Make `CPUOffloadPolicy.pin_memory` configurable** via `fsdp_offload_pin_memory` (default unchanged = `True`). Pinned host shards are non-reclaimable `Shmem` against the container memcg; for very large offloaded models N ranks/host can OOM the cgroup during load. Setting it `False` keeps shards in pageable anon memory.

### DIFF
--- a/veomni/distributed/torch_parallelize.py
+++ b/veomni/distributed/torch_parallelize.py
@@ -27,7 +27,7 @@ from torch.nn.parallel import DistributedDataParallel as DDP
 from torch.utils.checkpoint import noop_context_fn
 
 from ..arguments import MixedPrecisionConfig
-from ..models import load_model_weights, rank0_load_and_broadcast_weights
+from ..models import load_model_weights, load_model_weights_ep_sharded, rank0_load_and_broadcast_weights
 from ..utils import logging
 from ..utils.device import IS_NPU_AVAILABLE, get_device_type
 from .checkpoint import CheckpointFunction
@@ -142,7 +142,14 @@ def parallelize_model_fsdp2(
         _extra_parallel_mesh = {}
         _extra_parallel_map = {}
         for para in parallel_state.extra_parallel_names:
-            if parallel_state.extra_parallel_enabled(para):
+            # A globally-enabled ExtraParallel (e.g. ``ep``) may not be declared by
+            # THIS model's plan -- e.g. a text-encoder / over-encoder has no experts,
+            # so ``ep`` shards nothing here even though a MoE sibling module uses it.
+            # Treat such a para as a no-op for this model (``None``) instead of
+            # crashing in ``get_extra_parallel_fsdp_no_shard_info`` (which requires
+            # >=1 matching module). Nothing downstream wraps it (its layer para_mods
+            # stay empty).
+            if parallel_state.extra_parallel_enabled(para) and para in parallel_plan.extra_parallel_plan:
                 _extra_parallel_mesh[para] = parallel_state.extra_parallel_fsdp_device_mesh[para]
                 _extra_parallel_map[para] = parallel_plan.get_extra_parallel_fsdp_no_shard_info(model, para)
             else:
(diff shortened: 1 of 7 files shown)


veomni/distributed/torch_parallelize.py
```diff
@@ -409,17 +435,45 @@ def parallelize_model_fsdp2(
...
+                        model,
+                        weights_path,
+                        materialize_device,
+                        dtensor_factory=_dt_local_split,
+                        is_peft_model=is_peft_model,
+                        adapter_path=adapter_path,
+                        fqn_to_index_mapping=fqn_to_index_mapping,
+                    )
+                    loaded_ep_sharded = True
+                except NotImplementedError as e:
```
Luosuu: let's just fail early here? fallback is hard to manage

veomni/distributed/torch_parallelize.py
```diff
@@ -409,17 +435,45 @@ def parallelize_model_fsdp2(
...
-                adapter_path=adapter_path,
-                fqn_to_index_mapping=fqn_to_index_mapping,
-            )
+            # Opt-in fast/low-memory path for large MoE checkpoints: each rank reads
+            # only its ExtraParallel dim-0 slice of the expert tensors straight from
+            # the checkpoint (see ``load_model_weights_ep_sharded``). Only valid on the
+            # every-rank-reads (non-broadcast) path; unsupported model/checkpoint raises
+            # ``NotImplementedError`` and we fall back to the whole-tensor loader.
+            loaded_ep_sharded = False
+            if kwargs.get("ep_sharded_stream_load"):
```
Luosuu, line 439: how do we pass this kwarg? do we need a new field in job yaml

Coach257: fixed

veomni/distributed/torch_parallelize.py
Coach257: fixed

Luosuu (approved):

PR #891, opened 2026-07-08
[ckpt] feat: add dcp_save_to_lowest_rank option for DCP save planner

williamLyh:
## What

Add an opt-in `train.checkpoint.dcp_save_to_lowest_rank` flag (default `False`) that forwards `dedup_save_to_lowest_rank` to DCP's `DefaultSavePlanner`.

## Why

VeOmni's `execute_save` passes no planner to `dcp.save`/`dcp.async_save`, so it uses torch's default (`dedup_save_to_lowest_rank=False`), which load-balances each replicated shard's write across all replica holders. On a **non-shared filesystem** (each node writes to local disk), this scatters the single deduplicated checkpoint copy across nodes — the global `.metadata` on rank 0 then references shard files that physically live on other nodes, so no single node holds a loadable checkpoint.

## Behavior

When enabled, each replicated shard is written by the lowest-ranked replica holder, concentrating the deduplicated copy onto the lowest replica group. In the standard HSDP layout (FSDP shard within a node, replication across nodes) that group is one node, which then holds a complete, self-contained checkpoint. Only affects **replicated** data; unique expert/tensor/pipeline-parallel shards are never deduplicated and stay distributed. Trades write parallelism for locality — leave `False` when `output_dir` is shared.

Default (`False`) preserves the existing load-balanced write behavior.

## Tests

- `tests/checkpoints/test_dcp_checkpointer.py::TestSaveToLowestRank` — asserts the config default is `False` and that `execute_save` forwards the flag to the planner for both values.
- Also updates `docs/usage/arguments.md`.

### DIFF
--- a/docs/usage/arguments.md
+++ b/docs/usage/arguments.md
@@ -354,6 +354,7 @@ NPU validation runs at two times:
 | output_dir | `str` | `"output"` | Path to save model checkpoints. |
 | manager | `str` | `"dcp"` | Checkpoint manager. |
 | save_async | `bool` | `False` | Save checkpoints asynchronously. |
+| dcp_save_to_lowest_rank | `bool` | `False` | Write each replicated DCP shard from the lowest global rank that holds it instead of load-balancing across replicas. On a non-shared filesystem this concentrates the deduplicated copy onto the lowest-ranked replica group rather than scattering it across replicas; in the standard HSDP layout (shard within a node, replicate across nodes) that group is one node, which then holds a complete checkpoint. Only affects replicated data — unique expert/tensor/pipeline-parallel shards stay distributed. Leave `False` when `output_dir` is shared. |
 | load_path | `Optional[str]` | `None` | Path to checkpoint for resuming training. Use `"auto"` for auto-detection. |
 | save_steps | `int` | `0` | Steps between checkpoint saves. `0` to disable. |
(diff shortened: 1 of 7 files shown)


williamLyh:
@CLAassistant Many thanks for considering my PR. Here's some comments to the failing review:

---
This is a test-fixture problem, not a bug in production code

The error:
AttributeError: 'types.SimpleNamespace' object has no attribute 'dcp_save_to_lowest_rank'

Why it's a fixture issue:

- The failure originates in the test's hand-rolled _make_mock_trainer(), which builds a fake args.train.checkpoint out of a SimpleNamespace. That mock bypasses the real TrainingArguments, so it never inherits the default value you defined for dcp_save_to_lowest_rank in arguments_types.py.
- Production line — save_to_lowest_rank=args.train.checkpoint.dcp_save_to_lowest_rank — is correct. In any real run (or any test that goes through parse_args, like test_trainer_saveload.py), the field is guaranteed to exist. The only things that blow up are the unit tests that assemble their config from a bare SimpleNamespace.

So the fix is to update the test fixtures (add the new field to the mocks), not to change production code.
---

Many thanks again.

Best,
Yinhong

Coach257:
Please add `dcp_save_to_lowest_rank` to the `test.py` to pass the ci.

Luosuu (approved):

Luosuu:
@williamLyh thanks. please fix unit tests

PR #892, opened 2026-07-08
[trainer] feat: add channel loss logging callback

YLlllllllllll:
## Summary
- Add `train.channel_loss.*` config for detached per-channel causal-LM loss logging.
- Add a generic `ChannelLossCallback` that captures batch source metadata, strips non-model metadata before forward, and logs per-channel loss/token metrics through existing trainer metrics.
- Cover eager `loss_function` and fused `veomni_causal_lm_loss` OpSlot paths without changing the main training loss.

## Tests
- `PATH=.venv/bin:$PATH make quality`
- `TORCH_DEVICE_BACKEND_AUTOLOAD=0 PYTHONPATH=/home/tiger/Open-VeOmni-github uv run --project /home/tiger/modelchef/submodules/Open-VeOmni --frozen --group test python -m pytest /home/tiger/Open-VeOmni-github/tests/trainer/test_channel_loss_callback.py -q`

### DIFF
--- a/veomni/trainer/base.py
+++ b/veomni/trainer/base.py
@@ -32,6 +32,7 @@ import queue
 import threading
 from abc import ABC
 from collections import defaultdict
+from contextlib import nullcontext
 from dataclasses import asdict, fields
 from typing import Any, Callable, Dict, List
 
@@ -74,6 +75,7 @@ from ..utils.device import (
 from ..utils.loss_utils import count_loss_token, mean_global_loss
 from ..utils.model_utils import pretty_print_trainable_parameters
 from .callbacks import (
+    ChannelLossCallback,
     CheckpointerCallback,
     EnvironMeterCallback,
     EvaluateCallback,
@@ -577,9 +579,11 @@ class BaseTrainer(Stateful, ABC):
             self.hf_ckpt_callback = HuggingfaceCkptCallback(self)
         self.evaluate_callback = EvaluateCallback(self)
         self.moe_monitor_callback = MoERouterMonitorCallback(self)
+        self.channel_loss_callback = ChannelLossCallback(self)
         self.state = TrainerState()
 
     def on_train_begin(self):
+        self.channel_loss_callback.on_train_begin(self.state)
         self.environ_meter_callback.on_train_begin(self.state)
         self.tqdm_callback.on_train_begin(self.state)
         self.wandb_callback.on_train_begin(self.state)
@@ -598,6 +602,7 @@ class BaseTrainer(Stateful, ABC):
         self.hf_ckpt_callback.on_train_end(self.state)
         self.evaluate_callback.on_train_end(self.state)
         self.moe_monitor_callback.on_train_end(self.state)
+        self.channel_loss_callback.on_train_end(self.state)
 
     def on_epoch_begin(self):
         self.environ_meter_callback.on_epoch_begin(self.state)
@@ -617,7 +622,12 @@ class BaseTrainer(Stateful, ABC):
         self.hf_ckpt_callback.on_epoch_end(self.state)
         self.evaluate_callback.on_epoch_end(self.state)
 
-    def on_step_begin(self, micro_batches=None):
+    def on_step_begin(self, micro_batches=None, channel_loss_source_repeat: int = 1):
+        self.channel_loss_callback.on_step_begin(
+            self.state,
+            micro_batches=micro_batches,
+            source_repeat=channel_loss_source_repeat,
+        )
         self.environ_meter_callback.on_step_begin(self.state, micro_batches=micro_batches)
         self.tqdm_callback.on_step_begin(self.state, micro_batches=micro_batches)
         self.wandb_callback.on_step_begin(self.state, micro_batches=micro_batches)
@@ -629,6 +639,7 @@ class BaseTrainer(Stateful, ABC):
     def on_step_end(self, loss=None, loss_dict=None, grad_norm=None):
         self.environ_meter_callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
         self.tqdm_callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
+        self.channel_loss_callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
         self.wandb_callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
         self.profile_callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
         self.checkpointer_callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
@@ -671,21 +682,37 @@ class BaseTrainer(Stateful, ABC):
     def forward_backward_step(
         self, micro_batch: dict[str, torch.Tensor]
     ) -> tuple[torch.Tensor, dict[str, torch.Tensor]]:
-        micro_batch = self.preforward(micro_batch)
+        channel_loss_callback = getattr(self, "channel_loss_callback", None)
+        micro_step_context = (
+            channel_loss_callback.micro_step_context(self.state, micro_batch)
+            if channel_loss_callback is not None
+            else nullcontext()
+        )
+        with micro_step_context:
+            micro_batch = self.preforward(micro_batch)
+            if channel_loss_callback is not None:
+                channel_loss_callback.strip_model_inputs(micro_batch)
 
-        with self.model_fwd_context, set_batch_invariant_mode(self.args.train.enable_batch_invariant_mode):
-            outputs: ModelOutput = self.model(**micro_batch, use_cache=False)
+            channel_forward_context = (
+                channel_loss_callback.model_forward_context() if channel_loss_callback is not None else nullcontext()
+            )
+            with (
+                self.model_fwd_context,
+                set_batch_invariant_mode(self.args.train.enable_batch_invariant_mode),
+                channel_forward_context,
+            ):
+                outputs: ModelOutput = self.model(**micro_batch, use_cache=False)
 
-        loss: torch.Tensor
-        loss_dict: Dict[str, torch.Tensor]
-        loss, loss_dict = self.postforward(outputs, micro_batch)
+            loss: torch.Tensor
+            loss_dict: Dict[str, torch.Tensor]
+            loss, loss_dict = self.postforward(outputs, micro_batch)
 
-        # Backward pass
-        with self.model_bwd_context, set_batch_invariant_mode(self.args.train.enable_batch_invariant_mode):
-            loss.backward()
+            # Backward pass
+            with self.model_bwd_context, set_batch_invariant_mode(self.args.train.enable_batch_invariant_mode):
+                loss.backward()
 
-        del micro_batch
-        return loss, loss_dict
+            del micro_batch
+            return loss, loss_dict
 
     def model_reshard(self, micro_step: int, num_micro_steps: int):
         """Reshard model after backward pass."""
--- a/veomni/trainer/callbacks/channel_loss_callback.py
+++ b/veomni/trainer/callbacks/channel_loss_callback.py
@@ -1,0 +1,1526 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Detached per-channel loss logging for causal language-model training.
+
+The callback computes per-token cross entropy as an observability side channel.
+It never contributes to the returned model loss and all tensors are detached
+under ``torch.no_grad()`` before the extra CE work runs.
+"""
+
+from __future__ import annotations
+
+import importlib
+import inspect
+import re
+from collections import defaultdict
+from collections.abc import Hashable, Iterator
+from contextlib import contextmanager
+from contextvars import ContextVar
+from numbers import Integral
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
+from weakref import WeakSet
+
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+
+from ...distributed.parallel_state import get_parallel_state
+from ...utils.constants import IGNORE_INDEX
+from ...utils.logging import get_logger
+from .base import Callback, TrainerState
+
+
+logger = get_logger(__name__)
+
+
+if TYPE_CHECKING:
+    from ..base import BaseTrainer
+
+
+try:
+    from ...distributed.sequence_parallel import (
+        get_unified_sequence_parallel_group,
+        get_unified_sequence_parallel_rank,
+        get_unified_sequence_parallel_world_size,
+    )
+except ImportError:
+    get_unified_sequence_parallel_group = None  # type: ignore[assignment]
+    get_unified_sequence_parallel_rank = None  # type: ignore[assignment]
+    get_unified_sequence_parallel_world_size = None  # type: ignore[assignment]
+
+
+ChannelKey = Hashable
+_LOSS_CALL_POSITIONAL_KEYS = ("logits", "labels", "vocab_size", "num_items_in_batch", "ignore_index")
+_ACTIVE_CHANNEL_LOSS_COMPUTER: ContextVar[Optional["ChannelLossComputer"]] = ContextVar(
+    "veomni_active_channel_loss_computer", default=None
+)
+
+
+class _OpSlotDispatcher:
+    """Shared dispatcher for a module-global causal-loss OpSlot."""
+
+    def __init__(self, slot: Any, original_kernel: Callable[..., Any]) -> None:
+        self.slot = slot
+        self.original_kernel = original_kernel
+        self.owners: WeakSet[Any] = WeakSet()
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        result = self.original_kernel(*args, **kwargs)
+        owner = _ACTIVE_CHANNEL_LOSS_COMPUTER.get()
+        if owner is not None and owner in self.owners:
+            owner._observe_opslot_call(self.original_kernel, args, kwargs)
+        return result
+
+
+_OP_SLOT_DISPATCHERS: Dict[Any, _OpSlotDispatcher] = {}
+
+
+class ChannelLossMetadataError(ValueError):
+    """Raised when channel metadata cannot be aligned with packed segments."""
+
+
+def _is_sp_enabled() -> bool:
+    try:
+        return get_parallel_state().sp_enabled
+    except Exception:
+        return False
+
+
+class ChannelLossComputer:
+    """Computes detached per-source CE from model loss-function inputs."""
+
+    EAGER_CHUNK_SIZE = 512
+
+    def __init__(self) -> None:
+        self._original_loss_fn: Optional[Callable[..., Any]] = None
+        self._model_ref: Optional[torch.nn.Module] = None
+        self._installed = False
+        self._wrapped_opslots: List[Any] = []
+
+        self._source_ids: List[ChannelKey] = []
+        self._position_ids: Optional[torch.Tensor] = None
+        self._attention_mask: Optional[torch.Tensor] = None
+        self._result: Optional[List[Dict[str, Any]]] = None
+        self._per_mb_source_ids: List[List[ChannelKey]] = []
+        self._per_mb_position_ids: List[Optional[torch.Tensor]] = []
+        self._per_mb_attention_masks: List[Optional[torch.Tensor]] = []
+        self._micro_step = 0
+        self.step_totals: Dict[ChannelKey, tuple[torch.Tensor, torch.Tensor]] = {}
+        self.source_names: Dict[ChannelKey, str] = {}
+        self.strict = False
+
+    @staticmethod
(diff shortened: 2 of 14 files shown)


veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,769 @@
...
+        if not source_ids or not source_names:
+            return
+        if len(source_names) == 1:
+            for source_id in source_ids:
+                self.source_names[source_id] = source_names[0]
+            return
+        for source_id, name in zip(source_ids, source_names):
+            self.source_names[source_id] = name
+
+    def _wrapped_loss_fn(
```
FoolPlayer: This wrapper should be transparent. Replaying the original loss as fixed keyword arguments can break models that pass `num_items_in_batch`, `ignore_index`, or `num_labels` positionally. Prefer accepting `*args/**kwargs`, calling the original loss unchanged, then binding arguments only for side-channel extraction.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,769 @@
...
+    ) -> List[Dict[str, Any]]:
+        group = get_unified_sequence_parallel_group()
+        world = get_unified_sequence_parallel_world_size()
+        if group is None or world <= 1:
+            gathered = [local_segments]
+        else:
+            gathered: List[Optional[List[Dict[str, Any]]]] = [None] * world
+            dist.all_gather_object(gathered, local_segments, group=group)
+
+        all_segments: List[Dict[str, Any]] = []
```
FoolPlayer: Gathering local segments rank-major can merge continuation tokens into the wrong source when batch size is greater than one. Preserve global/batch segment order when reducing SP partials, and add an SP test covering `B > 1` with multiple sources.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,769 @@
...
+
+            start_indices = pos_flat.eq(0).nonzero(as_tuple=True)[0].cpu()
+            if start_indices.numel() == 0:
+                return []
+            end_indices = torch.cat([start_indices[1:] - 1, torch.tensor([seq_len - 1])])
+        else:
+            start_indices = torch.tensor([0])
+            end_indices = torch.tensor([seq_len - 1])
+
+        n_segments = min(len(start_indices), len(source_ids))
```
FoolPlayer: Using `min(len(start_indices), len(source_ids))` hides metadata/segment mismatches and drops later segments from channel metrics. Validate the counts; strict mode should raise, while non-strict mode should warn/skip or intentionally document singleton-source repetition.

FoolPlayer (commented):
Review notes for the channel loss callback implementation.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `1fcedf2f`: `_wrapped_loss_fn` now accepts `*args/**kwargs`, calls the original loss unchanged, and only binds the call afterward for side-channel extraction. Added a regression test for positional `num_items_in_batch` / `ignore_index` forwarding.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `1fcedf2f`: SP partials now carry `batch_idx`, `sp_rank`, and `local_order`, then reduce in batch-major global order so continuations attach to the correct source for `B > 1`. Added a B>1 multi-source SP regression test.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `1fcedf2f`: segment/source count mismatches are now validated. `strict=True` raises `ChannelLossMetadataError`; non-strict mode warns and skips the mismatched micro-batch. Added tests and updated the strict-mode docs.

YLlllllllllll:
Status update on NPU CI:

- The failed NPU unit run is from old head `1fcedf2f` (`NPU unit tests` run `28947602822`, job `86032196435`). The traceback is the same missing optional callback attribute fixed in the latest head: `AttributeError: 'TrainerTest' object has no attribute 'channel_loss_callback'`.
- Latest head is `28dce77a`, which guards `channel_loss_callback` access and adds `test_base_forward_backward_allows_missing_channel_loss_callback`. Local targeted test passes: `tests/trainer/test_channel_loss_callback.py -q` -> `11 passed`.
- The NPU unit/e2e workflows for `28dce77a` are currently `action_required` rather than failed. I do not have repository admin permission to approve those fork PR workflow runs, so a maintainer approval/rerun is needed for fresh NPU CI evidence.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,841 @@
...
+                    per_token_loss=per_token_loss,
+                    labels_flat=labels_flat,
+                    source_ids=source_ids,
+                    positions=pos,
+                    seq_len=seq_len,
+                    ignore_index=ignore_index,
+                    strict=strict,
+                )
+
+            start_indices = pos_flat.eq(0).nonzero(as_tuple=True)[0].cpu()
```
FoolPlayer: Tail padding can use `position_id=0`, so counting every `position_ids == 0` as a packed segment start can create fake segments and make source/segment counts mismatch. Filter out padding-only segments using labels/attention_mask before validating the count, and add `pad_to_length`/SP padding tests.

veomni/trainer/base.py
```diff
@@ -672,21 +678,28 @@ def postforward(
     def forward_backward_step(
         self, micro_batch: dict[str, torch.Tensor]
     ) -> tuple[torch.Tensor, dict[str, torch.Tensor]]:
-        micro_batch = self.preforward(micro_batch)
+        channel_loss_callback = getattr(self, "channel_loss_callback", None)
```
FoolPlayer, line 685: `on_micro_step_begin()` strips channel/source metadata before `preforward()`, so any trainer-specific `preforward` logic that needs `cur_token_num` or source metadata will see different inputs when channel loss is enabled. Extract metadata early, but remove model-only keys after `preforward()` and immediately before model forward.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,841 @@
...
+        hidden_states: Optional[torch.Tensor],
+        weights: Optional[torch.Tensor],
+        ignore_index: int,
+    ) -> Optional[torch.Tensor]:
+        if hidden_states is not None and weights is not None:
+            hs = hidden_states.detach().reshape(-1, hidden_states.size(-1))
+            w = weights.detach()
+            return self._eager_chunked(hs, w, labels_flat.to(hs.device), ignore_index)
+
+        if logits is not None:
```
FoolPlayer, line 469: The logits path converts the full `[tokens, vocab]` tensor to fp32 at once. For large vocabularies or long packed batches this observability path can OOM training; compute per-token CE in chunks, similar to the hidden_states path.

tests/trainer/conftest.py
```diff
@@ -0,0 +1,17 @@
...
+_real_find_spec = importlib.util.find_spec
+
+
+def _find_spec_without_torch_npu(name: str, package: str | None = None) -> ModuleSpec | None:
+    if name == "torch_npu":
+        return None
+    return _real_find_spec(name, package)
+
+
+importlib.util.find_spec = _find_spec_without_torch_npu  # type: ignore[assignment]
```
FoolPlayer: This conftest globally replaces `importlib.util.find_spec` for all tests under `tests/trainer`, making them all behave as if `torch_npu` is unavailable. Scope the patch to this test module or a fixture so future trainer NPU import-guard regressions are not masked.

FoolPlayer (commented):
Follow-up review after the latest update.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `9ca8d9a4`: padding-only `position_id == 0` segments are filtered before source/segment count validation. The filter uses shifted labels plus attention-mask/position evidence and is applied to both normal and SP aggregation. Added `pad_to_length`-style and SP padding regression tests.

veomni/trainer/base.py
YLlllllllllll: Fixed in `9ca8d9a4`: `on_micro_step_begin()` now only advances the channel-loss side-channel state. Metadata stripping happens after `preforward()` and immediately before model forward via `strip_model_inputs()`, so trainer-specific `preforward` logic still sees `cur_token_num` and source metadata. Added a BaseTrainer ordering regression test.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `9ca8d9a4`: the logits path now computes detached CE in `EAGER_CHUNK_SIZE` chunks instead of casting the full `[tokens, vocab]` tensor to fp32 at once. Added a regression test that forces chunk size 2 and verifies chunked CE calls.

tests/trainer/conftest.py
YLlllllllllll: Fixed in `9ca8d9a4`: removed the `tests/trainer/conftest.py` global `find_spec` monkeypatch. The `torch_npu` hiding is now limited to this channel-loss test module's import window and restored immediately after imports, so it no longer changes behavior for all `tests/trainer` tests.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,969 @@
...
+            if start_indices.numel() == 0:
+                return []
+            end_indices = torch.cat([start_indices[1:] - 1, torch.tensor([seq_len - 1])])
+        else:
+            start_indices = torch.tensor([0])
+            end_indices = torch.tensor([seq_len - 1])
+
+        segments = [(int(start), int(end)) for start, end in zip(start_indices.tolist(), end_indices.tolist())]
+        if len(segments) > len(source_ids):
+            segments = _filter_padding_only_segments(
```
FoolPlayer: This filters all padding-only segments whenever the raw segment count exceeds the number of sources. A real packed sample can also be a zero-supervision segment (for example a one-token sample after causal shifting), with all labels ignored and `position_ids == 0`; if tail padding is present, this can remove the real segment as well and make the remaining count too small, causing the whole micro-batch metrics to be skipped. Consider filtering only surplus tail segments beyond `len(source_ids)` instead of all padding-looking segments.

FoolPlayer (commented):
Follow-up nit on channel-loss segment filtering.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `16d19dc1`: filtering now only removes surplus padding-looking segments from the tail until the raw segment count matches `len(source_ids)`. It no longer removes padding-looking segments from the middle/front, so real zero-supervision packed samples are preserved. I applied the same tail-only policy to SP reduction and added normal/SP regressions for a one-token zero-supervision segment followed by tail padding. Local `tests/trainer/test_channel_loss_callback.py -q` now passes with 17 tests, and `make quality` passes via `uv run`.

veomni/trainer/base.py
```diff
@@ -672,21 +678,30 @@ def postforward(
...
```
Luosuu, line 685: These hooks only cover `BaseTrainer.forward_backward_step`, but DPO and DiT override it while still enabling this callback through `BaseTrainer.on_*`. They therefore skip metadata stripping and micro-step state updates; DPO may forward source metadata as unexpected model kwargs. Please move this lifecycle around the common train-loop invocation, update every override, or explicitly reject unsupported trainers.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,999 @@
...
+                        vocab_size=call_args.get("vocab_size"),
+                        hidden_states=call_args.get("hidden_states"),
+                        weights=call_args.get("weights"),
+                        shift_labels=call_args.get("shift_labels"),
+                        ignore_index=ignore_index,
+                    )
+                return result
+
+            _channel_aware_kernel._veomni_channel_loss_wrapped = True  # type: ignore[attr-defined]
+            slot._kernel = _channel_aware_kernel
```
Luosuu: `slot` is module-global, so replacing `_kernel` affects every model instance using this modeling module. In DPO, policy and reference models share the slot, so reference forwards can also overwrite this callback's side-channel state. Please scope interception to the active target model/forward, or use ownership/ref-counted dispatch instead of a closure-bound global mutation.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,999 @@
...
+        hidden_states: torch.Tensor,
+        weights: torch.Tensor,
+        labels_flat: torch.Tensor,
+        ignore_index: int,
+    ) -> torch.Tensor:
+        seq_len = hidden_states.size(0)
+        result = torch.empty(seq_len, device=hidden_states.device, dtype=torch.float32)
+        for start in range(0, seq_len, cls.EAGER_CHUNK_SIZE):
+            end = min(start + cls.EAGER_CHUNK_SIZE, seq_len)
+            chunk_logits = F.linear(hidden_states[start:end], weights).float()
```
Luosuu, line 507: On the fused-loss path this recomputes the full LM-head projection after the main kernel already did that work. Chunking limits peak memory but not FLOPs, so enabling this every step may substantially reduce throughput. Please add a logging interval/sampling option and provide a step-time benchmark; ideally expose per-token aggregates from the fused loss instead.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,999 @@
...
+            dp_group = ps.dp_group
+        except Exception:
+            dp_size = 1
+            dp_group = None
+
+        if dp_size <= 1 or not dist.is_available() or not dist.is_initialized():
+            return list(local_data)
+
+        gathered: List[Optional[List[Dict[str, Any]]]] = [None] * dp_size
+        dist.all_gather_object(gathered, local_data, group=dp_group)
```
Luosuu: This gathers one Python dict per packed segment after per-segment `.item()` calls, and it does not gather the rank-local `source_id -> source_name` mapping. Please aggregate locally by channel first, reduce compact `(loss_sum, token_count)` tensors/dicts, and synchronize names with a consistency check before constructing metric keys.

veomni/trainer/base.py
YLlllllllllll: Fixed in `1bf7aabd`. The channel-loss micro-step lifecycle now wraps the actual forward/backward invocation, and metadata is stripped only after `preforward()` so packing logic still sees it. `TextDPOTrainer` mirrors that lifecycle: the reference forward is outside capture, the policy forward is inside capture, and each preference-pair source is explicitly expanded to its chosen/rejected segments. DiT, classification, and BaseRL paths are rejected explicitly where their objective/packing lifecycle is unsupported. VLM delegate wrappers (native LoRA, SeedOmni foundation, Qwen2.5-Omni thinker, and Qwen3-Omni thinker) are resolved to the module that actually computes loss; the legacy SeedOmni Qwen3-MoE foundation is rejected because it bypasses observable loss dispatch. Regression coverage includes metadata stripping order, DPO policy-only capture with two pairs, delegate models, and unsupported-trainer boundaries.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `1bf7aabd`. Module-global loss OpSlots now use one shared `_OpSlotDispatcher` per slot, a weak owner set, and a forward-scoped `ContextVar` selecting the active `ChannelLossComputer`. The original kernel is always called unchanged; only the active owner receives side-channel inputs. Uninstall removes one owner and restores the original kernel only after the final owner is gone. DPO keeps reference-model forward outside the capture context, so it cannot mutate policy metrics even when policy/reference share the same modeling module and OpSlot. Tests cover no-capture transparency, two owners, owner isolation, and teardown in either order.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `1bf7aabd`. Added `train.channel_loss.interval` with default `10`; metadata stripping remains active every step, while detached CE/LM-head recomputation and logging only run when `global_step % interval == 0` (`interval=1` preserves every-step behavior). Historical single-variable H20 2x8 runs measured the worst-case `interval=1` overhead over steps 10-50: CE-only job [9d7308ab0293ac4e](https://ml.bytedance.net/development/instance/jobs/9d7308ab0293ac4e?tabState=run_info&trialId=343626113) mean/median step time 1.2888s/1.1364s; channel-loss job [e5c46306a3eb8d13](https://ml.bytedance.net/development/instance/jobs/e5c46306a3eb8d13?tabState=run_info&trialId=343577767) 1.3129s/1.1619s, i.e. +1.87% mean and +2.24% median. Mean throughput was 0.026799 vs 0.026550 Mtok/s. The default interval=10 should amortize that to roughly 0.2%; that last number is a linear estimate, not a new measurement. Config validation, sampled/unsampled lifecycle tests, and fused hidden-state chunking tests are included.

veomni/trainer/callbacks/channel_loss_callback.py
YLlllllllllll: Fixed in `1bf7aabd`. Normal-path segment results are accumulated locally on device into one `(loss_sum, token_count)` pair per source, with no per-segment `.item()` calls. DP gathers only the compact current `source_id -> source_name` registry, validates name consistency, builds aligned float32 loss-sum/int64 token-count tensors, and all-reduces those tensors. Metric keys are source-qualified from their first emission, and the globally synchronized source registry is checkpointed/restored so naming remains stable after resume. SP still uses object gather only for the ordered partial-segment reconstruction that cannot be represented as a source-total reduction before continuations are joined; after reconstruction it also returns compact per-source tensors. Tests cover compact DP reduction, cross-rank name mismatches, persistent registry/resume, B>1 SP ordering, per-row padding, zero-supervision segments, and ambiguous padding rejection.

YLlllllllllll:
Updated the PR to `1bf7aabd` and replied to all four Luosuu review threads. The final patch also covers DPO pair metadata, shared OpSlot ownership, Liger/chunk/LoRA/VLM delegates, compact DP reduction, B>1 SP/non-SP padding ambiguity, stable source-qualified metric keys, and checkpointed source registry. Local verification: trainer+DPO 54 passed, channel+checkpoint 56 passed, `make quality` passed, and an independent VeOmni review gate returned `safe`. The fork workflows currently show `action_required` (maintainer approval needed to run), not test failures. @Luosuu please take another look when convenient.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -0,0 +1,1444 @@
...
+            flat_start = batch_idx * local_width + start
+            flat_end = min(batch_idx * local_width + end, seq_len)
+            labels_slice = labels_flat[flat_start:flat_end]
+            loss_slice = per_token_loss[flat_start:flat_end]
+            pos_slice = positions_2d[batch_idx, start:end].reshape(-1)[: labels_slice.numel()]
+            mask_slice = None
+            if attention_mask_flat is not None:
+                mask_slice = attention_mask_flat[flat_start:flat_end]
+            valid = labels_slice != ignore_index
+            token_count = valid.sum().item()
```
Luosuu: [P2] Calling `.item()` for both token count and loss sum synchronizes the accelerator once or twice per packed segment, so sampled SP steps incur O(segment-count) host/device stalls. Please keep these values as tensors and transfer or reduce them in one compact batch after local aggregation.

YLlllllllllll:
NPU CI triage update:

- `NPU e2e tests` passed. The only failing check is `NPU unit tests`, in `tests/checkpoints/test_trainer_saveload.py::test_trainer_saveload_hsdp[4]`.
- The failure is the exact HF checkpoint equality check for `lm_head.weight`, with `max diff: 1.9073486328125e-06`; the other 7 save/load cases passed.
- The same test failed with the same tensor and exact max diff on `main` at commit `bf0756ef`: https://github.com/ByteDance-Seed/VeOmni/actions/runs/29068458978. This is an existing async HSDP checkpoint comparison flake, not a channel-loss-path regression.
- I attempted to rerun failed jobs through the API, but the fork token is denied with `Must have admin rights to Repository`.

Could a maintainer please rerun the failed NPU unit job? https://github.com/ByteDance-Seed/VeOmni/actions/runs/29083545328

Luosuu (approved):

veomni/trainer/callbacks/channel_loss_callback.py
Luosuu: @YLlllllllllll please address this then I think we can merge

YLlllllllllll: Addressed in `4ba977ff`.

- Removed all per-segment `.item()` calls from the SP aggregation path. Loss sums, token counts, and padding evidence remain tensors.
- SP now gathers structure-only metadata separately, then performs a constant two dtype-preserving compact tensor gathers per sampled micro-batch (`float32` losses and exact `int64` counts/flags), independent of segment count.
- `position_ids` are copied to host once for segment ordering; only compact per-segment zero-position flags are transferred back to the stats device.
- Added regressions for no scalar extraction, B>1 rank ordering, empty local rank, unequal per-rank segment counts, and surplus remote padding filtering.

Verification: channel-loss `49 passed`; related channel/checkpoint/DPO `64 passed`; real 2-rank Gloo uneven collective smoke passed; `make quality` passed; independent VeOmni review gate returned `safe`. GPU/NPU CI has been retriggered on the new head.

YLlllllllllll:
`4ba977ff` is pushed and the SP synchronization review thread is addressed. The new-head GPU/NPU and quality workflows were created but are all currently `action_required` pending upstream approval. @Luosuu could you please approve this run set when convenient?

YLlllllllllll:
The latest `NPU unit tests` job was cancelled before any tests started: `uv sync` stalled while downloading `torch` until the 90-minute job timeout. The parallel NPU e2e job completed the same dependency sync and passed. @Luosuu @FoolPlayer, could you please re-run only this NPU unit test? https://github.com/ByteDance-Seed/VeOmni/actions/runs/29146253172/job/86563262997

PR #893, opened 2026-07-08
[BREAKING][dist, trainer] feat: per-module local parallel state via context manager

Coach257:
## What does this PR do?

Introduces **local parallel state**: distributed process groups (SP/DP/CP/EP)
resolve from the *current* `ParallelState` instead of module-level globals, and a
`use_parallel_state(ps)` context manager scopes which state is "current". This lets
different models/modules build and run under different parallel topologies — the
foundation for multi-module (omni) training.

### Core changes

- **`parallel_state.py`**
  - `init_parallel_state()` now **returns** the built `ParallelState` and caches it
    by topology (`_PARALLEL_STATE_CACHE`).
  - New `set_parallel_state(ps)` and `use_parallel_state(ps)` (contextmanager) to
    swap the current global state.
  - SP-related groups moved onto `ParallelState` (resolved from its `DeviceMesh`)
    rather than being set as global side effects in `__post_init__`.
  - Meshless SP is no longer supported: `__post_init__` raises if
    `sp_enabled and device_mesh is None`. Always build via `init_parallel_state`.

- **`sequence_parallel/comm.py`** — **removed all group globals**. Getters
  (`get_ulysses_sequence_parallel_group`, `get_unified_sequence_parallel_group`,
  `get_context_parallel_group`, `get_data_parallel_group`) now resolve from the
  current `ParallelState`'s mesh (exactly how `fsdp_group` already worked), and
  return `None` when uninitialized. `set_ulysses_sequence_parallel_group` survives
  ONLY as a unit-test injection seam (`_ULYSSES_SP_GROUP_OVERRIDE`, `None` in prod).

- **`sequence_parallel/{data,loss}.py`** — additive per-module SP primitives;
  `reduce_sequence_parallel_loss` takes an optional `group=` (defaults to current
  state's unified SP group).

- **Trainers** — each trainer scopes its `ParallelState` in exactly **two regions**:
  1. **Build region**: one `with use_parallel_state(self.parallel_state):` around
     the whole build sequence in `__init__` (right after `_setup()`) — covers
     meta-init, FSDP2/EP wrap + weight load, optimizer grouping, and SP data pipeline.
  2. **Run region**: `forward_backward_step` (forward + loss-reduce + backward).

  Applied to `BaseTrainer`, `TextTrainer`, `VLMTrainer`, `DiTTrainer`, `TextDPOTrainer`
  (the latter's build region also covers the reference model). This matters because
  `init_parallel_state()` only sets the module global on its *first* call, so the
  explicit scope keeps each module on its own mesh once multiple modules build
  separately (a no-op for today's single-model case).

## Breaking changes

- `comm.py` group-setter globals (`set_data_parallel_group`,
  `set_context_parallel_group`, `set_unified_sequence_parallel_group`, and the
  module-level `_*_GROUP` variables) are removed. Groups now come from the current
  `ParallelState`. External code that called these setters or read the globals must
  switch to `init_parallel_state()` + `use_parallel_state()`.
- Meshless sequence-parallel `ParallelState()` construction now raises.


Luosuu:
let's discuss design of this PR

veomni/trainer/base.py
```diff
@@ -827,7 +815,11 @@ def train(self):
 
             for _ in range(self.start_step, args.train_steps):
                 try:
-                    self.train_step(self.data_iterator)
+                    # Scope the whole optimize step (fwd/bwd, grad clip, optimizer)
+                    # to this model's parallel state so every group getter resolves
+                    # from its own device mesh.
+                    with use_parallel_state(self.parallel_state):
```
Luosuu: This scope is too far outside the operation that requires it. A multi-module orchestrator can call `forward_backward_step()` directly and bypass `train()`, leaving the first/ambient module's PS active. A local probe against this head returned `direct_call_used_trainer_state=False`.

Please make the public execution boundaries self-scoping: at minimum `BaseTrainer.forward_backward_step()` plus the DiT and DPO overrides, while keeping the optimizer/grad-clip portion of `train_step()` under the same state (a shared wrapper/decorator would avoid drift). Add a regression test with two distinct states and direct calls, including exception restoration.

veomni/distributed/parallel_state.py
```diff
@@ -493,11 +445,49 @@ def init_parallel_state(
...
+        ulysses_size,
+        dp_mode,
+        device_type,
+        include_sp_in_fsdp,
+        extra_parallel_sizes,
+        extra_parallel_placement_innermost,
+        extra_parallel_names,
+        async_enabled,
+    )
+    cached_state = _PARALLEL_STATE_CACHE.get(cache_key)
```
Luosuu, line 518: This cache is not tied to the lifetime of the default process group. After `destroy_process_group()`, clearing `_PARALLEL_STATE`, and initializing a new Gloo group with the same topology, this lookup returned the old state; accessing `p2.dp_group` then failed with `RuntimeError: Could not resolve the process group registered under the name ...`.

Please scope/invalidate the cache per distributed session, or clear both `_PARALLEL_STATE` and `_PARALLEL_STATE_CACHE` during teardown. A destroy/reinitialize-same-topology regression test should cover this lifecycle.

Luosuu (changes requested):
Requesting changes for two correctness issues in the ParallelState ownership/lifecycle model:

1. The runtime scope currently lives in the outer `train()` loop, so direct calls to `forward_backward_step()`—the expected integration point for a multi-module orchestrator—can run under another module's ambient state. This affects `BaseTrainer` and the DiT/DPO overrides.
2. `_PARALLEL_STATE_CACHE` is keyed only by topology and can return `DeviceMesh` objects backed by process groups from a destroyed distributed session.

Both issues were reproduced locally: a direct-call probe observed the wrong state (`direct_call_used_trainer_state=False`), and a destroy/reinitialize Gloo probe reused the cached state then failed to resolve its process group.

Suggested acceptance tests:
- Build two heterogeneous states, make direct `forward_backward_step()` calls for each trainer override, and assert every group getter observes the callee's state with restoration after success/exception.
- Initialize a state, destroy and recreate the default process group with the same topology, and verify no stale cached group is reused.

Local validation otherwise looks healthy: `make quality` and `git diff --check` passed; 114 relevant tests passed, including all 48 selected data/torchrun tests. GPU-only coverage was limited by local hardware.

veomni/trainer/base.py
Coach257: The scope has now been set to a finer granularity.

veomni/distributed/parallel_state.py
Coach257: `clear_parallel_state` added

Coach257: We now use `_PARALLEL_STATE_REGISTRY`

Luosuu (approved):

PR #711, opened 2026-04-29
[ops, trainer] feat: chunked per-token log-probs for PPO-style RL

Luosuu:
## Summary

Adds a memory-efficient chunked linear cross-entropy that returns the **unreduced per-token NLL** — the building block PPO-style RL needs to compute policy and reference log-probs without materializing the `[B, L, V]` logits tensor.

After this change, both text and VLM RL training can run with long context + large vocab (e.g. Qwen3 V≈150k, L≈8k) without OOM. A single `model(..., return_log_probs=True)` call activates the path uniformly across every patched text and VLM model in-tree.

### What's in scope

- **Kernel**: `veomni/ops/kernels/cross_entropy/chunk_logprobs.py` — `chunk_logprobs_function(hidden_states, weights, labels, chunk_size, ignore_index, ...)` returns `[B, L]` per-token log-probs. Uses an explicit `torch.autograd.Function` with chunked forward + backward; routes through `flash_attn.ops.triton.cross_entropy.cross_entropy_loss` so it is **bitwise identical** to verl's `FusedLinearForPPOFunction`. Memory profile matches `chunk_loss` (peak `[chunk_size, V]` fp32 logits per chunk).
- **Centralized 3-tuple dispatch**: `ForCausalLMLoss` now returns `(loss, logits, log_probs)`. When `return_log_probs=True` is in `**kwargs`, the wrapper routes to `chunk_logprobs_function` and returns `(None, None, log_probs)`; otherwise it returns `(loss, logits, None)`. `_chunk_loss_dispatch` shim wraps `chunk_loss_function` so the OpSlot/`LOSS_MAPPING` install path obeys the same contract. `ForSequenceClassificationLoss` returns `(loss, logits, None)`.
- **Per-model code reduces from ~22 lines → 1 line.** Every model `forward()` is now `loss, logits, log_probs = self.loss_function(...); outputs.log_probs = log_probs`. No per-model early-exit, no per-model log-probs imports. Touches 11 hand-written modelings (qwen2/qwen2_vl/qwen2.5_vl/qwen3/qwen3_moe/qwen3_vl/qwen3_vl_moe/qwen3_omni_moe/llama/deepseek_v3/seed_oss/qwen2_5_omni) and 11 patchgen configs (incl. GPU+NPU twins for glm_moe_dsa) — `generated/*` regenerated via `make patchgen`.
[...]

PR #894, opened 2026-07-08
[trainer] fix: pass labels to model in DPO concatenated_forward

Winter-Dry:
Since #711 rewrote concatenated_forward to use the chunked log-probs kernel via return_log_probs=True, the model's fused-linear path is guarded by `if labels is not None:` — but _NON_MODEL_KEYS still stripped labels from model_inputs (a pre-#711 leftover to avoid the HF loss branch on the gather-on-logits path). Result: fused_linear_aux stays None and `outputs.fused_linear_aux.log_probs` raises AttributeError.

Empty the set so labels are forwarded; the chunk_logprobs kernel needs them for shift + ignore_index masking anyway.

### DIFF
--- a/veomni/trainer/text_dpo_trainer.py
+++ b/veomni/trainer/text_dpo_trainer.py
@@ -38,7 +38,7 @@ from .base import BaseTrainer, VeOmniIter
 
 logger = logging.get_logger(__name__)
 
-_NON_MODEL_KEYS = {"labels"}
+_NON_MODEL_KEYS = set()
 
 
 # ================================ DPO Arguments ======================================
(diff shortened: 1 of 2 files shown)


Coach257 (approved):

veomni/trainer/text_dpo_trainer.py
```diff
@@ -38,7 +38,7 @@
 
 logger = logging.get_logger(__name__)
 
-_NON_MODEL_KEYS = {"labels"}
+_NON_MODEL_KEYS = set()
```
FoolPlayer, line 41: Passing labels fixes return_log_probs, but under SP these labels are already globally shifted. For packed DPO segments, the chosen tail can target the rejected head. Please use segment-aware shifted labels or mask segment boundaries, and add an SP regression test.

FoolPlayer (approved):

PR #915, opened 2026-07-14
[dist, lora] feat: MoE ep_sharded_stream_load for PEFT models + NPU fused MoE-LoRA EP

Coach257:
## Summary

Extends the per-rank EP-slice weight loader (`ep_sharded_stream_load`) to **LoRA-wrapped (PEFT) MoE models**, and closes the last gap that kept MoE-LoRA + expert parallelism GPU-only by adding an **NPU fused MoE-LoRA kernel**. Net effect: an EP-sharded MoE model — with or without LoRA, on GPU or NPU — can be loaded via any of the three loaders (plain / rank-0 broadcast / `ep_sharded_stream_load`) and trained/inferred with EP.

### 1. PEFT support for `ep_sharded_stream_load` (`veomni/models/module_utils.py`)
- **Base experts**: fused `...experts.gate_up_proj` / `down_proj` tensors stream from the base checkpoint straight into the wrapped `...base_layer.weight` destinations; each rank reads only its `[E/ep, …]` dim-0 slice.
- **Adapter** (`_stream_lora_adapter_ep_sharded`, reads `adapter_model.safetensors`):
  - *independent* MoE-LoRA (EP-sharded in the runtime plan) → each rank reads its `[E/ep, r, …]` slice;
  - *shared* / dense LoRA → read whole by every rank (FSDP then shards).
- **Safety guard**: if a model's `CheckpointTensorConverter` needs a non-dim0-zero-pad transform (e.g. Qwen3-MoE's per-expert→fused fusion), streaming raises a clear `NotImplementedError` instead of silently corrupting weights, pointing the user at a fused checkpoint or the whole-tensor loader.

### 2. NPU fused MoE-LoRA EP kernel (`veomni/lora/ops/npu_moe_group_gemm.py`, new)
- Adds `npu_fused_lora_moe_forward` (shared / Mode 2) and `npu_fused_independent_lora_moe_forward` (independent / Mode 1), drop-in matching the Triton kernels' signature.
- MoE-LoRA and EP are orthogonal: the kernel **reuses the base NPU MoE EP primitives** (`dispatch_preprocess` / `alltoall_dispatch` / `alltoall_combine` + the `npu_group_gemm` autograd `Function`) and only injects the three seed-style LoRA deltas (gate/up pre-SwiGLU, down post-mid). Backward is derived by autograd — no hand-written kernel. Math mirrors `moe_layers._eager_forward` line-for-line.
- `bind_lora_moe_kernels("npu")` now binds these (previously cleared the pointers → eager fallback, which can't do EP). Both backends' LoRA kernels now live side-by-side under `veomni/lora/ops/` (`moe_group_gemm.py` = Triton, `npu_moe_group_gemm.py` = NPU).

### 3. Supporting refactors
- `make_peft_key_mapper` centralized in `veomni/lora/weight_loading.py` (removes duplicate PEFT-key remap logic across the three loaders).
- `get_runtime_parallel_plan` hardened to return `None` gracefully when a model has no `get_parallel_plan`, collapsing redundant guards in `module_utils.py`.
- `_reset_moe_wrapper` fix so LoRA params reset to kaiming-A / zero-B correctly (`_lora_fresh_reset_done` flag).
- LoRA adapters now saved as **`adapter_model.safetensors`** (`save_safetensor_utils.py`), required for per-rank slicing.

### 4. Tests + CI
- `tests/utils/test_moe_ep_sharded_load_matrix.py` (new): device-agnostic non-PEFT matrix — a fake MoE model × {merged, per-expert} checkpoints × {plain, rank-0 broadcast, `ep_sharded`} loaders, including the `NotImplementedError` bail case.
- `tests/lora/test_moe_lora_ep_sharded_stream_load.py` (new): trainer-driven PEFT test asserting streamed LoRA weights are **bit-identical** to the broadcast loader, for both independent and shared LoRA. Auto-selects `fused_triton` (GPU) / `fused_npu` (NPU).
- Wired into `gpu_unit_tests.yml` (both) and `npu_unit_tests.yml` (both — the PEFT test now validates the NPU kernel forward + the loader on real Ascend hardware).

### DIFF
--- a/.github/workflows/gpu_unit_tests.yml
+++ b/.github/workflows/gpu_unit_tests.yml
@@ -148,6 +148,11 @@ jobs:
       - name: Run fsdp2 all ranks or rank0 load model weights
         run: |
           uv run --frozen pytest -s -x tests/utils/test_rank0_load_and_broadcast_weights.py
+      - name: Run MoE ep_sharded_stream_load matrix (non-PEFT)
+        # Fake EP MoE model: {plain, rank0 broadcast, ep_sharded} x {merged,
+        # per-expert} bit-identical load + the per-expert->fused streaming bail.
+        run: |
+          uv run --frozen pytest -s -x tests/utils/test_moe_ep_sharded_load_matrix.py
       - name: Run checkpoint callback unit tests
         run: |
           uv run --frozen pytest -s -x tests/checkpoints/test_checkpoint_callback.py
@@ -184,6 +189,8 @@ jobs:
         #     bit-exact LoRA reload, also covers linear LoRA on q_proj/v_proj since the
         #     yaml enables both lora_modules and target_parameters)
         #   - moe_lora_ep2: EP=1 vs EP=2 trainer-driven integration + per-step loss/grad_norm alignment
+        #   - moe_lora_ep_sharded_stream_load: PEFT ep_sharded_stream_load vs broadcast, bit-exact
+        #     post-load LoRA weights (independent + shared) on a fused-base Qwen3-MoE at EP=2
         # Order: cheap unit tests first to fail fast, trainer/ep2 last (multi-subprocess, ~5-10 min each).
         run: |
           uv run --frozen pytest -s -x tests/lora/test_target_mapping.py
@@ -192,3 +199,4 @@ jobs:
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_fused.py
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_trainer.py
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_ep2.py
+          uv run --frozen pytest -s -x tests/lora/test_moe_lora_ep_sharded_stream_load.py
--- a/.github/workflows/gpu_unit_tests.yml
+++ b/.github/workflows/gpu_unit_tests.yml
@@ -183,6 +183,8 @@ jobs:
         # `uv sync --frozen --extra gpu --group dev` step is sufficient — no extra sync needed here.
         # Coverage:
         #   - lora_trainer_saveload: dense Qwen3 + linear LoRA save/load smoke test
+        #   - veomni_lora_moe_native: CPU MoE-LoRA injection/save-load + init_lora_parameter
+        #     preserve/partial/fresh reset contract (catches to_empty clobber regressions)
         #   - moe_lora_eager: wrapper layout + autograd parity (Mode 1 + Mode 2, all v5 MoE configs)
         #   - moe_lora_fused: triton kernel parity vs eager (forward + backward, EP autograd classes)
         #   - moe_lora_trainer: production save/load/resume round-trip (DCP shard + HF adapter,
@@ -194,6 +196,7 @@ jobs:
         # Order: cheap unit tests first to fail fast, trainer/ep2 last (multi-subprocess, ~5-10 min each).
         run: |
           uv run --frozen pytest -s -x tests/lora/test_target_mapping.py
+          uv run --frozen pytest -s -x tests/lora/test_veomni_lora_moe_native.py
           uv run --frozen pytest -s -x tests/lora/test_lora_trainer_saveload.py
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_eager.py
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_fused.py
--- a/tests/lora/test_moe_lora_ep_sharded_stream_load.py
+++ b/tests/lora/test_moe_lora_ep_sharded_stream_load.py
@@ -1,0 +1,302 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+r"""Trainer-driven EP=2 test for the **PEFT** ``ep_sharded_stream_load`` path.
+
+Companion to ``tests/utils/test_moe_ep_sharded_load_matrix.py`` (which
+covers the *non-PEFT* MoE load matrix with a fake model). This one
+exercises the real veomni LoRA stack on the Qwen3-MoE toy and asserts
+that streaming a LoRA-wrapped MoE checkpoint reconstructs **bit-identical**
+weights to the reference broadcast loader.
+
+What ``ep_sharded_stream_load`` does for a PEFT model
+-----------------------------------------------------
+``load_model_weights_ep_sharded`` under ``is_peft_model=True`` (wired in
+``veomni/models/module_utils.py``):
+
+  * **Base experts** — the fused ``...experts.gate_up_proj`` /
+    ``down_proj`` tensors are streamed from the base checkpoint straight
+    into the wrapped ``...experts.<spec>.base_layer.weight`` destinations;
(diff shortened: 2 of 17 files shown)


Luosuu (commented):
Thanks for tackling the PEFT + EP streaming path. The overall direction looks useful and should become the single implementation that downstream GPT-OSS work builds on. I found two correctness cases that should be resolved before merge, plus a few changes that would make this large diff easier to audit and maintain. In particular, please keep adapter loading fail-closed, distinguish fully fresh wrappers from partially loaded wrappers, and run the existing initialization regression test in CI. For readability, it would also help to consolidate the duplicated dim-0 slicing logic and organize/squash the history into logical units (loader/runtime plan, NPU kernel, tests/CI); alternatively, the NPU kernel could be a follow-up PR if it is independently mergeable.

veomni/models/module_utils.py
```diff
@@ -645,29 +690,152 @@ def load_model_weights_ep_sharded(
...
+    streamed per-rank). Newly saved adapters are safetensors (see
+    :func:`save_lora_adapter_with_dcp`).
+    """
+    from ..lora.state_dict import _find_adapter_file, insert_adapter_name
+
+    try:
+        file_path, is_safetensors = _find_adapter_file(adapter_path)
+    except FileNotFoundError:
+        logger.warning_rank0(f"ep_sharded adapter: no adapter file under {adapter_path}; skipping.")
+        return
```
Luosuu: An explicitly configured `adapter_path` should fail closed here. As written, a typo or incomplete checkpoint directory is converted into a warning, then `post_process_after_weight_loading` initializes a fresh random adapter and training proceeds as if resume succeeded. The other adapter loaders propagate this `FileNotFoundError`, so the EP-streaming path should preserve that contract too.

```suggestion
    file_path, is_safetensors = _find_adapter_file(adapter_path)
```

veomni/lora/weight_loading.py
```diff
@@ -227,6 +252,19 @@ def init_lora_parameter(model: nn.Module, name: str) -> None:
...
+    guard is dead under the standard loaders, which run ``model.to_empty(...)``
+    *before* init — after that no param is on meta, so the guard silently skipped
+    the reset and left the wrapper's ``lora_A`` / ``lora_B`` as the uninitialised
+    (garbage / NaN) storage ``to_empty`` produced. We reset once per wrapper
+    (idempotent); when an adapter *is* loaded, its LoRA tensors are removed from
+    the un-loaded set so this path is never taken for them (loaded weights are
+    never clobbered)."""
+    if getattr(module, "_lora_fresh_reset_done", False):
+        return
+    module.reset_lora_parameters(init_lora_weights=True)
```
Luosuu: The once-only marker fixes fresh `to_empty()` initialization, but it can clobber a partially loaded wrapper: if any LoRA key remains in `parameter_names_left`, this resets every A/B tensor in the wrapper, including tensors already loaded from the adapter. I reproduced this on the PR head with `pytest -q tests/lora/test_veomni_lora_moe_native.py -k preserves_loaded_weights`; both shared and independent cases fail. Please make the decision at wrapper granularity using the missing-name set: reset once only when *all* LoRA tensors for that wrapper are missing; do nothing when none are missing; and for a subset either raise a clear missing-key error (preferred for an adapter resume) or initialize only the exact missing tensors without touching loaded values. That handles fresh meta builds without weakening the existing preservation invariant.

.github/workflows/gpu_unit_tests.yml
```diff
@@ -192,3 +199,4 @@ jobs:
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_fused.py
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_trainer.py
           uv run --frozen pytest -s -x tests/lora/test_moe_lora_ep2.py
+          uv run --frozen pytest -s -x tests/lora/test_moe_lora_ep_sharded_stream_load.py
```
Luosuu, line 205: Please run `tests/lora/test_veomni_lora_moe_native.py` before the expensive trainer integrations. The existing `test_init_lora_parameter_preserves_loaded_weights` regression test fails in both modes on this PR head, while the newly selected files all pass, which is why the current green CI misses the initialization regression.

veomni/models/module_utils.py
```diff
@@ -645,29 +690,152 @@ def load_model_weights_ep_sharded(
...
+                real0 = sl.get_shape()[0]
+                expected_full0 = target0 * para_size
+                if real0 != expected_full0:
+                    raise RuntimeError(
+                        f"ep_sharded adapter {name}: file dim0={real0} != model dim0={expected_full0} "
+                        f"(independent LoRA expects the gathered [E, ...] tensor)."
+                    )
+                start = para_rank * target0
+                tensor = sl[start : start + target0]
+                _dispatch_parameter(model, name, tensor, dtensor_factory, None, dtensor_to_cpu)
```
Luosuu, line 851: This repeats the base-checkpoint path's group lookup, EP size/rank calculation, expected-full-dim validation, slice-bound calculation, and already-local dispatch convention. Could we extract a small helper for dim-0 EP slice metadata/validation and use it in both paths (with strict vs zero-pad behavior as an argument)? Keeping this arithmetic in one place would make the diff much easier to audit and prevent the base and adapter loaders from drifting on edge cases.

tests/lora/test_moe_lora_ep_sharded_stream_load.py
```diff
@@ -0,0 +1,302 @@
...
+        ],
+        nproc=nproc,
+    )
+
+    # ── 4. Post-load LoRA weights must be bit-identical ─────────────────
+    _compare_snapshots_bit_exact(
+        actual_path=os.path.join(stream_dir, SNAPSHOT_PRE),
+        ref_path=os.path.join(ref_dir, SNAPSHOT_PRE),
+        label=f"{mode}/ep_sharded_stream_vs_broadcast",
+    )
```
Luosuu, line 293: This is strong loader coverage, but on NPU it only compares the pre-training adapter snapshot; the subsequent step proves the new fused kernel executes, not that its forward or gradients match the reference. Since this PR adds a new autograd composition, please add a small NPU-gated numerical parity test against eager for outputs and gradients (hidden states plus A/B weights), covering shared and independent LoRA and ideally EP/non-EP. That gives the NPU portion a correctness test comparable to the existing Triton fused-LoRA coverage.

.github/workflows/gpu_unit_tests.yml
Coach257: added

veomni/models/module_utils.py
Coach257: helper `_read_ep_dim0_slice` added

tests/lora/test_moe_lora_ep_sharded_stream_load.py
Coach257: `test_moe_lora_fused_npu` added

tests/lora/test_moe_lora_fused_npu.py
```diff
@@ -0,0 +1,325 @@
...
+    """Gradients on <spec>.lora_A / <spec>.lora_B match between NPU fused and eager."""
+
+    def _grads(*, fused: bool):
+        model, fqn, exp, _ = _build_wrapped(mode=mode, fused=fused, lora_b_perturb_std=0.02)
+        wrapper = model.get_submodule(fqn)
+        wrapper.train()
+        h, idx, w = _make_inputs(exp)
+        loss = wrapper(h, idx, w).float().pow(2).sum()
+        loss.backward()
+        return {n: p.grad.detach().clone() for n, p in wrapper.named_parameters() if p.grad is not None}
```
Luosuu: Thanks for adding the parity suite. One part of the original request is still missing: hidden-state gradient parity. The current helper creates h without requires_grad and returns only parameter gradients, so a bug in GmmFunction.backward grad_input or in the all-to-all backward could pass while every A/B assertion remains green. Please make h a leaf requiring grad and compare h.grad in fused-vs-eager; likewise include a per-run hidden-state leaf in the EP-vs-non-EP autograd.grad inputs.

Luosuu (approved):

PR #917, opened 2026-07-15
[model, ops] feat: align DeepSeek V4 inference and add FP8 activation QAT

Luosuu:
> [!IMPORTANT]
> This integration draft is superseded for review by the stack tracked in #927: #928 → #929 → #931. Keep this PR as the end-to-end investigation and CI baseline; merge the child PRs instead.

## Summary

- keep DeepSeek V4 stateless prefill and training cache-free so the TileLang Lightning Indexer dispatches
- add official-compatible FP8/FP4 kernels and BF16-resident on-demand quantization diagnostics
- match official RMSNorm, routing, and shared-expert FP32 clamped SwiGLU semantics
- add an opt-in DeepSeek-V4 FP8 activation QAT path while retaining BF16 master weights
- add official/VeOmni tracing, terminal-head isolation, routing replay, and corrected causal logprob comparison

## Root causes and findings

1. `DeepseekV4Model.forward` created a `DynamicCache` even with `use_cache=False`, forcing stateless prefill/training off the TileLang indexer path.
2. The parity harness pre-shifted labels even though `chunk_logprobs_function` shifts internally. Earlier 12–16 logprob differences compared different target tokens and are invalid.
3. `DeepseekV4MLP`, used by shared experts, omitted official `swiglu_limit=10` FP32 clamp/SwiGLU arithmetic. This is now patched through patchgen and regression-tested.
4. The LM head is effectively exact (mean target-logprob diff `1.32e-7` with identical terminal hidden states). Remaining practical divergence is in the model body, primarily native MoE routing.

## Corrected parity results

At 256 tokens:

- native BF16: mean `0.06019`, p99 `1.08950`, max `1.39349`
- native full quant simulation: mean `0.08296`, p99 `1.51534`, max `3.14326`
- BF16 + official expert IDs/router weights: **mean `0.01774`, p99 `0.27628`, max `0.58198`**
- full quant + official routing: mean `0.02205`, p99 `0.36378`, max `0.97750`

Fresh 4096-token validation of the lowest path:

- **mean `0.005597`, p99 `0.17216`, max `0.45672`**
- normalized terminal-hidden mean/RMS diff `0.008071 / 0.011667`

The lowest path supports keeping weights BF16-resident and quantizing selectively/on demand. Official-routing replay is diagnostic, not deployable native inference.

## FP8 activation QAT

`model.ops_implementation.deepseek_v4_fp8_activation_qat=true` enables the Miles-compatible training path:

- main-attention and compressed KV fake-quantize non-RoPE channels once, in 64-value blocks
- indexer K/Q apply a normalized Hadamard rotation and block-128 FP8 fake quantization
- backward uses an identity STE through fake quantization; Hadamard gradients remain exact
- model parameters remain BF16; source activations are not mutated
- NVIDIA SM90+ is required and Ascend NPU configuration fails explicitly
- there is no numerical Hadamard fallback: missing `fast-hadamard-transform` fails during ops configuration
- the pinned CUDA extension is installed manually after the GPU environment sync, keeping no-`nvcc` patchgen CI installable

This is a QAT capability, not a claim that fake quantization alone lowers pretrained inference parity error. Quack autotuning control is tracked separately in #922.

## Validation

- 4-GPU 256- and 4096-token official-checkpoint comparisons
- LM-head isolation against official terminal hidden states
- exact packed-FP4 weight/scale round-trip checks
- `tests/ops/test_deepseek_v4_kernels.py`: 29 passed, including QAT model forward/backward on GB200 and explicit fail-fast/no-fallback coverage
- `tests/models/test_deepseek_v4_fused_moe.py`: 7 passed
- exact `uv run --extra gpu --dev patchgen --check` workflow command passes
- generated patch output refreshed through patchgen
- `uv lock --check`, Ruff, Python compile checks, and `git diff --check` passed
- independent `/veomni-review`: safe, no blocking findings

## Remaining work

- reduce native router/body divergence without depending on official routing replay
- measure whether training with the QAT path improves a quantized deployment target; the replay result establishes a `0.0056` long-context diagnostic lower bound for the current BF16 execution path


Luosuu:
### FP8 attention-KV parity experiment

Integrated the official-compatible `act_quant` kernel into an isolated trace mode that applies E4M3 fake quantization with UE8M0 scales to the non-RoPE attention KV channels (block size 64), matching the official inference placement without also enabling FP4 indexer or quantized-linear approximations.

| Metric | Baseline | FP8 KV simulation | Delta |
| --- | ---: | ---: | ---: |
| 256-token mean log-prob abs diff | 12.5888 | 12.3201 | -0.2687 |
| 4096-token mean log-prob abs diff | 16.1474 | 16.2609 | +0.1135 |
| 4096-token p99 log-prob abs diff | 25.1250 | 25.7500 | +0.6250 |
| 4096-token MoE top-k overlap | 62.94% | 62.30% | -0.64 pp |
| 4096-token indexer top-k overlap | 92.50% | 92.48% | -0.02 pp |
| 4096-token attention top-k overlap | 97.07% | 97.06% | -0.01 pp |

Conclusion: isolated attention-KV `act_quant` does not reduce the full-length official parity gap. The kernel and diagnostic switch remain useful, but fake quantization is intentionally not enabled in the production model path. The remaining gap likely requires matching the official FP8/FP4 GEMMs and compressor FP32 accumulation semantics rather than adding KV simulation alone.

Luosuu:
Quantization parity update\n\nImplemented and validated:\n- BF16-resident routed-expert weights with opt-in on-demand FP4/FP8 quantization.\n- Official-style E4M3/UE8M0 act_quant for non-RoPE attention KV channels.\n- Fixed-checkpoint-scale FP8 weight repacking and TileLang FP8 GEMM.\n- Optional FP32 router diagnostics and mutually exclusive trace modes.\n- EP safety guards, strict scale discovery/validation, and non-multiple-of-32 GEMM padding.\n\nParity results:\n- Baseline 256 mean absolute logprob diff: 12.588826.\n- KV act_quant at 256: 12.320136 (small improvement).\n- Baseline 4096: 16.147400.\n- KV act_quant at 4096: 16.260887 (regression).\n- On-demand FP4 experts at 256: 12.8834 (regression).\n- On-demand FP4 experts at 4096: 16.3772 (regression).\n\nConclusion: BF16-resident on-demand quantization is viable and avoids retaining duplicate checkpoint scales (~9 GB/rank in the discarded approach), but act_quant/weight precision does not explain or close the large long-context gap. These paths remain explicit diagnostics; next investigation should target earlier hidden-state divergence and non-quantization execution differences.\n\nValidation:\n- 13 focused CPU tests passed; 12 GPU-only skips.\n- 3 focused SM90 GPU tests passed.\n- make quality passed.\n- make check-patchgen passed.\n- Independent final review: safe.

Luosuu:
RMSNorm parity update

- Matched the official weighted RMSNorm ordering: normalize in FP32, multiply by `weight.float()` in FP32, then cast once to the input dtype.
- Added an exact BF16 regression test. The old ordering differed in 44/192 elements (max abs diff 0.015625); the patched output is bit-exact with the official formula.
- Added reusable layer-detail tracing and an identical-input official-vs-VeOmni sparse-attention comparator. The sparse kernels agree closely (mean abs diff 5.14e-08, RMS 1.43e-05, max 0.015625).
- 256-token end-to-end mean abs logprob diff changed from 12.588826 to 12.718123 (+0.129297, ~1.0%), so official RMSNorm fixes a real local semantic mismatch but does not close the global divergence. I skipped the 4096-token rerun because the directional gate did not improve.
- Earliest hidden-state mean error improved slightly at layer 0 (2.32e-05 -> 2.13e-05), but learned routing from layer 3 still amplifies another remaining source. Output projection / distributed accumulation is the next likely boundary to isolate.

Commits: `42424f0f`, `4a4ae127`. Validation: focused tests 5/5, `make quality`, `make check-patchgen`, explicit Ruff checks for diagnostic scripts, and independent review verdict `safe`.

Luosuu:
Numerical divergence update (256-token controlled traces, identical official layer inputs):

- Matched official weighted RMSNorm, BF16 per-head Q normalization, FP32 RoPE tables, and unconditional FP32 hash/score router projections.
- Q/KV projections are now bit-exact through `q_b_proj` / `kv_norm`; sparse-attention mean error falls to `8.37e-6`.
- `o_a` bmm vs official einsum is bit-exact on identical tensors. Its observed mean error is upstream amplification only (`1.94e-4`).
- Isolated `o_b` replay found a real structural mismatch: official 4 x K=2048 FP8 GEMMs plus reduction is bit-exact, while one K=8192 FP8 GEMM has mean/RMS/max error `0.0019404 / 0.0042269 / 0.0625`.
- Preserving those four partitions slightly improves layer-0 output (`0.0010542 -> 0.0010510`) but does not improve terminal logprobs, showing this is not the dominant source of the 16-point metric.
- The first discrete bifurcation is layer 3, the first score-routed MoE layer. Official FP32 routing doubles exact expert-set overlap from `31.64%` to `63.28%` (and improves later layers similarly), but terminal 256-token mean logprob diff moves `12.2808 -> 12.8273` because the top-k trajectory is chaotic.
- On identical layer-3 router inputs, FP32 projection reproduces official expert sets/order 100%; BF16 gives 97.66% exact sets and 91.02% exact order.

Conclusion: the huge terminal logprob gap is primarily amplification through later discrete MoE routing, seeded by small continuous attention/projection errors. Terminal logprob diff is not a monotonic acceptance metric for individual official-semantic fixes; first-divergence tensor error and routing overlap are more diagnostic.

Pushed: `1e57c89e` (projection diagnostics), `94b5a9e8` (official Q/RoPE/router semantics). Validation: focused tests 6 passed, `make quality`, patchgen consistency, explicit script Ruff checks, independent review safe.

Luosuu:
Official MoE top-k replay ablation (256 tokens):

Added trace-only `--reference-moe-topk` in `eecd2ac6`. It replaces only expert IDs with the official trace and recomputes weights from VeOmni router logits; attention, activations, experts, quantization, and output path remain VeOmni.

Results versus the identical FP32-router baseline:

- routing overlap: exact at every layer (`100%`)
- mean absolute target-logprob diff: `12.8273 -> 12.9082` (slightly worse)
- median: `12.8750 -> 12.7500` (better)
- p99: `22.9950 -> 22.5088` (better)
- max: `25.8749 -> 28.4375` (worse)
- tokens improved: `131 / 255`
- layer-42 sampled hidden-state mean error: `1.2688 -> 0.5143`

Conclusion: official expert IDs substantially reduce internal hidden-state divergence and improve the typical token, confirming MoE routing is a major amplifier. They do not reduce the mean logprob metric because a small set of larger terminal-output regressions dominates the average. The remaining terminal gap is therefore in continuous arithmetic/output-path parity, not expert-ID mismatch alone.

Validation: four-GPU controlled trace completed, script Ruff/format/compile checks passed, `make quality` passed, independent review safe.

Luosuu:
Parity correction and lowest-diff update (`1f713a76`):

The earlier 12–16 mean logprob differences were invalid. `trace_veomni.py` pre-shifted labels even though `chunk_logprobs_function` applies its own causal shift, so it compared official token `t+1` against VeOmni token `t+2`. The corrected harness passes `labels=input_ids`.

Root-cause findings:
- Replaying the identical official terminal hidden state through the VeOmni LM head gives mean abs logprob diff `1.32e-7` (max `9.54e-7`), so the LM head is not the source.
- Official and HF packed FP4 expert weights/scales are byte-identical; BF16 -> on-demand FP4 reproduces all nonzero FP4 codes and every E8M0 scale (only negative zero canonicalizes to positive zero).
- VeOmni shared experts missed official `swiglu_limit=10` FP32 clamp/SwiGLU semantics. The patchgen source now matches official inference and includes a regression test.

Corrected 256-token results (mean / p99 / max abs logprob diff):
- practical native BF16: `0.06019 / 1.08950 / 1.39349`
- native full official quant simulation: `0.08296 / 1.51534 / 3.14326`
- BF16 + official expert IDs and router weights: **`0.01774 / 0.27628 / 0.58198`** (lowest valid path)
- full quant + official IDs and weights: `0.02205 / 0.36378 / 0.97750`

Fresh corrected 4096-token validation of the lowest path:
- **mean `0.005597`, p99 `0.17216`, max `0.45672`**
- terminal normalized hidden mean/RMS diff `0.008071 / 0.011667`

Conclusion: keep weights BF16-resident, use the official shared-expert arithmetic, and quantize selectively/on demand. Reusing official routing is a strong diagnostic lower bound (70.5% lower mean at 256), but is not a deployable native-inference solution; native router divergence remains the largest practical gap.

PR #921, opened 2026-07-15
[model, data, dist] feat: add GPT-OSS 120B EP4 LoRA recipe

Luosuu:
## Summary

- add a local EP=4 LoRA SFT recipe for `unsloth/gpt-oss-120b-BF16` and `allenai/tulu-3-sft-mixture`
- use the tokenizer-native GPT-OSS chat template with assistant-only loss masking, including safe handling of the GPT-OSS terminal-token rewrite
- support native LoRA parameters in the EP-streamed checkpoint loader
- disable Quack forward/backward autotuning for the dynamic MoE GEMMs
- avoid full-size FP32 gradient temporaries during FSDP2 norm clipping

The recipe keeps CPU and activation offloading disabled. Attention LoRA keeps optimizer state within device memory after full-parameter AdamW exhausted the remaining VRAM, allowing mixed-precision FSDP with FP32 sharded storage, BF16 compute/all-gathers, and FP32 gradient reduction to fit safely.

## Local storage

The recipe comments show how to place temporary files, Hugging Face cache, Quack cache, the model, and dataset on local NVMe storage instead of the root disk.

## Validation

- `make quality`
- 28 focused tests covering chat-template masking, FSDP2 gradient norms, LoRA targeting, GPT-OSS integration, and FA4 attention
- actual three-turn preprocessing with the downloaded GPT-OSS tokenizer
- a 20-step four-GPU training run with EP=4, no offloading, mixed precision, and LoRA:
  - first-five mean loss: 3.036
  - last-five mean loss: 1.396 (54% lower)
  - peak allocated VRAM: 115.50 GB per GPU
  - 73 checkpoint shards streamed successfully
  - DCP checkpoint and LoRA adapter saved at step 20; process exited normally


veomni/distributed/fsdp2/clip_grad_norm.py
```diff
@@ -275,13 +256,12 @@ def flush_chunk(device: torch.device) -> None:
...
                 g = g.to_local()
             g = g.detach()
-            chunk = chunks.setdefault(g.device, [])
-            chunk.append(g)
-            if len(chunk) >= _LOCAL_NORM_CHUNK_SIZE:
-                flush_chunk(g.device)
-
-        for device in list(chunks):
-            flush_chunk(device)
+            # ``dtype`` controls accumulation inside the reduction kernel. In
```
Luosuu, line 259: is this fully equivelant?

veomni/models/module_utils.py
Luosuu, line 530: does test cover this and our modifications?

veomni/data/chat_template.py
```diff
@@ -96,6 +96,72 @@ def get_jinja_template(self) -> str:
         )
 
 
+@CHAT_TEMPLATE_REGISTRY.register("tokenizer")
```
Luosuu, line 99: is this GPT-OSS specific?

Luosuu (commented):
try to make diff smaller and safer

veomni/data/chat_template.py
```diff
@@ -96,6 +96,72 @@ def get_jinja_template(self) -> str:
...
+                rewritten_tokens = self.tokenizer.convert_ids_to_tokens(
+                    [input_ids[-1], current_ids[previous_length - 1], current_ids[previous_length]]
+                )
+                is_gpt_oss_terminal_rewrite = rewritten_tokens == ["<|return|>", "<|end|>", "<|start|>"]
+
+            if rewritten_positions and not is_gpt_oss_terminal_rewrite:
+                raise ValueError(
+                    "The tokenizer chat template structurally rewrote an earlier conversation prefix; "
+                    "only the GPT-OSS terminal-token substitution is supported."
+                )
```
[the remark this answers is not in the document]
Luosuu: Addressed in f0b121de. The native tokenizer path no longer compares GPT-OSS token strings. It accepts a token-agnostic terminal substitution only when the previous prefix differs at its final position and the old terminal was not displaced into the appended suffix. Structural insertion tests cover both one- and multi-token displacement, and the downloaded GPT-OSS tokenizer still passes the multi-turn alignment check.

veomni/models/module_utils.py
Luosuu: Yes. f0b121de adds a focused EP-stream loader test that verifies base-checkpoint keys are remapped into LoRA base_layer parameters, the runtime parallel plan is used, only adapter parameters remain after base loading, and the adapter path plus plan are forwarded to the native LoRA loader. The real EP=4 GPT-OSS run also streamed all 73 shards and completed 20 optimizer steps successfully.

veomni/data/chat_template.py
Luosuu: No. TokenizerTemplate is a generic tokenizer-native template with role/loss-mask-based assistant labels. The GPT-OSS-specific token-name check has been removed; terminal substitutions are validated structurally and without model-specific IDs or strings. GPT-OSS is one consumer of this generic path, and both synthetic generic tests and the actual GPT-OSS tokenizer are covered.

Luosuu:
Addressed the review feedback in f0b121de:

- removed GPT-OSS token-string coupling from the generic tokenizer template and strengthened structural rewrite checks;
- preserved prior gradient-norm behavior while avoiding full FP32 copies for common training dtypes, with dtype/order equivalence tests;
- added direct EP-streamed LoRA loader coverage for base-key remapping and adapter handoff.

Validation is green locally: 42 focused tests, the downloaded GPT-OSS tokenizer regression, and the repository quality gate.

veomni/distributed/fsdp2/clip_grad_norm.py
Luosuu: The memory-sensitive path is equivalent to the previous FP32 conversion for FP16, BF16, and FP32 gradients: vector_norm performs FP32 accumulation directly, without first allocating a full-size FP32 gradient. Wider or uncommon dtypes still use the prior explicit FP32 conversion behavior. I added comparisons against the old implementation across FP16, BF16, FP32, and FP64 with norm orders 1, 2, and 3.5; all cases pass.

veomni/data/chat_template.py
```diff
@@ -96,6 +96,71 @@ def get_jinja_template(self) -> str:
...
+            )
+            current_ids = encoded["input_ids"]
+            current_length = len(current_ids)
+            if current_length < previous_length:
+                raise ValueError(
+                    "The tokenizer chat template shortened the conversation after adding a message; "
+                    "assistant-only loss masking requires monotonic message boundaries."
+                )
+
+            rewritten_positions = [index for index in range(previous_length) if input_ids[index] != current_ids[index]]
```
Coach257: This is gpt-oss specific. Prefer a base tokenizer ChatTemplate for the prefix-stable `apply_chat_template` path, and a GPT-OSS subclass that owns the rewrite logic.

veomni/models/module_utils.py
```diff
@@ -515,24 +515,41 @@ def load_model_weights_ep_sharded(
...
+    needs the whole tensor set). A converter that only zero-pads dim-0 stays streamable -- each
     rank reads its real-row slice and zero-fills the tail -- and opts in via the
     optional ``CheckpointTensorConverter.is_dim0_zero_pad`` capability (see
     :func:`checkpoint_converter_is_dim0_zero_pad`).
     """
-    if kwargs.get("is_peft_model", False):
-        raise NotImplementedError("ep_sharded_stream_load does not support PEFT models.")
+    is_peft_model = kwargs.get("is_peft_model", False)
+    adapter_path = kwargs.get("adapter_path", None)
+    if is_peft_model:
```
Coach257: #915

veomni/data/chat_template.py
Luosuu: Follow-up in f50b16f2: the prefix-stable tokenizer-native masking algorithm remains generic, but the terminal rewrite compatibility is indeed GPT-OSS-specific. It now lives only in , registered as ; the generic  rejects all prefix rewrites.

veomni/models/module_utils.py
Luosuu: Follow-up: after considering #915, 7970dabd removes this loader implementation and its focused test from #921 entirely. The earlier coverage answer applied to the previous revision; the current diff deliberately delegates PEFT + EP-sharded loading and its coverage to #915.

veomni/models/module_utils.py
Luosuu: Addressed in 7970dabd. PR #915 is now the sole owner of PEFT + EP-sharded loading: #921 removes its  overlap and the duplicate loader test. The GPT-OSS recipe therefore treats #915 as a prerequisite and will be rebased after #915's loader correctness fixes land, keeping this PR's diff focused.

veomni/data/chat_template.py
Luosuu: Addressed in f50b16f2. The generic tokenizer template now enforces strict prefix stability. A separately registered GPT-OSS subclass owns the exact return-to-end terminal-token rewrite and label update, and the GPT-OSS recipe selects that subclass. Tests verify the generic path rejects rewrites while the GPT-OSS path accepts only its expected terminal substitution.

veomni/models/module_utils.py
Luosuu: Small correction to the previous reply: the removed overlap was in veomni/models/module_utils.py. The implementation and duplicate loader test are both absent from the current PR diff.

veomni/data/chat_template.py
Luosuu: For clarity, the new classes are GptOssTokenizerTemplate for the registered gpt_oss path and TokenizerTemplate for the strict generic path.

Coach257 (approved):

PR #879, opened 2026-07-02
[ops, model] feat: add NPU backend for Qwen3.5 gated delta-rule ops

xssty123:
### What
Add an NPU (`npu`) backend for Qwen3.5's gated delta-rule linear-attention
ops — `causal_conv1d` and `chunk_gated_delta_rule` — through the existing
`OpSlot` / `KERNEL_REGISTRY` dispatch, so NPU users can run Qwen3.5(-MoE)
varlen training. `rms_norm_gated` already had an `npu` backend; this fills the
remaining two.

### Why
On GPU these ops dispatch to `fla` / `flash_qla`. On NPU they had no non-eager
backend, so varlen (`dyn_bsz`) training would error out. This adds the missing
NPU path.

### How
Split into two commits so copied vs authored code can be reviewed separately:
- **`chore`** — ~4.1k lines of Ascend Triton kernels vendored **verbatim** from
  MindSpeed-MM (`https://gitcode.com/Ascend/MindSpeed-MM`), commit `39775d87`
  (2026-05-13), under `veomni/ops/kernels/gated_delta_rule/_ascend/`
  (ruff-excluded). All 10 files are byte-identical to that upstream commit;
  FLA-derived (Songlin Yang, Yu Zhang) + Huawei Ascend port, Apache-2.0,
  copyright headers preserved. No kernel logic modified.
- **`feat`** — ~300 lines of VeOmni integration:
  - `causal_conv1d`: thin adapter (`npu_causal_conv1d.py`) bridging the FLA call
    signature to the vendored kernel. `chunk_gated_delta_rule`: bound directly.
  - Registered as `name="npu"` with `HardwareRequirement(device_type="npu")`,
    mirroring the existing `rms_norm_gated` npu backend. Lazy factory imports —
    non-NPU hosts don't pull Triton.
  - No hand-edits under `generated/`: `patched_modeling_*_npu.py` stay on the
    pristine OpSlot path (dispatch happens via the registry).
  - Docs: `docs/examples/qwen3_5.md` (linear-attention kernels section) now
    documents how to enable the NPU backend (config + yaml) and the arch35
    limitation below.

### Testing
- NPU e2e: Qwen3.5-MoE VL SFT on Ascend 910B2C — loss/accuracy matches the
  reference run. The vendored kernels are exercised only by this manual run;
  no CI e2e covers the varlen-GDN NPU path.
- CI: `tests/ops/test_gdn_ascend_gate.py` — hardware-gate unit tests (mocked;
  run on any host).

[...]

PR #924, opened 2026-07-16
[ops] feat: add AscendC fused backend (npu_ascendc) for Qwen3.5 chunk gated delta rule

xssty123:
### What does this PR do?

Adds a new **`npu_ascendc`** backend for Qwen3.5 GatedDeltaNet's
`chunk_gated_delta_rule`, backed by the AscendC fused kernels from upstream
`fla_npu`. It complements the `fla` (GPU) and `npu` (Triton) backends added in
#879, giving NPU users a fused AscendC path for `chunk_gated_delta_rule`.

> **Draft** — numerically verified and running end-to-end on Ascend 910B; opened
> as draft for provenance/design review.

### Checklist Before Starting

- Related PR: #879 (added the `fla` / `npu` GDN backends this builds on).
- PR title follows `[{modules}] {type}: {description}` —
  `[ops] feat: add AscendC fused backend (npu_ascendc) for Qwen3.5 chunk gated delta rule`.

### Test

- [x] `make quality` (ruff check + format) passes.
- [x] End-to-end training runs on Ascend 910B with
  `chunk_gated_delta_rule_implementation=npu_ascendc`.
- [x] Numerical parity (`get_err_ratio`) vs GPU `fla` on the Qwen3.5
  GatedDeltaNet forward/backward.

No CI e2e covers the AscendC GDN path; the kernels are manually verified on
Ascend 910B via the end-to-end run and numerical-parity check above.

### API and Usage Example

New enum value `npu_ascendc` for `chunk_gated_delta_rule_implementation`. On NPU,
set all three GDN op implementations explicitly (the `fla` default raises when
bound on NPU):

```yaml
model:
  chunk_gated_delta_rule_implementation: npu_ascendc
  causal_conv1d_implementation: npu
  rms_norm_gated_implementation: npu
```

The AscendC path additionally requires the `fla_npu` package. The prebuilt NPU
images on [quay.io](https://quay.io/repository/ascend/veomni?tab=tags) already
bundle it (A2 and A3); manual install instructions are in
`docs/examples/qwen3_5.md`.

### Design & Code Changes

Split into two commits so vendored vs authored code review separately (same
convention as #879):

**1. `chore` — vendor the gated delta-rule Triton kernels**

4 kernels under the ruff-excluded
`veomni/ops/kernels/gated_delta_rule/_ascend/triton_core/` tree —
`{chunk_scaled_dot_kkt, l2norm, solve_tril, utils}.py`.
Vendored via MindSpeed-MM's `triton_core`, which redistributes **fla_npu commit
`c2e3d83f`** (https://github.com/flashserve/flash-linear-attention-npu). No VeOmni
logic; kept
as a drop-in vendor blob so they stay diff-able against upstream — any
kernel-logic change must go upstream. (The one intentional deviation, the
`l2norm_bwd` fix below, lives in the `feat` commit, not here.)

**2. `feat` — VeOmni integration**

- **AscendC entry** (`_ascend/flash_gated_delta_rule.py`): adapted from
  MindSpeed-MM's wrapper (itself derived from fla_npu commit `c2e3d83f`) — strips MM's
  async-offload / `skip_recompute` machinery down to the plain compute path, and
  rewrites imports to package-relative. Delegates the heavy GDN compute to the
  external `fla_npu` `torch.ops.npu.*` fused ops. This is an *adapted* file, so
  it lives here rather than in the verbatim `chore` commit.
- **Package init** (`_ascend/triton_core/__init__.py`): VeOmni-authored init for
  the vendored kernels.
- **Layout adapter** (`npu_ascendc_gated_delta_rule.py`): the generated modeling
  emits `q/k/v` as `[B, T, H, D]` (FLA layout), but `flash_gated_delta_rule`
  expects — and validates — `[B, H, T, D]`. Rather than fork the generated
  `forward` into per-implementation branching (a patchgen change), the adapter
  transposes `q/k/v` at the op boundary. `g`/`beta` (`[B, T, H]`) and the
  returned `o` (`[B, T, H, V]`) already match, so no output transpose.
  Numerically equal to MindSpeed-MM's AscendC path; cost is one input transpose.
- **Registration**: `_ascend/__init__.py` exposes the `npu_ascendc` KernelSpec;
  `gated_delta_rule/__init__.py` + `arguments/arguments_types.py` surface and
  validate the selection. Lazy factory import — non-NPU hosts don't pull
  `fla_npu` / Triton.
- **Docs**: `docs/examples/qwen3_5.md`, `docs/design/kernel_selection.md`,
  `docs/design/unified_kernel_registry.md`.

**One intentional deviation from vendored — `l2norm_bwd` (in the `feat` commit)**

`l2norm_bwd_kernel` computed its task count as `T // BT` (floor), dropping the
partial last block when `T` (the flattened `B·H·seq` row count) isn't a multiple
of `BT=64`. Those rows are never scheduled, so their `dx` keeps uninitialized
`torch.empty_like` memory and explodes in the l2norm backward for non-64-aligned
(packed/varlen ragged) segments — grad_norm ~11× high. Forward already
ceil-covers all blocks; the fix aligns backward (`floor -> ceil`) so the existing
`block_start < T` guard + `boundary_check` handle the partial block. Reported
upstream to `fla_npu`.

**Limitation — arch35:** upstream guards `is_arch35()`
(`Ascend910_95` / `Ascend950`) with `NotImplementedError`; validated target is
910B, where the guard doesn't fire. arch35 support needs upstream.

- [x] Applied pre-commit checks (`make quality`).
- [x] Added/updated documentation (`docs/examples/qwen3_5.md`,
  `docs/design/unified_kernel_registry.md`).
- [x] Manually verified on Ascend 910B (end-to-end run + numerical parity vs
  eager); no CI e2e for the AscendC GDN path (needs NPU hardware).
- [x] No `tasks/` training scripts moved or renamed.

### DIFF
--- a/docs/design/kernel_selection.md
+++ b/docs/design/kernel_selection.md
@@ -25,7 +25,7 @@ selection knob.
 | Vision rotary embedding | `rotary_pos_emb_vision_implementation` | `eager`, `npu` | `"eager"` | Model registration via ops config singleton |
 | Gated RMSNorm | `rms_norm_gated_implementation` | `eager`, `fla`, `npu` | `"fla"` (GPU) | Qwen3.5 OpSlot binding |
 | Causal Conv1D | `causal_conv1d_implementation` | `eager`, `fla`, `npu` | `"fla"` (GPU) | Qwen3.5 OpSlot binding |
-| Gated delta rule | `chunk_gated_delta_rule_implementation` | `eager`, `fla`, `flash_qla` (SM90), `npu` | `"fla"` (GPU) | Qwen3.5 OpSlot binding |
+| Gated delta rule | `chunk_gated_delta_rule_implementation` | `eager`, `fla`, `flash_qla` (SM90), `npu`, `npu_ascendc` | `"fla"` (GPU) | Qwen3.5 OpSlot binding |
 | Load-balancing loss | `load_balancing_loss_implementation` | `eager`, `triton` (CUDA; NPU config normalizes this default to `eager`) | `"triton"` | `apply_ops_config()` (before model build) |
 | MoE experts | `moe_implementation` | `eager`, `fused_triton`, `fused_quack` (SM90+), `fused_npu` | `"fused_triton"` (GPU) | `build_foundation_model` |
 
@@ -292,12 +292,16 @@ model:
 |---|---|---|---|
 | `rms_norm_gated_implementation` | `fla` | `npu` | HuggingFace reference implementation |
 | `causal_conv1d_implementation` | `fla` | `npu` | No `cu_seqlens` path |
-| `chunk_gated_delta_rule_implementation` | `fla`, `flash_qla` (SM90 only) | `npu` | No `cu_seqlens` path |
+| `chunk_gated_delta_rule_implementation` | `fla`, `flash_qla` (SM90 only) | `npu`, `npu_ascendc` | No `cu_seqlens` path |
 
 The NPU gated RMSNorm uses `torch_npu`. The NPU causal Conv1D and gated
-delta-rule implementations additionally require `triton-ascend`. Registrations
-live in `veomni/ops/kernels/gated_delta_rule/__init__.py`; field defaults and
-allowed values are documented by `OpsImplementationConfig`.
+delta-rule implementations additionally require `triton-ascend`. For the gated
+delta rule there are two NPU backends: `npu` (the vendored MindSpeed-MM Triton
+kernel) and `npu_ascendc` (an AscendC fused `torch.ops.npu.*` path), the latter
+requiring a manual `fla_npu` install and `ASCEND_CUSTOM_OPP_PATH` exported before
+process start. Registrations live in
+`veomni/ops/kernels/gated_delta_rule/__init__.py`; field defaults and allowed
+values are documented by `OpsImplementationConfig`.
 
 ---
 
--- a/docs/design/kernel_selection.md
+++ b/docs/design/kernel_selection.md
@@ -298,8 +298,7 @@ The NPU gated RMSNorm uses `torch_npu`. The NPU causal Conv1D and gated
 delta-rule implementations additionally require `triton-ascend`. For the gated
 delta rule there are two NPU backends: `npu` (the vendored MindSpeed-MM Triton
 kernel) and `npu_ascendc` (an AscendC fused `torch.ops.npu.*` path), the latter
-requiring a manual `fla_npu` install and `ASCEND_CUSTOM_OPP_PATH` exported before
-process start. Registrations live in
+requiring a manual `fla_npu` install. Registrations live in
 `veomni/ops/kernels/gated_delta_rule/__init__.py`; field defaults and allowed
 values are documented by `OpsImplementationConfig`.
 
--- a/docs/examples/qwen3_5.md
+++ b/docs/examples/qwen3_5.md
@@ -207,6 +207,53 @@ model:
     chunk_gated_delta_rule_implementation: flash_qla
 ```
 
+#### `npu_ascendc` — AscendC fused backend (NPU, `chunk_gated_delta_rule` only)
+
+`npu_ascendc` is a second NPU backend for `chunk_gated_delta_rule` that delegates the heavy GDN
+compute to the external [`fla_npu`](https://github.com/flashserve/flash-linear-attention-npu)
+package (registered as `torch.ops.npu.*` fused ops); only the Triton glue stays vendored under
+`_ascend/triton_core`. It coexists with `npu` (pure vendored Triton), which remains the fallback.
+Set only `chunk_gated_delta_rule` to `npu_ascendc`; keep `rms_norm_gated` / `causal_conv1d` on `npu`:
+
+```yaml
+model:
+  ops_implementation:
+    rms_norm_gated_implementation: npu
+    causal_conv1d_implementation: npu
+    chunk_gated_delta_rule_implementation: npu_ascendc
+```
+
+`fla_npu` is not a declared VeOmni dependency (same as `triton-ascend`). It supports Ascend
+910B (A2) / 910_93 (A3) / 950 (A5); on non-arch35 chips the `solve_tril` step auto-falls-back to
+the vendored Triton kernel via `is_arch35()`.
+
+The prebuilt NPU images on [quay.io](https://quay.io/repository/ascend/veomni?tab=tags) already
+bundle `fla_npu`:
+
+- A2: `quay.io/ascend/veomni:v0.1.11-cann9.0.0-torch_npu2.10.0.post2-A2-ubuntu22.04-py3.11-veomni-latest`
+- A3: `quay.io/ascend/veomni:v0.1.11-cann9.0.0-torch_npu2.10.0.post2-A3-ubuntu22.04-py3.11-veomni-latest`
+
+To build on a bare host instead, install from the
+[`flash-linear-attention-npu`](https://github.com/flashserve/flash-linear-attention-npu) sources:
+
+```bash
+git clone https://github.com/flashserve/flash-linear-attention-npu
+cd flash-linear-attention-npu
+git checkout c2e3d83f
+
+source /usr/local/Ascend/cann/set_env.sh          # source your actual CANN path
+
+# --soc must match the chip: {ascend910b, ascend910_93, ascend950}
+bash build.sh --soc=ascend910b --pkg --vendor_name=fla_npu
+bash build_out/fla-npu_*.run
+cd torch_custom/fla_npu/ && bash build.sh
+
+pip list | grep fla_npu   # verify it is installed
+```
+
+If `fla_npu` is absent when `npu_ascendc` is selected, the backend raises an actionable error at
+`OpSlot.bind()` time (pointing back to the install step or to `npu` / `eager`).
+
 ### How It Works
 
 Qwen3.5 is a hybrid model alternating between softmax and linear attention layers:
--- a/docs/examples/qwen3_5.md
+++ b/docs/examples/qwen3_5.md
@@ -223,9 +223,14 @@ model:
     chunk_gated_delta_rule_implementation: npu_ascendc
 ```
 
+A ready-to-run MoE-VL training config wired for this backend (plus `fused_npu` MoE and
+Ulysses SP) is provided at
+[`configs/multimodal/qwen3_5_moe/qwen3_5_moe_vl_ascendc.yaml`](../../configs/multimodal/qwen3_5_moe/qwen3_5_moe_vl_ascendc.yaml).
+
 `fla_npu` is not a declared VeOmni dependency (same as `triton-ascend`). It supports Ascend
-910B (A2) / 910_93 (A3) / 950 (A5); on non-arch35 chips the `solve_tril` step auto-falls-back to
-the vendored Triton kernel via `is_arch35()`.
+910B (A2) / 910_93 (A3), both non-arch35 chips where the `solve_tril` step runs the vendored
+Triton kernel (the arch35 native path is gated behind `is_arch35()`). arch35 (`Ascend910_95` /
+`Ascend950`) is not supported end-to-end: the vendored `causal_conv1d` raises there (see above).
 
 The prebuilt NPU images on [quay.io](https://quay.io/repository/ascend/veomni?tab=tags) already
 bundle `fla_npu`:
--- a/veomni/ops/kernels/gated_delta_rule/__init__.py
+++ b/veomni/ops/kernels/gated_delta_rule/__init__.py
@@ -36,11 +36,18 @@ Backends per op:
 - ``rms_norm_gated``: ``fla`` (GPU), ``npu``.
 - ``causal_conv1d``: ``fla`` (GPU), ``npu``.
 - ``chunk_gated_delta_rule``: ``fla`` (GPU), ``flash_qla`` (GPU ``gpu`` extra,
-  Hopper SM90), ``npu``.
+  Hopper SM90), ``npu`` (vendored Triton), ``npu_ascendc`` (AscendC fused ops).
 
 The ``npu`` ``causal_conv1d`` backend is a thin adapter (``npu_causal_conv1d``)
 over the vendored kernel; the ``npu`` ``chunk_gated_delta_rule`` binds the
 vendored kernel directly (its signature already matches the call site).
+
+The ``npu_ascendc`` ``chunk_gated_delta_rule`` backend is a second NPU path that
+delegates the heavy compute to the external ``fla_npu`` package
+(``torch.ops.npu.*`` fused ops, installed manually on NPU), using a few Triton
+kernels from ``_ascend/triton_core`` (a newer generation of Triton kernels) as
+glue around them. It coexists with the pure-Triton ``npu`` backend, which stays
+as the fallback.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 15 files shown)


xssty123:
**Re: the three automated findings — all in the vendored `_ascend/triton_core/` tree**

Thanks for the review. All three land in the vendored Triton tree
(`_ascend/triton_core/`), kept diff-able against upstream fla_npu (commit
`c2e3d83f`, via MindSpeed-MM) — the only intentional deviation is the
`l2norm_bwd` floor→ceil fix called out in the PR description. Traced against
VeOmni's actual `npu_ascendc` dispatch path:

1. **`cumsum.py:94`** — this file has been **removed from the PR**.
   `triton_core/cumsum.py` (and `causal_conv1d.py`) were vendored but never
   imported on this path — `flash_gated_delta_rule.py` uses the sibling
   `.triton.cumsum`, and nothing in the repo imports `triton_core.cumsum` — so I
   dropped both unused files rather than carry them. The finding is therefore
   moot.

2. **`l2norm.py:202` (unflattened `rstd`)** — not a bug on this path: `rstd`
   comes from `l2norm_fwd` as a freshly-allocated **contiguous** tensor
   `.view(...)`d to `[B, H, T]`; a contiguous `[B, H, T]` read as a flat
   `[B*H*T]` array is byte-identical to `.flatten()`, so there is no numerical
   difference (the suggested `.flatten()` is a no-op here). End-to-end varlen
   numerics match GPU `fla` via `get_err_ratio`.

3. **`chunk_scaled_dot_kkt.py:287` (`g=None` → `AttributeError`)** — the sole
   caller (`flash_gated_delta_rule`) always passes a non-`None` `g` (the output
   of `chunk_local_cumsum`) and never passes `gk`, so the `g is None` branch is
   unreachable in this integration.

(2) and (3) are latent robustness points in upstream code; I'm keeping the
vendored files unmodified here for clean provenance, and am happy to forward
them to fla_npu.

docs/examples/qwen3_5.md
```diff
@@ -207,6 +207,48 @@ model:
...
+```yaml
+model:
+  ops_implementation:
+    rms_norm_gated_implementation: npu
+    causal_conv1d_implementation: npu
+    chunk_gated_delta_rule_implementation: npu_ascendc
+```
+
+`fla_npu` is not a declared VeOmni dependency (same as `triton-ascend`). It supports Ascend
+910B / 910_93 / 950 and A2 / A3 / A5; on non-arch35 chips the `solve_tril` step auto-falls-back to
```
Crystal-jiang: Please clarify the mapping between 910B/910_93/950 and A2/A3/A5.

xssty123: 910B(A2)、910_93(A3)、950(A5)

docs/examples/qwen3_5.md
```diff
@@ -207,6 +207,48 @@ model:
...
+the vendored Triton kernel via `is_arch35()`.
+
+Install it from the [`flash-linear-attention-npu`](https://github.com/flashserve/flash-linear-attention-npu)
+sources, following MindSpeed-MM's Qwen3.5 README
+([`examples/qwen3_5/README.md`](https://gitcode.com/Ascend/MindSpeed-MM/blob/master/examples/qwen3_5/README.md)):
```
xssty123: remove MindSpeed-MM link

acat-rw:
Why is similar code needed in both **./gated_delta_rule/_ascend/triton_core** and **./gated_delta_rule/_ascend/triton**?

docs/examples/qwen3_5.md
xssty123: Done — the doc now spells out the mapping: `910B (A2) / 910_93 (A3) / 950 (A5)`.

docs/examples/qwen3_5.md
xssty123: Removed — the manual-build section now links only the `flash-linear-attention-npu` source repo, no MindSpeed-MM link.

xssty123:
@acat-rw `triton/` and `triton_core/` are two different upstream kernel *generations*, not a copy. `triton/` is the pure-`npu` Triton set vendored via MindSpeed-MM (added in #879); `triton_core/` is the newer generation that fla_npu's AscendC path pairs with (commit `c2e3d83f`) — the three overlapping files differ (rewritten `solve_tril`, autotuned `chunk_scaled_dot_kkt`). Each is kept as a byte-faithful vendor blob so it stays diff-able against its own upstream; merging them would mean hand-editing vendored code. The `npu_ascendc` entry deliberately reuses the unchanged `triton/` pieces (`cumsum`, `is_arch35`) and only pulls the rewritten kernels from `triton_core/`.

veomni/ops/kernels/gated_delta_rule/_ascend/flash_gated_delta_rule.py
```diff
@@ -0,0 +1,623 @@
...
+        chunk_indices=_chunk_tensor(chunk_indices, chunk_size),
+        chunk_size=chunk_size,
+        output_dtype=torch.float32,
+    )
+
+    if is_arch35():
+        A = solve_tril(
+            A=A,
+            cu_seqlens=cu_seqlens,
+            chunk_indices=chunk_indices,
```
FoolPlayer: This passes the per-chunk-size **dict**, but `triton_core/solve_tril.py:315` takes a `LongTensor` and does `NT = len(chunk_indices)` at `:340` — with a dict that's the key count (~7), not the chunk count, and the dict then gets forwarded into the Triton launch. Every other consumer in this file unwraps it (`:191`, `:223`, `:237`, …); this branch looks like it just missed it.

```suggestion
            chunk_indices=_chunk_tensor(chunk_indices, chunk_size),
```

Only fires on arch35 (`Ascend910_95` / `Ascend950`), which is unreachable end-to-end today because `_ascend/causal_conv1d.py:37` raises first — but that is the chip `docs/examples/qwen3_5.md:227` advertises.

veomni/ops/kernels/gated_delta_rule/_ascend/flash_gated_delta_rule.py
```diff
@@ -0,0 +1,623 @@
...
+  (``TrainingContext`` / ``OffloadManager`` / ``SwapTensor`` / ``get_current_stream``):
+  VeOmni has no such framework, so only the plain compute path remains.
+"""
+
+import warnings
+from typing import Dict, Optional
+
+import torch
+import torch_npu
+import fla_npu
```
FoolPlayer: `import fla_npu` looks unused but is load-bearing (it registers the `torch.ops.npu.*` ops) — worth a `# noqa: F401` plus a one-line note so nobody prunes it.

FWIW ruff never sees it: `pyproject.toml` `extend-exclude` covers all of `_ascend/`, so these 620 lines of VeOmni-authored glue get no lint *or* format coverage (also why the unsorted import blocks and the `if is_arch35()` interleaved at `:29` pass `make quality`). Consider narrowing the exclude to the genuinely-verbatim `triton*/` subtrees.

docs/examples/qwen3_5.md
```diff
@@ -207,6 +207,53 @@ model:
...
+```yaml
+model:
+  ops_implementation:
+    rms_norm_gated_implementation: npu
+    causal_conv1d_implementation: npu
+    chunk_gated_delta_rule_implementation: npu_ascendc
+```
+
+`fla_npu` is not a declared VeOmni dependency (same as `triton-ascend`). It supports Ascend
+910B (A2) / 910_93 (A3) / 950 (A5); on non-arch35 chips the `solve_tril` step auto-falls-back to
```
FoolPlayer: The 950 (A5) claim contradicts `:196-198` in this same file: the vendored `causal_conv1d` raises `NotImplementedError` on arch35 (`Ascend910_95` / `Ascend950`), and the generated modeling raises when `causal_conv1d_fn is None`. So the configuration documented just above (`causal_conv1d_implementation: npu`) cannot run on A5. Suggest dropping 950 here, or noting that A5 additionally needs an arch35-capable `causal_conv1d`.

docs/design/kernel_selection.md
```diff
@@ -292,12 +292,16 @@ model:
...
+| `chunk_gated_delta_rule_implementation` | `fla`, `flash_qla` (SM90 only) | `npu`, `npu_ascendc` | No `cu_seqlens` path |
 
 The NPU gated RMSNorm uses `torch_npu`. The NPU causal Conv1D and gated
-delta-rule implementations additionally require `triton-ascend`. Registrations
-live in `veomni/ops/kernels/gated_delta_rule/__init__.py`; field defaults and
-allowed values are documented by `OpsImplementationConfig`.
+delta-rule implementations additionally require `triton-ascend`. For the gated
+delta rule there are two NPU backends: `npu` (the vendored MindSpeed-MM Triton
+kernel) and `npu_ascendc` (an AscendC fused `torch.ops.npu.*` path), the latter
+requiring a manual `fla_npu` install and `ASCEND_CUSTOM_OPP_PATH` exported before
```
FoolPlayer: This is the only mention of `ASCEND_CUSTOM_OPP_PATH` in the repo — the install recipe users actually follow (`docs/examples/qwen3_5.md`) doesn't mention it, and no code checks it. That means "raises an actionable error at `OpSlot.bind()` time" only holds for a *missing package*: with `fla_npu` installed but the var unset, the import succeeds and the first forward dies inside `npu_recompute_w_u_fwd`. Either add it to the install steps or check it in the factory.

veomni/ops/kernels/gated_delta_rule/__init__.py
```diff
@@ -235,3 +242,50 @@ def _npu_chunk_gated_delta_rule_factory():
...
+
+    Lazily imported so the transitive ``torch_npu`` / ``fla_npu`` / triton-ascend
+    imports only fire when this backend is selected. ``fla_npu`` is installed
+    manually on NPU (not a declared dependency); re-raise its absence with an
+    actionable message instead of a bare ``ModuleNotFoundError``.
+    """
+    try:
+        from .npu_ascendc_gated_delta_rule import chunk_gated_delta_rule
+    except ModuleNotFoundError as e:
+        if e.name in ("fla_npu", "torch_npu"):
```
FoolPlayer, line 281: The docstring above and the `arguments_types.py` help text both list `triton-ascend` as a requirement, but it isn't in this tuple — a missing triton-ascend still surfaces as a bare `ModuleNotFoundError` instead of the actionable message. `("fla_npu", "torch_npu", "triton")` would cover it.

veomni/ops/kernels/gated_delta_rule/__init__.py
```diff
@@ -235,3 +242,50 @@ def _npu_chunk_gated_delta_rule_factory():
...
+KERNEL_REGISTRY.register(
+    KernelSpec(
+        name="npu_ascendc",
+        op_name="chunk_gated_delta_rule",
+        variant="standard",
+        factory=_npu_ascendc_chunk_gated_delta_rule_factory,
+        hardware=HardwareRequirement(device_type="npu"),
+        description="NPU AscendC fused chunk gated delta rule (torch.ops.npu.* via fla_npu, varlen-aware, MindSpeed-MM triton_core)",
+    )
+)
```
FoolPlayer, line 302: `tests/ops/test_gdn_ascend_gate.py:102` (`test_gdn_registry_has_fla_and_npu`) exists specifically to catch a backend being dropped and doesn't know about `npu_ascendc`; the hardware-gate parametrization at `:53` doesn't cover it either. Both run on any host (the `HardwareRequirement` check fires before `spec.factory()`), so adding it is free and would cover registration plus the missing-package guard.

FoolPlayer (commented):
Reviewed the integration path. The A2/910B wiring looks correct: the adapter's signature, layouts and `(o, final_state)` return convention match both generated call sites (`patched_modeling_qwen3_5_npu.py` and the MoE equivalent), and every cross-tree import from `triton/` into `flash_gated_delta_rule.py` resolves with a compatible signature. Left a few inline notes — one real bug on the arch35 branch, the rest docs/tests/lint-coverage.

veomni/ops/kernels/gated_delta_rule/_ascend/flash_gated_delta_rule.py
xssty123: Fixed in dd4ec62d — the arch35 branch now passes `_chunk_tensor(chunk_indices, chunk_size)` (the per-size LongTensor), matching every other consumer and the `else` branch. Good catch.

veomni/ops/kernels/gated_delta_rule/_ascend/flash_gated_delta_rule.py
xssty123: Added `# noqa: F401` + a comment noting `import fla_npu` registers the `torch.ops.npu.*` ops (dd4ec62d). On the exclude: agreed — the whole `_ascend/` being ruff-excluded also skips our authored glue. Narrowing it to the verbatim `triton*/` subtrees needs a `make style` reformat pass, so I am tracking that as a follow-up to avoid adding format churn to this PR.

docs/examples/qwen3_5.md
xssty123: Fixed in dd4ec62d — dropped 950 (A5) from the supported list and added a note that arch35 (`Ascend910_95` / `Ascend950`) is unsupported end-to-end because the vendored `causal_conv1d` raises there.

docs/design/kernel_selection.md
xssty123: Removed in dd4ec62d — `ASCEND_CUSTOM_OPP_PATH` is no longer needed (verified: `fla_npu` runs without it), so the stale mention is dropped rather than added to the install steps.

veomni/ops/kernels/gated_delta_rule/__init__.py
xssty123: Fixed in dd4ec62d — added a dedicated `e.name == "triton"` branch pointing at the `triton-ascend` install, with `eager` as the only dep-free fallback (`npu` needs triton-ascend too). The `fla_npu` / `torch_npu` branch is unchanged.

veomni/ops/kernels/gated_delta_rule/__init__.py
xssty123: Fixed in dd4ec62d — re-added the three hardware-agnostic tests: `test_chunk_gdr_registry_has_npu_ascendc` (registration + not on `causal_conv1d`), `test_opslot_npu_ascendc_backend_on_gpu_raises`, and `test_npu_ascendc_factory_missing_dep_raises_actionable`. All run in GPU CI via `pytest tests/ops`.

FoolPlayer:
Can you help add a qwen3.5 training yaml with ascendc kernels?

xssty123:
Added in be216174: [`configs/multimodal/qwen3_5_moe/qwen3_5_moe_vl_ascendc.yaml`](https://github.com/ByteDance-Seed/VeOmni/blob/feat/qwen3_5-gdn-npu-ascendc/configs/multimodal/qwen3_5_moe/qwen3_5_moe_vl_ascendc.yaml) — a Qwen3.5-35B-A3B MoE-VL SFT config wired for the ascendc kernels: `chunk_gated_delta_rule_implementation: npu_ascendc` with `causal_conv1d` / `rms_norm_gated` on `npu`, plus `fused_npu` MoE and Ulysses SP (`veomni_flash_attention_2_with_sp`). All ops choices are in the NPU allow-list. It is referenced from the `npu_ascendc` section of `docs/examples/qwen3_5.md`.

Zhang1Sheng (approved):

wangtiance:
Hello, will Ascend 950 be supported soon? Thanks!

Issue #927, opened 2026-07-17
[DeepSeek V4] Track runtime parity, FP8 activation QAT, and diagnostics stack


Luosuu:
## Context

Draft PR #917 combines runtime correctness, quantized-inference experiments, FP8 activation QAT, and parity diagnostics. This issue tracks the reviewable training-focused replacement.

## Merge stack

Merge in order because each PR is based on the preceding branch:

- [x] #928 — runtime parity: stateless prefill and official RMSNorm, RoPE, router, and shared-expert semantics
- [ ] #929 — FP8 activation QAT: STE fake quantization, Hadamard rotation, configuration, and fail-fast validation
- [ ] #931 — parity diagnostics: tracing, sparse-attention comparisons, and routing replay

## Superseded experiment

- [x] #930 — closed. FP4 quantization, quantized GEMM, weight repacking, and on-demand expert inference are not used by QAT training and are intentionally excluded from the merge path.

## Required invariants

- BF16 trainable parameters and master weights remain resident.
- QAT simulates FP8 activations and casts back to BF16; it does not use quantized GEMM.
- Unsupported DeepSeek-V4 QAT platforms and missing Hadamard support fail early with no fallback.
- Generated DeepSeek-V4 artifacts remain synchronized with the patch-generation config.
- Quack GEMM autotuning changes remain separate in #922.

## Validation baseline

The integrated implementation in #917 retains the end-to-end investigation history. The slim QAT branch passes eight QAT-specific tests, six model-parity tests, patchgen dry-run, and diff checks. Diagnostic scripts pass compilation and Ruff.

## Integration baseline

#917 remains open as a draft for investigation history and prior CI results. It should not be merged directly after the three child PRs replace it.

PR #928, opened 2026-07-17
[model] fix: align DeepSeek V4 runtime numerical semantics

Luosuu:
## Summary

Align the core DeepSeek-V4 runtime with official execution semantics before introducing quantization experiments.

- keep stateless prefill and training cache-free so the TileLang indexer remains reachable
- match official weighted RMSNorm multiplication order
- preserve checkpoint-stable RoPE dtypes in training and FP32 tables in inference
- normalize attention queries in the official BF16 order
- run top-k and hash router projections in FP32
- match the shared-expert FP32 clamped SwiGLU path

## Root cause

VeOmni created a DynamicCache even when use_cache was false, which forced the indexer onto its eager decode fallback. Several arithmetic-order differences in RMSNorm, routing, attention normalization, and shared experts also accumulated model-body divergence from official inference.

## Review scope

This is stack 1 of 3. It intentionally contains no FP8/FP4 inference kernels, QAT, or tracing scripts.

Tracking issue: #927
Original integration draft: #917

## Validation

- uv run pytest -q tests/models/test_deepseek_v4_fused_moe.py
- stateless-forward focused kernel test
- patchgen regeneration and dry-run
- git diff --check

## Stack navigation

Tracked by #927. Merge and review in this order:

1. #928 — runtime parity
2. #929 — FP8 activation QAT
3. #931 — parity diagnostics

#930 is closed as superseded because its inference-only FP4/GEMM paths are not used by training. The original #917 remains the integration and CI baseline; it is not the merge target for this stack.


veomni/models/transformers/deepseek_v4/deepseek_v4_gpu_patch_gen_config.py
```diff
@@ -914,6 +968,67 @@ def _apply_gate(self, gate_up: torch.Tensor) -> torch.Tensor:
...
+@config.override_method(
+    "DeepseekV4TopKRouter.forward",
+    description="Match the official DeepSeek-V4 FP32 router projection",
+)
+def deepseek_v4_topk_router_forward_patched(
+    self,
+    hidden_states: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    flat = hidden_states.reshape(-1, self.hidden_dim)
+    logits = F.linear(flat.float(), self.weight.float())
```
FoolPlayer: `F.linear` may be cast back to BF16 under an outer autocast context despite the explicit `.float()` calls. Please disable autocast around this projection; the same applies to `DeepseekV4HashRouter`, and an autocast-enabled regression test would help.

Luosuu: Fixed by disabling autocast around both TopK and HashRouter FP32 F.linear projections and adding an autocast-enabled CPU regression test covering both routers.

FoolPlayer (approved):
Reviewed the runtime parity changes, including stateless cache behavior, RMSNorm/RoPE dtype semantics, attention query normalization, FP32 router projections, shared-expert SwiGLU, patchgen consistency, and test coverage. No blocking issues found.

PR #942, opened 2026-07-17
[trainer, perf] refactor: coalesce train metric reductions

zhangxin81:
## Summary

- Coalesce scalar training metric all-reduces in `EnvironMeterCallback` into a single collective when possible.
- Preserve the previous per-key fallback for non-numeric/custom metric values.
- Add focused tests for coalesced numeric reductions and fallback behavior.

## Performance Rationale

`EnvironMeterCallback` previously reduced `total_loss`, every loss entry, and `grad_norm` with one scalar all-reduce per key on every step. These are small control-plane collectives on the hot path. Coalescing numeric scalar metrics into one list all-reduce preserves the logged values while reducing per-step collective count.

## Validation

- `PYTHONPATH=/Users/bytedance/Documents/VeOmni-metric-reduce PATH=/Users/bytedance/Documents/VeOmni/.venv/bin:$PATH pytest tests/trainer/test_trace_callback_metrics.py -q` -> `2 passed`
- `PYTHONPATH=/Users/bytedance/Documents/VeOmni-metric-reduce PATH=/Users/bytedance/Documents/VeOmni/.venv/bin:$PATH make quality` -> ruff check and format passed
- `/veomni-review` verdict: safe

Notes: local CPU cannot run a meaningful distributed trainer smoke. The hardware-side follow-up is a small distributed run that exercises `EnvironMeterCallback` with tqdm/wandb consuming `step_train_metrics` and `step_env_metrics`.

### DIFF
--- a/tests/trainer/test_trace_callback_metrics.py
+++ b/tests/trainer/test_trace_callback_metrics.py
@@ -1,0 +1,32 @@
+import veomni.trainer.callbacks.trace_callback as trace_callback
+
+
+def test_reduce_training_metrics_coalesces_numeric_values(monkeypatch):
+    calls = []
+
+    def fake_all_reduce(value, group=None):
+        calls.append((value, group))
+        return [item + 1 for item in value]
+
+    monkeypatch.setattr(trace_callback, "all_reduce", fake_all_reduce)
+
+    reduced = trace_callback._reduce_training_metrics({"loss": 2.0, "grad_norm": 3.0}, group="fsdp")
+
+    assert reduced == {"loss": 3.0, "grad_norm": 4.0}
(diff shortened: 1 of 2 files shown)


Coach257 (changes requested):
Agree with the single-element concern: `all_reduce([x])` returns a scalar via `numel()==1` → `.item()`, so `zip(keys, reduced_values)` will TypeError. Current EnvironMeter path usually has ≥2 float metrics so it may not hit today, but the helper should still handle `len==1` (and ideally coerce 0-dim tensors with `.item()` before coalescing). Also, the non-numeric fallback that calls `all_reduce` on every value is unsafe for non-reducible objects — please coalesce only numeric scalars and leave/skip the rest.

PR #954, opened 2026-07-21
[trainer, data] feat: add channel loss data observability

YLlllllllllll:
## Summary

Add loader-agnostic channel-loss observability for causal-LM training:

- per-source detached CE, weighted contribution, and supervised-token counts;
- per-source samples, input tokens, label tokens, and labels per sample;
- stable collision-safe metric keys;
- a bounded final `channel_overview` dashboard;
- an optional opaque, consume-once trace-step hook for dataloader adapters.

The observer does not replace or modify the training loss, objective, RNG flow, or gradients.

## Correctness hardening

- Accumulate multiple causal-loss observations within one model forward instead of overwriting the first result.
- Cache the trainer-owned `ParallelState`; do not depend on ambient parallel state inside callbacks.
- Prepare SP payloads locally, run a world topology preflight, then execute a fixed SP collective sequence outside the conditional loss hook.
- Synchronize strict missing-observation and reduction failures so one rank cannot continue while another raises.
- Keep each strict micro-step atomic: a failed observation does not leave partial metrics.
- Dispatch DPO source repetition only to the channel-loss callback; observe policy-model totals while excluding reference-model totals.
- Add `release_cache` for memory-constrained diagnostic runs. It synchronizes and releases detached CE cache before backward, at an explicit performance cost.

## Public boundary

This PR contains no Byted dataloader, HDFS, parquet locator, sample head, or private corpus implementation. Public and private loaders may pass source metadata and an opaque trace ID; physical locators and content remain adapter-owned.

The dashboard contains aggregate metrics only. Native scalar metrics remain the live, machine-readable contract.

## Verification

- channel callback and dashboard tests: `109 passed`
- trainer, ChunkMBS, and checkpoint regression set: `170 passed`
- `make quality`: passed
- `git diff --check`: passed
- independent VeOmni review: `safe` after two findings were fixed

DPO coverage validates the current BaseTrainer dispatch, source repetition, policy-only totals, and policy/reference `return_log_probs=True` calls. It is not a claim that every model-specific DPO backend has hardware coverage.

## Hardware status

The current head still requires the planned public-backend Ascend 910B4-1 integration test. Older Ascend channel-loss runs predate this SP/fail-closed hardening and are not claimed as validation of this head.

### DIFF
--- a/veomni/trainer/callbacks/channel_loss_callback.py
+++ b/veomni/trainer/callbacks/channel_loss_callback.py
@@ -103,6 +103,7 @@ class ChannelLossComputer:
         self._per_mb_attention_masks: list[torch.Tensor | None] = []
         self._micro_step = 0
         self.step_totals: dict[ChannelKey, tuple[torch.Tensor, torch.Tensor]] = {}
+        self.step_data_totals: dict[ChannelKey, tuple[torch.Tensor, torch.Tensor, torch.Tensor]] = {}
         self.source_names: dict[ChannelKey, str] = {}
         self.strict = False
 
@@ -299,6 +300,7 @@ class ChannelLossComputer:
         self._per_mb_attention_masks = per_mb_attention_masks
         self._micro_step = 0
         self.step_totals = {}
+        self.step_data_totals = {}
 
     def before_micro_step(self) -> None:
         step = self._micro_step
@@ -318,11 +320,23 @@ class ChannelLossComputer:
                 source_id = item["source_id"]
                 loss_sum = item["loss_sum"].detach()
                 token_count = item["token_count"].detach()
+                sample_count = item["sample_count"].detach()
+                input_token_count = item["input_token_count"].detach()
                 previous = self.step_totals.get(source_id)
                 if previous is None:
                     self.step_totals[source_id] = (loss_sum, token_count)
                 else:
                     self.step_totals[source_id] = (previous[0] + loss_sum, previous[1] + token_count)
+
+                data_previous = self.step_data_totals.get(source_id)
+                if data_previous is None:
+                    self.step_data_totals[source_id] = (sample_count, input_token_count, token_count)
+                else:
+                    self.step_data_totals[source_id] = (
+                        data_previous[0] + sample_count,
+                        data_previous[1] + input_token_count,
+                        data_previous[2] + token_count,
+                    )
         self._source_ids = []
         self._position_ids = None
         self._attention_mask = None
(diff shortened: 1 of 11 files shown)


YLlllllllllll:
@Luosuu All CI checks passed on `b6e3e2a` (3 green) — ready to merge when you have a moment.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -579,16 +596,24 @@ def _aggregate_by_source(
...
+                if mask_slice is not None
+                else torch.as_tensor(labels_slice.numel(), device=labels_slice.device, dtype=torch.int64)
             )
+            item = {
+                "source_id": source_ids[i],
+                "loss_sum": loss_sum,
+                "token_count": token_count,
+            }
+            if include_data_stats:
+                item["sample_count"] = torch.ones((), device=labels_slice.device, dtype=torch.int64)
```
FoolPlayer, line 1147: Could we preserve the logical sample count here instead of assigning one sample to every repeated segment? `TextDPOTrainer` calls `on_step_begin(..., source_repeat=2)` because each preference pair produces chosen and rejected causal-LM segments. With `sample_count=1` for both segments, one DPO dataset sample is reported as two samples, and `label_tokens_per_sample` is divided by the number of sequences rather than preference pairs. One option is to propagate a per-segment logical-sample weight when source metadata is repeated (for example, `1` for the first segment and `0` for the second), while still summing input and label tokens across both segments.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -1176,21 +1284,67 @@ def _build_metrics(
...
+        source_names: dict[ChannelKey, str] | None,
+    ) -> dict[ChannelKey, str]:
+        base_names = self._render_base_metric_names(totals, source_names)
+        name_counts: dict[str, int] = defaultdict(int)
+        for name in base_names.values():
+            name_counts[name] += 1
+        return {
+            source_id: (
+                base_names[source_id]
+                if name_counts[base_names[source_id]] == 1
```
FoolPlayer: This makes the metric key depend on which source names have been discovered so far. A source may initially emit `samples/train_a`, then switch to `samples/source-i-1__train_a` when a colliding sanitized name appears on a later step. Monitoring systems will treat those as separate time series, and the earlier unqualified points remain ambiguous. Could we always include the stable source-ID fragment, as the channel-loss metrics already do, or otherwise register the complete source set before the first emission?

FoolPlayer (commented):
I found two correctness issues in the new per-source data metrics: DPO logical samples are double-counted, and data metric keys can change after a later source-name collision.

YLlllllllllll:
@Luosuu All CI checks passed on `a73eab0` (3 green) — ready to merge when you have a moment.

PR #955, opened 2026-07-21
[dist, trainer, ckpt] fix: skip redundant HF weight loading on DCP resume

Luosuu:
## Summary

- skip redundant pretrained-weight materialization when resuming a complete, non-LoRA distributed checkpoint
- keep HF base-weight loading for LoRA/PEFT resumes
- require complete model metadata for full-model restores while retaining partial optimizer-state support
- release transient allocator buffers after checkpoint restore

## Motivation

A cold DCP resume previously loaded and broadcast the HF checkpoint before restoring model and optimizer state from DCP. Those weights were immediately overwritten, adding avoidable GPU memory pressure, allocation churn, and startup latency.

The updated path materializes empty parameters and restores them directly from DCP. Full-model restores validate that every required model key is present, preventing incomplete checkpoints from leaving uninitialized parameters.

## Validation

- focused checkpoint and resume tests: 6 passed
- DCP checkpoint suite: 26 passed, 2 expected failures
- distributed skip-load tests: 2 passed
- DeepSeek-V4 model/optimizer DCP save-load and HF conversion test passed

Allocator profiling with 7.13 GB of BF16 model weights showed:

- parallelization peak allocated: 4,012.75 MiB → 3,563.89 MiB (-448.86 MiB)
- parallelization peak reserved: 4,062 MiB → 3,578 MiB (-484 MiB)
- allocation volume: 27.84 GiB → 3.48 GiB (-87.5%)
- parallelization time: 9.013 s → 0.535 s (-94.1%)

An 8-GPU FSDP2 round-trip using the full DeepSeek-V4 model configuration and deterministic dummy weights also passed:

- model-only DCP size: 530 GiB
- empty-model materialization: 4.24–5.00 s per rank
- DCP save: 46.92–47.06 s per rank
- DCP load: 58.85–58.89 s per rank
- peak allocated memory: approximately 66.22 GiB per rank
- zero mismatches across all parameters and persistent buffers on every rank

Post-DCP memory remained equivalent in the controlled A/B test, as expected: this change removes the redundant cold-start peak rather than changing steady-state training memory.


FoolPlayer:
There is one correctness issue that should be addressed before merging:

1. **Non-persistent buffers are not restored on the skip-HF-load path.** `model.to_empty()` replaces buffer storage as well as parameter storage, but DCP only restores entries present in `state_dict()`. Buffers registered with `persistent=False`—for example the RoPE `inv_freq` / `original_inv_freq` buffers used by several supported models—are excluded from DCP and can therefore remain uninitialized after resume. This path also bypasses the existing buffer restoration/post-processing performed by the normal HF weight loader. Please preserve and restore these buffers, or explicitly reinitialize them, before training starts.

2. **Please add a non-persistent-buffer resume test to CI.** The test should save a full DCP checkpoint, rebuild the model through the new `should_skip_hf_weight_load=True` path, restore the checkpoint, and verify that a known `persistent=False` buffer retains its expected value. Ideally it should also compare a forward result before and after resume so that uninitialized RoPE state is caught as a correctness regression.

Luosuu:
Addressed in e772c8e27. The DCP skip-HF path now snapshots and restores only buffers registered with `persistent=False` around `to_empty()`, so config-derived state such as RoPE `inv_freq` survives without reinitializing parameters or duplicating persistent buffers that DCP restores. Added a regression test that saves a real DCP checkpoint, rebuilds through `should_skip_hf_weight_load=True` using VeOmni meta initialization, verifies the non-persistent buffer is absent from `state_dict()`, restores DCP, and checks exact buffer and forward-output equality. Focused suite: 7 passed.

FoolPlayer (approved):

tests/distributed/test_skip_weights_load.py
```diff
@@ -0,0 +1,100 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+"""Guards for should_skip_hf_weight_load (resume skips HF materialization)."""
```
FoolPlayer: Could we move these tests into `tests/checkpoints/test_dcp_checkpointer.py` under a dedicated section such as `TestSkipHfWeightLoadOnResume`, instead of adding a standalone test file? The behavior being covered is specifically full DCP resume semantics (skip-HF decision, materialization forwarding, real DCP restore, and non-persistent-buffer preservation), and the existing checkpointer test module already owns the related strict/partial-load coverage.

Luosuu: Addressed in 15539c827. All seven skip-HF/full-DCP resume tests now live under `TestSkipHfWeightLoadOnResume` in `tests/checkpoints/test_dcp_checkpointer.py`, and the standalone distributed test file has been removed. The moved section passes 7/7, with Ruff, formatting, compile, and diff checks passing.

PR #957, opened 2026-07-22
[perf, config, docs, trainer, task, ci] feat: add safe NPU profile dump modes

YLlllllllllll:
## What changed

- Keep the user interface small: set `train.profile.enable=true`; select only `train.profile.npu_analysis_mode=offline|async` when needed. `offline` is the default.
- In `offline`, synchronize ranks only around raw Ascend finalization, then analyse, gzip, and upload from a detached sidecar. Upload selection is deterministic: an explicit `VEOMNI_UPLOAD_CMD` environment value, then ModelChef's injected file uploader, then `merlin-cli`. Automatic uploads prefer the current Arnold Trial so assets appear in the selected Trial's JobRun Profiling tab and preserve the robust-run association; JobRun association is the fallback only when no Trial is available. Child-only environment changes leave the parent/training environment unchanged.
- Do not fall back to the `bytedmerlin` base64-in-JSON SDK. If no supported file uploader is available, postprocessing reports the failure and preserves the raw capture; analyse/upload never run in the distributed training critical path.
- Treat NPU profiler initialization, finalization, and cleanup failures as non-fatal observability failures. The affected rank disables further profiling, all ranks still leave the paired barrier, and training continues instead of diverging into a later collective timeout.
- Add `async` as the lightweight official torch_npu alternative using `tensorboard_trace_handler(analyse_flag=True, async_mode=True)`. Daemon processes or older torch_npu versions safely fall back to offline raw capture.
- Rebase profile schedules after resume/hot update and reject deprecated entrypoints that cannot honor the distributed synchronization contract. `VEOMNI_NPU_OFFLINE_MERLIN_UPLOAD=0` or `VEOMNI_NPU_OFFLINE_POSTPROCESS=0` remains an explicit opt-out.

This is the CLA-enabled replacement [PR #957](https://github.com/ByteDance-Seed/VeOmni/pull/957); the previous [PR #950](https://github.com/ByteDance-Seed/VeOmni/pull/950) is closed.

## Why the old path could hang

The torch_npu online implementation submits parsing to a process pool, but with `async_mode=False` it calls `close_pool()`, whose executor shutdown uses `wait=True`. Our representative traces take about 180 seconds to parse. With `rank0_only=true`, leaving that synchronous wait outside a barrier lets peers enter the next HCCL collective roughly three minutes before rank 0; putting it inside a barrier instead stalls every rank for the same period. The new contract leaves only raw finalization (0.28-0.35 seconds in the runs below) in the synchronized offline window. `async` uses the official non-waiting path.

## Validation

- Final verification: 70/70 public focused tests and 68/68 internal-branch focused tests passed, including Trial-first/JobRun-fallback upload association, ambient `VEOMNI_UPLOAD_CMD`, and non-fatal initialization/finalization/cleanup failure cases; Ruff check/format covered 607 public and 604 internal files, and both diff checks passed.
- Final public head: `76c6732dd4ec4932580f6280c189e8cbf616524c` (one commit over `upstream/main`, CLA-enabled author/committer).
- 16x Ascend910B4-64G, Qwen3.5-9B SFT, 2.03 GB raw trace / about 8.99M events:
  - bracket baseline: 7,865.062821 token/s
  - offline: 7,777.402343 token/s (`-1.114555%`); raw finalization 0.280974 s; parser 184.212045 s outside the training critical section
  - async: 7,800.651314 token/s (`-0.818957%`); handler 0.047221 s; finalization 0.364660 s; parser 180.044590 s, of which 158.524 s overlapped training, leaving a 21.831 s process-exit tail
  - baseline drift was `-0.008302%`; neither successful mode produced HCCL errors/timeouts, tracebacks, NaNs, or parser failures
- A retained internal 1x8 Ascend910B4-64G run independently completed 30/30 steps. Over W&B steps 5-30 it processed 819,200 tokens in 263.842916 s (3,104.877755 token/s, `-0.543177%` versus its pre-profile baseline). Raw finalization was 0.350485 s; parsing took 180.257409 s outside the barrier/training thread, with a 64.892838 s teardown wait after training. Its 1,914,933,802-byte JSON compressed to 116,203,920 bytes.
- The unsafe SDK attempt expanded that 116,203,920-byte gzip to 154,938,560 base64 characters and received HTTP 413. The exact same file uploaded successfully through the file-based uploader. After correcting the association to `arnold_trial=382328633, robust_run_id=3`, the JobRun page's exact asset-list request returned HTTP 200 in 13 ms and listed both robust runs 2 and 3; the run-3 asset is previewable. The SDK failure happened after successful training and left raw/artifacts intact.
- A separate internal 16-card validation confirmed that both the offline and async traces were uploadable and previewable as Perfetto assets (178,622,728 and 178,651,831 bytes).

## Usage

```yaml
train:
  profile:
    enable: true
    trace_dir: /tmp/veomni_npu_profile
    rank0_only: true
    npu_analysis_mode: offline  # default; use async as the alternative
```

## Summary by CodeRabbit

- **New Features**
  - Added configurable NPU profiling with offline/asynchronous analysis, detached processing, uploads, and timeout controls.
  - Added NPU trace postprocessing for analysis, compression, copying, and uploading.
  - Added auxiliary metric reporting and MLU device support.
- **Bug Fixes**
  - Improved profiling synchronization, cleanup, validation, and failure recovery.
  - Added MLU-compatible kernel and profiler handling.
- **Documentation**
  - Expanded NPU profiling guidance and updated rotary embedding support.
  - Removed obsolete chunked microbatch configuration documentation.


YLlllllllllll:
@Luosuu All CI checks passed on `76c6732` (3 green) — ready to merge when you have a moment.

veomni/utils/helper.py
```diff
@@ -685,62 +828,195 @@ def create_profiler(
...
+                f"duration_seconds={handler_seconds:.6f} raw_dir={trace_file}"
+            )
         elif IS_CUDA_AVAILABLE:
             p.export_chrome_trace(trace_file)
         logger.info(f"Profiling result saved at {trace_file}.")
 
+        if IS_NPU_AVAILABLE:
+            offline_postprocess_flag = _env_flag(VEOMNI_NPU_OFFLINE_POSTPROCESS)
+            merlin_upload = _should_upload_npu_profile_to_merlin()
+            explicit_upload_cmd = os.getenv("VEOMNI_UPLOAD_CMD")
```
FoolPlayer: [P2] Please distinguish the platform-provided uploader from an explicit user override. `VEOMNI_UPLOAD_CMD` is initialized from `os.getenv("VEOMNI_UPLOAD_CMD")` at module import, and this line reads the same environment variable again. When the platform injects the command through the environment, this branch therefore wins and the `elif VEOMNI_UPLOAD_CMD` platform branch is normally unreachable unless the environment changes after import. As a result, `platform_associated_upload` remains false and the sidecar skips the Trial-first environment normalization, potentially associating the asset with a stale JobRun instead of the selected Trial. Please use separate sources or metadata for the platform command and the user override, or apply platform association when the selected command is the injected default. A test where the environment is already populated before module import would cover the production path.

FoolPlayer (commented):
I found two correctness issues in the upload-selection and profiler-cleanup paths. I left inline comments at each affected location.

veomni/trainer/callbacks/trace_callback.py
```diff
@@ -160,27 +161,192 @@ def on_step_end(self, state: TrainerState, **kwargs) -> None:
...
+
+                    if state.global_step == args.train.profile.end_step:
+                        self.profiler.stop()
+                        self._profiler_stopped = True
+                except Exception as exc:
+                    if not helper.IS_NPU_AVAILABLE:
+                        raise
+                    profile_error = exc
+                    self._profiler_failed = True
+                    self._profiler_stopped = True
```
FoolPlayer: [P2] A `profiler.step()` exception does not guarantee that the underlying profiler completed shutdown. Marking it as stopped here makes `on_train_end()` skip its guarded `stop()` call, which can leave profiler hooks, buffers, or background resources active for the rest of training. Please track failure separately from successful shutdown: disable further `step()` calls after the exception, but still attempt a best-effort `stop()` during cleanup and set `_profiler_stopped` only after that attempt.

tasks/omni/train_omni_model.py
```diff
@@ -446,11 +474,39 @@ def main():
...
+                        profiler.step()
+                        if global_step == args.train.profile.end_step:
+                            profiler.stop()
+                            profiler_stopped = True
+                            # Persistence/upload is owned by create_profiler.on_trace_ready.
+                    except Exception as exc:
+                        if not helper.IS_NPU_AVAILABLE:
+                            raise
+                        profile_error = exc
+                        profiler_stopped = True
```
FoolPlayer: [P2] This has the same cleanup issue as `ProfileTraceCallback`: a failed `profiler.step()` is treated as if `stop()` already completed, so the final cleanup block is skipped. Please keep separate failed/stopped state and perform a guarded best-effort `profiler.stop()` at teardown before marking it stopped.

YLlllllllllll: Fixed in `3602176f` with the same state separation in the standalone omni entrypoint. `profiler_failed` disables subsequent `step()` calls without setting `profiler_stopped`; the existing all-rank teardown barrier then performs a best-effort `profiler.stop()` before marking cleanup complete.

veomni/trainer/callbacks/trace_callback.py
YLlllllllllll: Fixed in `3602176f`. A `profiler.step()` or scheduled `stop()` exception now sets only `_profiler_failed`; it no longer claims `_profiler_stopped`. Further sampling is disabled by checking both states, while `on_train_end()` performs one synchronized best-effort `stop()` and marks the profiler stopped only after that cleanup attempt. The cleanup barrier uses a rank-shared flag, so `rank0_only` cannot create mismatched barriers. Added regressions for both a failed `step()` and a failed scheduled `stop()` followed by teardown retry.

veomni/utils/helper.py
YLlllllllllll: Thanks — this was valid on `76c6732d`. `7b5947ca` restored the two distinct sources, and `3602176f` makes that contract explicit in the code: `os.getenv("VEOMNI_UPLOAD_CMD")` is the user override, while ModelChef injects its platform uploader into the module-level `helper.VEOMNI_UPLOAD_CMD` via `veomni_patch.apply_utils_patch()`. Only the platform branch sets `platform_associated_upload=True`, so the sidecar normalizes to the current Trial and removes stale JobRun identifiers. `test_npu_offline_merlin_upload_precedence` covers platform-vs-user selection, and `test_npu_sidecar_platform_upload_prefers_trial_in_child_env` covers the Trial-first child environment.

YLlllllllllll:
@FoolPlayer Thanks for the careful review. The three inline findings are addressed in `3602176f`:

- uploader selection now explicitly separates the user environment override from ModelChef's module-level platform injection, preserving Trial-first association;
- callback and standalone omni paths now track `failed` separately from `stopped`, disable further sampling after failure, and still perform synchronized best-effort cleanup at teardown;
- focused regressions cover failed `step()` and scheduled `stop()` cleanup retries.

Validation: 72/72 focused tests passed, `make quality` passed across 607 files, and CLA remains green. All three threads have replies and are resolved. Could you please take another look?

YLlllllllllll:

YLlllllllllll:
@cursor Supplemental status against the current head `b84382a`:

- `tests/utils/test_profiler_config.py` and `tests/utils/test_npu_offline_postprocess.py` are now explicitly wired into the GPU workflow.
- NPU `profile_memory` behavior is now explicit: the current head emits a warning and the docs state that Ascend records memory in the torch_npu trace rather than producing a CUDA-style `.pkl` snapshot.
- The upload path now uses an argv list with `subprocess.run(..., check=True)`; there is no remaining `shell=True` in the profiler paths.

The remaining actionable issue is the sidecar wait at train teardown: with `rank0_only`, only profiled ranks wait (up to the configured 300 seconds) before later callbacks can enter collectives. We will keep this thread open for a synchronization fix. The other findings above appear addressed in the current head and only need stale-thread resolution/re-review.

YLlllllllllll:
@cursor The NPU offline profiler teardown fix is now pushed at `40c05a328`:

- synchronize all ranks after the rank-local sidecar wait in both `ProfileTraceCallback` and `tasks/omni/train_omni_model.py`;
- catch/log sidecar-wait failures while still releasing the final barrier;
- keep the cleanup barrier flag set until sidecar wait and synchronization finish;
- add regression coverage for the completed profile window and sidecar-wait failure.

Validation: `50 passed` in `tests/utils/test_profiler_config.py`, scoped Ruff/format checks and `py_compile` pass. Please re-review the sidecar teardown thread.

YLlllllllllll:
Merged current `main` at `b1a4b5170` so this PR is no longer dirty.

`#1082` removed SeedOmni V1 and the deprecated standalone entrypoints. This update:
- keeps the `ProfileTraceCallback` teardown sync from `40c05a3` (sidecar wait + all-rank barrier + failed/stopped separation)
- drops the deleted `tasks/omni/train_omni_model.py` and `tasks/deprecated_task/*` profiler hunks instead of resurrecting them
- updates the profiling doc to say NPU dump modes are trainer-callback only

@cursor please re-review this head. The earlier omni/deprecated-task comments are obsolete after `#1082`.

tests/utils/test_profiler_config.py
```diff
@@ -0,0 +1,1354 @@
...
+    profile = _ProfileCallback()
+    before = _Callback("before")
+    after = _Callback("after")
+    trainer.checkpointer_callback = checkpoint
+    trainer.profile_callback = profile
+    trainer._callbacks = [before, checkpoint, profile, after]
+
+    trainer.on_train_begin()
+
+    assert events == [("before", 0), ("checkpoint", 0), ("profile", 4), ("after", 4)]
```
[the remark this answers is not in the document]
YLlllllllllll: Fixed on `e36f528b8`. The test now calls `BaseTrainer._init_callbacks` with stubbed callback constructors and asserts `profile_callback` is after `checkpointer_callback` (and last in the list), then checks that `on_train_begin` still dispatches checkpoint restore before profile. Also corrected the sidecar wait logs to say `total timeout` / `total wait budget` instead of `each`.

YLlllllllllll:
Addressed the new CodeRabbit notes on `e36f528b8`: the callback-order test now exercises `_init_callbacks`, and sidecar wait logs describe the shared total timeout. Please re-review this head.

PR #959, opened 2026-07-22
[ops, perf] feat: dedicated K==1 outer-product path for deterministic RoPE on H100

zhangxin81:
### What does this PR do?

Adds a dedicated `K == 1` outer-product path to the deterministic batched matmul
Triton kernel used by DeepSeek V3's deterministic RoPE.

The RoPE frequency computation `inv_freq @ position_ids` is a rank-1 product
with `K == 1` (shapes `(B, D/2, 1) x (B, 1, S)`) — a batched **outer product**,
not a real contraction. Routing it through the generic `tl.dot` GEMM padded K
from 1 to 16 (`BLOCK_K=16`), wasting ~15/16 of the tensor-core MMA, and used
lossy TF32.

The new broadcast kernel computes `C[b, m, n] = A[b, m] * B[b, n]`. With no
summation the result is a single fp32 multiply — **bitwise identical** to the
intended GEMM, strictly **more** accurate than the old TF32 `tl.dot` path, and
fully deterministic (no cuBLAS, no split-K). General `K > 1` inputs fall back to
the original deterministic `_bmm_kernel`.

### Checklist Before Starting

- Search for relative PRs/issues and link here: no existing PR touches
  `veomni/ops/kernels/rotary/triton_deterministic.py`.

### Test

Measured on **H100-SXM** (torch 2.9.1+cu129, triton 3.5.1), seqlen 100-200 grid:

- geomean **1.11x** (up to **1.19x**), all shapes improved
- Exact fp32 vs `torch.bmm` (max abs diff **0.0**)

### API and Usage Example

No API change. `triton_bmm(a, b)` transparently dispatches to the outer-product
kernel when `K == 1`; all other shapes are unchanged.

### Design & Code Changes

- Add `_outer_kernel`: batched outer product `C[b, m, n] = A[b, m] * B[b, n]`
  with 3D grid `(B, cdiv(M, BLOCK_M), cdiv(N, BLOCK_N))` and int64 offset
  promotion guarded by a `LARGE` constexpr for >2^31-element tensors.
- Dispatch to `_outer_kernel` for `K == 1` in `triton_bmm`, keeping the original
  `_bmm_kernel` GEMM path for `K > 1`.

- Applied pre-commit checks.
- Correctness (exact fp32 vs `torch.bmm`) and determinism validated on H100 (see
  Test section); no CI GPU coverage for this deterministic kernel path.

### DIFF
--- a/veomni/ops/kernels/rotary/triton_deterministic.py
+++ b/veomni/ops/kernels/rotary/triton_deterministic.py
@@ -15,11 +15,62 @@
 """Deterministic batched matmul Triton kernel used by DeepSeek V3's deterministic RoPE path.
 
 Originally adapted from https://github.com/thinking-machines-lab/batch_invariant_ops.
+
+H100 optimization: the RoPE frequency computation ``inv_freq @ position_ids`` is
+a rank-1 product with ``K == 1`` (``(B, D/2, 1) x (B, 1, S)``), i.e. a batched
+**outer product**, not a real contraction. Routing it through ``tl.dot`` with
+``BLOCK_K=16`` pads K from 1 to 16 and wastes ~15/16 of the tensor-core MMA. For
+the ``K == 1`` case we use a dedicated broadcast (outer-product) kernel:
+``C[b, m, n] = A[b, m, 0] * B[b, 0, n]``. Because there is no summation, the
+result is **bitwise identical** to the GEMM path (a single fp32 multiply) and
+remains fully deterministic (no cuBLAS, no split-K). General ``K > 1`` inputs
+fall back to the original deterministic ``_bmm_kernel``.
 """
 
 import torch
 import triton
 import triton.language as tl
+
+
+@triton.jit
+def _outer_kernel(
+    a_ptr,  # [B, M] (K==1 squeezed)
+    b_ptr,  # [B, N]
+    c_ptr,  # [B, M, N]
+    B,
+    M,
+    N,
+    stride_ab,
+    stride_am,
+    stride_bb,
+    stride_bn,
+    stride_cb,
+    stride_cm,
+    stride_cn,
+    BLOCK_M: tl.constexpr,
+    BLOCK_N: tl.constexpr,
+    LARGE: tl.constexpr,
+):
+    """Batched outer product: C[b, m, n] = A[b, m] * B[b, n]."""
+    pid_b = tl.program_id(0)
+    pid_m = tl.program_id(1)
+    pid_n = tl.program_id(2)
+    if pid_b >= B:
+        return
+    offs_m = pid_m * BLOCK_M + tl.arange(0, BLOCK_M)
+    offs_n = pid_n * BLOCK_N + tl.arange(0, BLOCK_N)
+    mask_m = offs_m < M
+    mask_n = offs_n < N
+    if LARGE:
+        offs_m = offs_m.to(tl.int64)
+        offs_n = offs_n.to(tl.int64)
+
+    a = tl.load(a_ptr + pid_b * stride_ab + offs_m * stride_am, mask=mask_m, other=0.0)
+    b = tl.load(b_ptr + pid_b * stride_bb + offs_n * stride_bn, mask=mask_n, other=0.0)
+    c = a[:, None] * b[None, :]
+
+    c_ptrs = c_ptr + pid_b * stride_cb + offs_m[:, None] * stride_cm + offs_n[None, :] * stride_cn
+    tl.store(c_ptrs, c.to(c_ptr.dtype.element_ty), mask=mask_m[:, None] & mask_n[None, :])
 
 
 @triton.jit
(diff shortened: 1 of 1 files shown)


veomni/ops/kernels/rotary/triton_deterministic.py
```diff
@@ -15,13 +15,64 @@
...
+    offs_n = pid_n * BLOCK_N + tl.arange(0, BLOCK_N)
+    mask_m = offs_m < M
+    mask_n = offs_n < N
+    if LARGE:
+        offs_m = offs_m.to(tl.int64)
+        offs_n = offs_n.to(tl.int64)
+
+    a = tl.load(a_ptr + pid_b * stride_ab + offs_m * stride_am, mask=mask_m, other=0.0)
+    b = tl.load(b_ptr + pid_b * stride_bb + offs_n * stride_bn, mask=mask_n, other=0.0)
+    c = a[:, None] * b[None, :]
```
FoolPlayer, line 70: This is not bitwise-equivalent to the previous Triton GEMM path. The old `tl.dot` uses TF32 input precision for FP32 operands by default on NVIDIA GPUs, whereas this expression performs an IEEE FP32 multiply. On a RoPE-shaped long-position case I observed 258,017 / 262,144 frequency values change. If this is an intentional precision correction, please document the numerical change and add long-position tests against the chosen reference; otherwise the outer-product path needs to preserve the previous TF32 rounding behavior.

veomni/ops/kernels/rotary/triton_deterministic.py
```diff
@@ -106,6 +157,34 @@ def triton_bmm(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
     a = a.contiguous()
     b = b.contiguous()
     c = torch.empty((B, M, N), device=a.device, dtype=a.dtype)
+
+    if K == 1:
+        # Rank-1 outer product (the RoPE case). No contraction => a plain
+        # broadcast multiply, bitwise-identical to the GEMM and deterministic.
+        BLOCK_M = triton.next_power_of_2(M)
```
FoolPlayer, line 164: `BLOCK_M` is unbounded here. For example, `M=1025, N=1024, K=1` selects a `2048 x 1024` tile, and the materialized `c` tensor exceeds Triton's 1,048,576-element tensor limit, causing compilation to fail even though the previous tiled path handles the shape. Please cap `BLOCK_M` (for example at 128), rely on the existing M grid for tiling, and add this shape as a regression test.

veomni/ops/kernels/rotary/triton_deterministic.py
```diff
@@ -15,13 +15,64 @@
...
 
 Originally adapted from https://github.com/thinking-machines-lab/batch_invariant_ops.
+
+H100 optimization: the RoPE frequency computation ``inv_freq @ position_ids`` is
+a rank-1 product with ``K == 1`` (``(B, D/2, 1) x (B, 1, S)``), i.e. a batched
+**outer product**, not a real contraction. Routing it through ``tl.dot`` with
+``BLOCK_K=16`` pads K from 1 to 16 and wastes ~15/16 of the tensor-core MMA. For
+the ``K == 1`` case we use a dedicated broadcast (outer-product) kernel:
+``C[b, m, n] = A[b, m, 0] * B[b, 0, n]``. Because there is no summation, the
+result is **bitwise identical** to the GEMM path (a single fp32 multiply) and
```
FoolPlayer, line 25: This claim is inverted, and it undersells the PR. For fp16/bf16 the two paths *are* bitwise identical; fp32 is where they differ — and fp32 is the only dtype that reaches `K == 1`, since `device_patch.py` calls `.float()` on both operands. `tl.dot` defaults to TF32 and truncates fp32 to an 11-bit significand, so `position_ids` above ~2048 are not representable.

Against an fp64 reference with DSv3 rope params (dim=64, base=10000), cos max/mean error on main: `0.076 / 0.0024` at S=200, `0.763 / 0.025` at S=2048, **`2.000 / 0.134`** at S=8192. A cos error of 2.0 is the full range of the cosine. `configs/text/deepseek.yaml` defaults to `max_seq_len: 8192`, so this is a real accuracy bug on main for anyone running `rotary_pos_emb_implementation: triton` — and this PR fixes it.

Suggest retitling as `fix` and leading with the accuracy story; it matters far more than the 1.11x.

veomni/ops/kernels/rotary/triton_deterministic.py
```diff
@@ -106,6 +157,34 @@ def triton_bmm(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
     a = a.contiguous()
     b = b.contiguous()
     c = torch.empty((B, M, N), device=a.device, dtype=a.dtype)
+
+    if K == 1:
+        # Rank-1 outer product (the RoPE case). No contraction => a plain
+        # broadcast multiply, bitwise-identical to the GEMM and deterministic.
+        BLOCK_M = triton.next_power_of_2(M)
+        BLOCK_N = min(triton.next_power_of_2(N), 1024)
```
FoolPlayer, line 165: `BLOCK_N=1024` gives 8x fewer CTAs than the old path's `BLOCK_N=128`. Pure kernel time (CUDA-graph captured, M=32 fp32): 1.11x at S=200, then 0.82x at 512, **0.58x at 2048-8192**, 0.71x at 32768. The 1.11x geomean only holds inside the benchmarked 100-200 window, while every `max_seq_len` in `configs/` is 2048-8192.

The absolute cost is ~1.4us once per model forward, so the practical impact is nil either way — but `min(triton.next_power_of_2(N), 128)` would reproduce the old grid and keep the win at all lengths. Worth setting `num_warps` explicitly too rather than inheriting 4.

veomni/ops/kernels/rotary/triton_deterministic.py
FoolPlayer, line 164: `BLOCK_M` is uncapped while `BLOCK_N` is capped at 1024. At M=1024, N=1024 the tile is 1M elements and compiles with `n_spills=4948`, running 16x slower than the fallback; M=4096 fails to compile outright.

Also `triton.next_power_of_2(0) == 0`, so `M == 0` or `N == 0` now raises `ZeroDivisionError` where the old path returned an empty tensor. Both are latent (the sole caller is M=32), but `triton_bmm` is in `__all__`. Suggest a joint bound like `BLOCK_M * BLOCK_N <= 8192` and an early return when `c.numel() == 0`.

FoolPlayer (commented):
Reviewed on GB200 (Blackwell sm_100, torch 2.9.1 / triton 3.5.1). The outer-product path is the right call, but I think the PR is framed around the wrong benefit — see the docstring comment. There is a real accuracy fix buried in here that deserves the headline.

PR #962, opened 2026-07-23
[model] feat: Deepseek V4 export fp8/fp4 quant weights

wuxibin89:
### What does this PR do?

1. Support HDFS for model_path
2. Support Deepseek V4 export fp8/fp4 quant weights

### DIFF
--- a/veomni/models/checkpoint_tensor_loading.py
+++ b/veomni/models/checkpoint_tensor_loading.py
@@ -28,10 +28,13 @@ load when an index mapping is supplied).
 
 import json
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Callable, Dict, List, Optional, Protocol, Union
+from typing import TYPE_CHECKING, Callable, Dict, Generator, List, Optional, Protocol, Union
 
 import torch
 from torch import nn
+from torch.distributed.tensor import DTensor
+
+from veomni.distributed import parallel_state
 
 from ..utils import logging
 
@@ -209,6 +212,7 @@ def prepare_fqn_to_index_mapping_for_model(
     prepared = maybe_convert_fqn_to_index_mapping(fqn_to_index_mapping, model)
     if prepared is not None:
         model._veomni_prepared_fqn_to_index_mapping = prepared
+        model._veomni_fqn_to_index_mapping = fqn_to_index_mapping
     return prepared
 
 
@@ -243,3 +247,58 @@ def maybe_convert_checkpoint_tensor(
     if not converter.can_handle(name):
         return ConvertedCheckpointTensor(name=name, tensor=tensor)
     return converter.convert(name, tensor)
+
+
+def _map_moe_params_common(name, tensor, ep_rank):
+    num_experts_per_rank = tensor.size(0)
+    for i in range(num_experts_per_rank):
+        idx = ep_rank * num_experts_per_rank + i
+        new_key = name.replace("mlp.experts.", f"mlp.experts.{idx}.") + ".weight"
+        yield new_key, tensor[i].to(tensor.device, non_blocking=True)
+
+
+def _process_moe_params(name, tensor, ep_rank):
+    if "gate_up_proj" in name:
+        gate, up = tensor.chunk(2, dim=1)
+        params = {
+            name.replace("gate_up_proj", "gate_proj"): gate,
+            name.replace("gate_up_proj", "up_proj"): up,
+        }
+    else:
+        params = {name: tensor}
+
+    for key, value in params.items():
+        yield from _map_moe_params_common(key, value, ep_rank)
+
+
+def export_weights(model: torch.nn.Module) -> Generator[tuple[str, torch.Tensor], None, None]:
+    """
+    Export master weights from fully_sharded model, fused MoE weights are split into per-expert weights.
+
+    Args:
+        model: The fully_sharded model to export the master weights from.
+
+    Returns:
+        A generator of tuples of the form (name, tensor).
+    """
+    ps = parallel_state.get_parallel_state()
+    params = model.state_dict()
+    for name, param in params.items():
+        unsharded_tensor = param.full_tensor() if isinstance(param, DTensor) else param
+
+        is_expert_layer = "mlp.experts." in name
+        is_proj = any(p in name for p in ["down_proj", "gate_proj", "up_proj", "gate_up_proj"])
+
+        if is_expert_layer and is_proj and ps.ep_enabled:
+            ep_rank, ep_size = ps.ep_rank, ps.ep_size
+            buffer = torch.empty_like(unsharded_tensor)  # [num_experts/ep_size, H, I]
+            for src_ep_rank in range(ep_size):
+                tensor = unsharded_tensor if src_ep_rank == ep_rank else buffer
+                torch.distributed.broadcast(tensor, group_src=src_ep_rank, group=ps.ep_group)
+                yield from _process_moe_params(name, tensor, ep_rank=src_ep_rank)
+
+        else:
+            if is_expert_layer:
+                yield from _process_moe_params(name, unsharded_tensor, ep_rank=0)
+            else:
+                yield name, unsharded_tensor
--- a/veomni/models/checkpoint_tensor_loading.py
+++ b/veomni/models/checkpoint_tensor_loading.py
@@ -37,6 +37,7 @@ from torch.distributed.tensor import DTensor
 from veomni.distributed import parallel_state
 
 from ..utils import logging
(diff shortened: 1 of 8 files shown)


veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
```diff
@@ -328,6 +378,53 @@ def finalize(self) -> list[ConvertedCheckpointTensor]:
         self._scaled_weight_buffer.clear()
         return finalized
 
+    def export_weights(self, model: torch.nn.Module) -> Generator[tuple[str, torch.Tensor]]:
```
FoolPlayer, line 381: I could not find a save or consolidation path that calls this method; the existing HF save flow still consumes `ModelState.state_dict()`. As a result, the expert splitting, reverse renaming, and FP8/FP4 quantization added here are currently unreachable. Please wire the converter into the distributed save path and add an export -> save -> reload integration test.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
```diff
@@ -328,6 +378,53 @@ def finalize(self) -> list[ConvertedCheckpointTensor]:
...
+        config = model.config
+        fqn_to_index_mapping = model._veomni_fqn_to_index_mapping
```
FoolPlayer, line 383: This attribute is not guaranteed to exist. It is only installed when checkpoint index preparation receives a non-`None` mapping; single-file checkpoints and full DCP resumes that skip HF weight loading can therefore raise `AttributeError` here. Please pass the original mapping explicitly from the save API, or support/reconstruct a `None` mapping for single-shard export.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
```diff
@@ -328,6 +378,53 @@ def finalize(self) -> list[ConvertedCheckpointTensor]:
...
+            if scale_name is None or scale_name not in fqn_to_index_mapping:
+                export_weight_names.add(origin)
+                yield origin, param
+                continue
+
+            # 4. quantize the weight
+            fp4_quantize = "ffn.experts." in origin and expert_dtype == "fp4"
+            param = param.to(torch.bfloat16)
+            if fp4_quantize:
+                weight, scale = fp4_act_quant(param, block_size=32, inplace=False)
```
FoolPlayer, line 413: `fp4_act_quant()` returns `torch.float4_e2m1fn_x2` with a half-width last dimension, while the official V4-Pro checkpoints store packed FP4 bits as `int8`, and the current VeOmni loader only recognizes the packed layout when the dtype is `int8`. Reloading this export therefore treats it as an ordinary half-width weight and produces a shape mismatch. Please export the packed bits as `int8` (or extend the loader to support both representations) and add an FP4 save/load round-trip test.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
```diff
@@ -328,6 +378,53 @@ def finalize(self) -> list[ConvertedCheckpointTensor]:
...
+            else:
+                weight, scale = act_quant(param, block_size=128, scale_fmt="ue8m0", scale_dtype=torch.float8_e8m0fnu)
```
FoolPlayer: `act_quant()` implements activation-style row-wise blocks, producing scales shaped `[out, in / 128]`. DeepSeek weight checkpoints require 2-D `128 x 128` weight blocks with scales shaped `[ceil(out / 128), ceil(in / 128)]`; for a `[1024, 4096]` weight this emits `[1024, 32]` instead of `[8, 32]`. Please implement the official 2-D weight quantization layout and validate it against the inference GEMM.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
```diff
@@ -328,6 +378,53 @@ def finalize(self) -> list[ConvertedCheckpointTensor]:
...
+
+            export_weight_names.add(origin)
+            export_weight_names.add(scale_name)
+            yield origin, weight
+            yield scale_name, scale
```
FoolPlayer, line 423: The tensors yielded by this path can contain `F8_E8M0` scales and `F4` weights, but the current DCP consolidation metadata parser has no mapping for either dtype and raises `KeyError`. Please extend consolidation dtype/element-size handling before integrating this export and add distributed consolidation coverage for both formats.

veomni/models/checkpoint_tensor_loading.py
```diff
@@ -243,3 +247,58 @@ def maybe_convert_checkpoint_tensor(
...
         return ConvertedCheckpointTensor(name=name, tensor=tensor)
     return converter.convert(name, tensor)
+
+
+def _map_moe_params_common(name, tensor, ep_rank):
+    num_experts_per_rank = tensor.size(0)
+    for i in range(num_experts_per_rank):
+        idx = ep_rank * num_experts_per_rank + i
+        new_key = name.replace("mlp.experts.", f"mlp.experts.{idx}.") + ".weight"
+        yield new_key, tensor[i].to(tensor.device, non_blocking=True)
```
FoolPlayer, line 258: `.to(tensor.device)` does not copy when the tensor is already on that device. For remote EP ranks, these yielded slices therefore alias the reusable broadcast buffer, and the next broadcast can overwrite weights that were yielded earlier (for example when the generator is materialized). Please clone remote slices or guarantee synchronous write-out before buffer reuse, and add a multi-rank test with distinct values per expert/rank.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
wuxibin89: `export_weights` is called by verl to sync weights to rollout.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
wuxibin89: Add `fp8_weight_quant` to support 128x128 block-wise quant.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
wuxibin89: Cast `torch.float4_e2m1fn_x2` to `int8`

veomni/models/checkpoint_tensor_loading.py
wuxibin89: For engine side, it's safe to yield view tensor, verl side will clone the tensor into bucket.

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
wuxibin89: `_veomni_fqn_to_index_mapping` is needed for sanity check when export_weights to rollout, so verl side will guarantee to pass `fqn_to_index_mapping` when build_parallelize_model

veomni/models/transformers/deepseek_v4/checkpoint_tensor_converter.py
```diff
@@ -328,6 +378,56 @@ def finalize(self) -> list[ConvertedCheckpointTensor]:
...
+            if origin.endswith(".weight"):
+                scale_name = origin.removesuffix(".weight") + ".scale"
+
+            if scale_name is None or scale_name not in fqn_to_index_mapping:
+                export_weight_names.add(origin)
+                yield origin, param
+                continue
+
+            # 4. quantize the weight
+            fp4_quantize = "ffn.experts." in origin and expert_dtype == "fp4"
```
FoolPlayer, line 410: ffn for sft model is fp4 but fp8 for base model. We need apply both of these case

wuxibin89: Yes, we distinguish base and sft model by `config. expert_dtype`:
- DeepSeek-V4-Flash-Base: expert_dtype=fp8
   - moe experts: (weight, scale) = (torch.float8_e4m3fn, torch.float8_e8m0fnu)
   - linear layers: (weight, scale) = (torch.float8_e4m3fn, torch.float8_e8m0fnu)
- DeepSeek-V4-Flash: expert_dtype=fp4
  - moe experts: (weight, scale) = (torch.int8, torch.float8_e8m0fnu)
  - linear layers: (weight, scale) = (torch.float8_e4m3fn, torch.float8_e8m0fnu)

FoolPlayer (approved):

PR #964, opened 2026-07-23
[ops, dist, config, docs, ci] fix: make chunk loss compatible with activation offload

YLlllllllllll:
### What does this PR do?

> Make `chunk_loss` / `npu` cross-entropy work with activation offload.
>
> Previously, `ChunkLoss` used `torch.func.grad_and_value`, which rejects any
> active `saved_tensors_hooks` — including the hooks installed by
> `enable_activation`. Enabling both paths failed closed.
>
> This PR replaces that transform with an equivalent first-order
> `torch.autograd.grad` VJP, keeps the short-lived chunk CE graph on device,
> and lets only the long-lived precomputed grads go through the outer offload
> hooks. An opt-in `cross_entropy_loss_release_cache` flag can reclaim the
> transient CE allocator cache before model backward on memory-constrained
> profiles.

### Checklist Before Starting

- Search for relative PRs/issues and link here: none found on GitHub main; extracted from internal NPU/act-offload bring-up (kept separate from Ascend 910B4 Mojo overlay work)
- PR title follows `[{modules}] {type}: {description}` format

### Test

> - New unit tests in `tests/ops/test_chunk_loss_saved_tensor_hooks.py`:
>   - chunk loss under synthetic saved-tensor hooks preserves loss/grads
>   - production `chunk_loss_function` under `custom_save_on_cpu`
>   - opt-in `release_cache` runs only after outer pack completes
>   - factory binds `cross_entropy_loss_release_cache`
> - `python3 -m py_compile` on touched Python files
> - Validated in internal NPU training profiles that combine
>   `chunk_loss`/`npu` with `train.accelerator.offload_config.enable_activation`

### API and Usage Example

```yaml
model:
  ops_implementation:
    cross_entropy_loss_implementation: chunk_loss   # or npu
    cross_entropy_loss_release_cache: false          # set true on tight HBM profiles
train:
  accelerator:
    offload_config:
      enable_activation: true
```

### Design & Code Changes

1. **`veomni/ops/kernels/cross_entropy/chunk_loss.py`**
   - Drop `torch.func`; compute per-chunk VJP with `torch.autograd.grad`
   - Shadow outer activation-offload hooks with identity hooks for the
     short-lived CE graph so vocab-sized logits are not forced through CPU
   - After `ChunkLoss.apply`, optionally sync + `empty_cache` when
     `release_cache=true`
2. **`veomni/ops/kernels/cross_entropy/__init__.py`**
   - Thread `cross_entropy_loss_release_cache` from ops config into the
[...]

PR #872, opened 2026-07-01
[perf] feat: Add Aysnc Activation Offload

Zhang1Sheng:
### What does this PR do?

> Add async activation offload — a memory optimization that offloads hidden states from GPU to CPU during the forward pass and prefetches them back during the backward pass, using dedicated CUDA/NPU streams to overlap transfers with computation. Combined with gradient checkpointing (recomputation for intermediate activations), this enables training with significantly reduced peak GPU memory with minimal throughput impact.

Key design decisions:

- **Class-level `__call__` patching**: places `async_save_on_cpu` **outside** the `GradientCheckpointingLayer` checkpoint boundary so that hidden states saved by `_NoopSaveInputs.apply` pass through the offload hook, while intermediate activations are handled by GC recomputation — yielding zero persistent GPU memory for both.
- **Auto-discovery from `model._no_split_modules`**: mirrors the FSDP sharding pattern — offload targets are the same transformer block classes used for FSDP boundaries, requiring zero extra configuration.
- **Single-stream architecture with prefetch**: uses one swap stream for both D2H and H2D; during backward, launches H2D for layer N-1's tensors while computing layer N's backward, overlapping transfers with computation.

### Checklist Before Starting

- Search for relative PRs/issues and link here: ...
- PR title follows `[{modules}] {type}: {description}` format (enforced by [check_pr_title.yml](.github/workflows/check_pr_title.yml))
  - **Allowed modules:** `agent`, `ci`, `ckpt`, `config`, `data`, `dist`, `docker`, `docs`, `logging`, `lora`, `misc`, `model`, `omni`, `optim`, `ops`, `parallel`, `perf`, `release`, `task`, `trainer`
  - **Allowed types:** `feat`, `fix`, `refactor`, `chore`, `test`
  - Breaking changes: prepend `[BREAKING]` — e.g. `[BREAKING][parallel, model] feat: dynamic batching`

### Test

- [x] Unit tests in `tests/distributed/test_async_activation_offload.py`:
  - **Core components**: `SwapTensor` lifecycle (D2H/H2D state machine), `GetCnt` key generation (unique keys across blocks, multi-pass handling), `OffloadManager` singleton behavior, `async_save_on_cpu` pack/unpack round-trip correctness
  - **Module patching**: `get_offload_modules_from_class_names` auto-discovery, `async_offload_modules` per-instance attribute injection (`_veomni_offload_layer_idx`, `_veomni_offload_depth`), class-level patching idempotency
[...]

PR #965, opened 2026-07-23
[dist, trainer, config, docs] feat: add MindSpeed-style async activation offload

YLlllllllllll:
## Summary

Port MindSpeed-MM-style asynchronous activation offload into VeOmni: stream-based D2H/H2D activation swaps with prefetch, wired through `OffloadConfig` and `BaseTrainer`.

This complements the synchronous activation-offload path and the chunk-loss compatibility work in #964. Async offload applies per-module hooks before FSDP2 sharding so the offload context sits outside gradient-checkpoint boundaries.

## What changed

- Add the async swap manager, saved-tensor hooks, and prefetch flow.
- Add device-agnostic stream/event helpers and module-pattern matching.
- Add `enable_async_activation` and `activation_offload_modules`.
- Apply async hooks before FSDP2; retain null global offload contexts in async mode.
- Fail closed when synchronous and asynchronous activation offload are both enabled.
- Fail closed when configured async patterns match no model modules.
- Add documentation and CPU unit coverage for config validation, pattern expansion, unmatched patterns, and key reuse.

## Usage

```yaml
train:
  accelerator:
    offload_config:
      enable_async_activation: true
      activation_offload_modules:
        - model.layers.{*}
```

`enable_activation` and `enable_async_activation` are mutually exclusive. Every configured async run must match at least one model module.

## Verification

- `python -m py_compile` on the changed Python modules
- `ruff check` on the changed Python modules and tests
- `pytest -q tests/distributed/test_async_offload_unit.py` (6 passed)

Hardware throughput and numerical validation are tracked separately from this CPU/unit gate.

## Scope

This PR does not include internal-only upload code, private NPU/Mojo overlays, or FSDP2 parameter-state CPU offload.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous activation offload for CUDA and NPU devices to reduce accelerator memory usage.
  * Added configurable module selection, host-memory cache limits, validation, and automatic cache management.
  * Integrated asynchronous offload with training and gradient checkpointing.
  * Added device-agnostic stream and event support.

* **Documentation**
  * Expanded activation offload guidance, including supported tensors, cache behavior, reset handling, and compilation limitations.

* **Tests**
  * Added comprehensive GPU and NPU coverage for asynchronous activation offload.


YLlllllllllll:
@Luosuu All CI checks passed on `b8e71f1` (3 green) — ready to merge when you have a moment.

Coach257 (commented):
The new training_context.py / singleton.py / module_match.py are currently only used by async offload (TrainingContext is written but never read). Please keep them private inside async_offload.py until a second real consumer appears.

veomni/utils/module_match.py
```diff
@@ -0,0 +1,24 @@
...
+
+import fnmatch
+
+
+def module_name_match(pattern: str, name: str) -> bool:
+    """Check whether a module name matches a pattern.
+
+    Supports Unix shell-style wildcards via ``fnmatch`` (e.g. ``model.layers.*``).
+    """
+    return fnmatch.fnmatch(name, pattern)
```
Coach257: `fnmatch.fnmatch` treats `*` as matching any characters including `.`, so a documented pattern like `model.layers.*` also matches nested modules such as `model.layers.0.self_attn` / `model.layers.0.mlp`. That can attach async-offload hooks to unintended submodules and break the expected sequential layer order. Could we use segment-aware matching (e.g. `*` → `[^.]+`), or recommend/`{*}`-only expansion for decoder stacks?

veomni/distributed/async_offload.py
```diff
@@ -0,0 +1,447 @@
...
+            self._block_idx = block_idx
+
+        offload_tensor_key = "{}_{}".format(self._block_idx, self._block_tensor_nums[self._block_idx] - 1)
+        return offload_tensor_key, after_block
+
+    def reset(self):
+        self._block_idx = -1
+        self._block_tensor_nums = {}
+
+    def get_prefetch_keys(self, block_idx, tensor_idx):
```
Coach257: This computes `prefetch_block_idx` as the previous offloaded layer, but then ignores it: keys are built with `block_idx - 1` and the tensor count of the *current* layer (`self._block_tensor_nums[block_idx]`). When layer indices are non-contiguous or tensor counts differ across layers, prefetch looks up the wrong keys and the async H2D overlap silently degrades. Prefer generating keys from `prefetch_block_idx` and `self._block_tensor_nums[prefetch_block_idx]`.

veomni/distributed/async_offload.py
```diff
@@ -0,0 +1,447 @@
...
+            if key in self.items.keys():
+                layer_items_keys.append(key)
+        return layer_items_keys
+
+    def get(self, key):
+        self.assert_exist(key)
+        item = self.items[key]
+
+        act = item.act
+        if item.event is not None:
```
Coach257: `OffloadItem` only stores `self.event` and has no `get_event()` method. This branch is currently unreachable because `put()` always passes `event=None`, but if a caller starts supplying an event this will raise `AttributeError`. Please use `item.event.wait()` (or drop the unused `event` API if it is not needed).

Coach257 (changes requested):

YLlllllllllll:
@Coach257 Thanks for the review. Addressed the feedback in `f9cc2c8b`: moved the single-consumer matching/singleton helpers into `async_offload.py` and removed the write-only `TrainingContext`; made glob matching segment-aware so `model.layers.*` cannot match nested submodules; fixed prefetch keys to use the actual previous offloaded layer and its tensor count; and removed the unused event API that could call a nonexistent method. I also added CPU regression coverage for segment-aware matching and non-contiguous/different-count prefetch layers. Could you please take another look?

Coach257:
Suggestion: when `enable_async_activation=True` and `activation_offload_modules` is empty, default to auto-discovering offload targets from `model._no_split_modules` (same class boundaries used for FSDP sharding), instead of requiring an explicit pattern list. Keep the current fail-closed behavior if `_no_split_modules` is missing or yields no matches. That would make the common case config-free while still allowing explicit `activation_offload_modules` (e.g. `model.layers.{*}`) for finer control.

YLlllllllllll:
@Coach257 Thanks for the suggestion. Addressed in `f395fdee`: when `enable_async_activation=True` and `activation_offload_modules` is empty, async offload now discovers targets by matching module class names from `model._no_split_modules`, following the same boundary metadata used by FSDP. Explicit `activation_offload_modules` still takes precedence for finer control. Auto-discovery fails closed when `_no_split_modules` is missing/empty or when its class names match no modules. I also updated the config/docs and added CPU unit coverage for the success and both fail-closed paths. Could you please take another look?

Coach257 (approved):

YLlllllllllll:
@Coach257 I addressed the correctness and lifecycle issues raised by the latest Cursor review in `b5017869` and force-pushed the branch:

- shared-storage views are now rejected from async offload, and storage handling uses `untyped_storage()` with an explicit private/dense-storage invariant;
- the original decoder class is no longer mutated globally: only selected module instances are rebound to an async-offload subclass;
- `OffloadManager` is now scoped per model/application rather than process-wide, abandoned autograd graphs do not stay alive through manager references, and forward exceptions reset manager state;
- the transfer state machine now distinguishes D2H in flight, host-resident, and H2D in flight;
- added regression coverage for instance/model isolation, exception recovery, shared-storage views, and a CUDA-gated offload on/off output + gradient parity test.

I did not add a custom pinned-buffer pool without profiling data because PyTorch already uses a caching pinned-memory allocator; I would prefer to measure allocation/throughput behavior before adding another allocator layer. #965 is the implementation we are carrying forward; #964/#994 can be rebased afterward if needed.

Could you please review the updated implementation?

Coach257:
Please fix the device-check.

YLlllllllllll:

YLlllllllllll:
@Coach257 @cursor Supplemental current-head check for `3e94ca9`:

I reran the repository device-API checker. It still fails on `veomni/distributed/async_offload.py` because of the direct `"cuda"` device literal, and on `tests/distributed/test_async_offload_unit.py` because of direct `.cuda()` / `torch.cuda.*` calls. This is still a required fix before requesting final approval.

The latest head does include the other Cursor hardening: private/dense storage checks, pooled host buffers, per-model manager reset/lifecycle handling, instance-local patching, and CUDA numerical/peak-memory parity coverage. The open overlap with [PR #872](https://github.com/ByteDance-Seed/VeOmni/pull/872) also still needs an explicit merge decision before either implementation is merged. Please re-review after the device-check and overlap decision are addressed.

YLlllllllllll:
@Coach257 @cursor Please re-review the current head `f679aaab8`.

This branch is now synced with `main@ea1e68da3`. The remaining review findings have been addressed:

- the repository's exact device-API checker passes for both `veomni/` and `tests/`; the async path no longer contains direct CUDA device APIs;
- host buffers now use a thread-safe, per-model pool keyed by shape/stride/dtype, with a configurable bounded free-buffer cache (`4 GiB` by default, `0` disables reuse), layout-LRU eviction, and explicit in-flight-memory semantics;
- shared-storage views remain fail-closed, selected modules use instance-local subclasses, and managers reset at step boundaries while abandoned graphs are weakly held;
- tests now include CUDA/NPU-gated output/gradient parity and peak allocated-memory checks, bounded-cache/reuse/eviction coverage, plus a two-process `BaseTrainer -> FSDP2 -> gradient checkpointing -> async offload -> backward` integration path.

Local validation: async-offload unit suite `22 passed, 2 skipped`; controlled two-process FSDP2/GC integration passed; targeted Ruff/format/diff checks and the full device-API CI checker passed. GitHub checks are `3/3` green (auto-label, CLA, Read the Docs).

All six obsolete inline threads were resolved after verifying their fixes. Current `main` still has no equivalent async activation-offload implementation; #872 remains unchanged, and the maintainer note on #872 already points to #965 as the implementation to carry forward.

Could you please take another look?

YLlllllllllll:
**Merge takeaway vs `#872`:** keep this PR.

`#872` (`Zhang1Sheng`, `async_offloading.py`) is the earlier MindSpeed-style port. This PR supersedes it:
- device-API checker clean (Coach's remaining request on `3e94ca9` is fixed on `f679aaab`)
- auto-discover from `model._no_split_modules` when the pattern list is empty
- instance-level subclassing instead of mutating the decoder class globally
- bounded host-buffer pool, shared-storage rejection, segment-aware matching
- unit tests wired into GPU/NPU CI

Please merge `#965` and close `#872` as superseded. Coach257's APPROVED is still on `ce44a346`; a re-approval of `f679aaab` would unblock merge.

@Coach257 @cursor

veomni/distributed/async_offload.py
```diff
@@ -0,0 +1,624 @@
...
+            get_current_stream().wait_event(swap_tensor.h2d_event)
+            swap_tensor.release_host_buffer()
+            swap_tensor.stat = "device"
+
+            block_idx, tensor_idx = swap_tensor.key.split("_")
+            manager.clear(swap_tensor.key)
+
+            if prefetch:
+                manager.prefetch_get(int(block_idx), h2d_stream, d2h_stream)
+            return swap_tensor.tensor
```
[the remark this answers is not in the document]
YLlllllllllll: Fixed in `cee847446`. `_unpack_swap_tensor` now calls `manager.clear` / `prefetch_get` only when `manager.exist(swap_tensor.key)` is true, so a second unpack (retain_graph / repeated saved-tensor access) returns the restored tensor instead of raising. Added `test_unpack_swap_tensor_survives_repeated_access`.

veomni/distributed/async_offload.py
```diff
@@ -0,0 +1,643 @@
...
+    if isinstance(tensor, torch.nn.parameter.Parameter) or tensor._base is not None:
+        return False
+    storage_nbytes = tensor.untyped_storage().nbytes()
+    tensor_nbytes = tensor.numel() * tensor.element_size()
+    # Expandable CUDA segments (PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True)
+    # round device allocations up.  A contiguous tensor with ``_base is None``
+    # still uniquely owns that storage; extra trailing bytes are padding, not a
+    # view, so rejecting ``storage_nbytes != tensor_nbytes`` silently disables
+    # offload on the L20 GPU CI image.
+    if storage_nbytes <= 0 or storage_nbytes < tensor_nbytes or not tensor.is_contiguous():
```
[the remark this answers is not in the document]
YLlllllllllll: Agreed — `Stream.wait_event` only orders later GPU work, and `resize_(0)` frees storage on the host thread. `wait_d2h_finished` now host-synchronizes `d2h_event` before shrinking storage, and still keeps the stream wait. Pushed in `c8c628e83`.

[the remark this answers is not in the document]
YLlllllllllll: Kept expandable-segment padding for privately owned contiguous tensors (`_base is None`, `storage_offset() == 0`); that is what GPU CI actually hands us. Views / non-zero offsets are now rejected via `_has_private_dense_storage`. Autograd saved activations are not `set_()` aliases, and cloning every candidate would cancel the memory win. Pushed in `c8c628e83`.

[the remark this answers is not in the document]
YLlllllllllll: Good catch — `_base is None` + offset 0 still allows a `Tensor.set_()` alias on the same StorageImpl, and `resize_(0)` would free it for every holder.

`wait_d2h_finished` now `set_()`s this tensor onto a fresh empty placeholder instead of resizing the original storage. H2D restores from a new allocation. A padded offset-zero `set_()` alias is covered by `test_wait_d2h_finished_preserves_offset_zero_set_alias`.

Pushed in `c137b56dd`.

veomni/trainer/base.py
```diff
@@ -652,6 +676,8 @@ def on_epoch_end(self):
             callback.on_epoch_end(self.state)
 
     def on_step_begin(self, micro_batches=None, **kwargs):
+        if self.args.train.accelerator.offload_config.enable_async_activation:
```
Coach257: Mv this to `train_step`. `on_xxx_xxx` only controls callbacks.

YLlllllllllll: Moved. `on_step_begin` is callback-only again; `reset_async_activation_offload` now runs at the start of every `train_step` (Base/Text/VLM/DPO/DiT).

PR #966, opened 2026-07-23
[ops, docs, ci] fix: support Ascend 910B4-1 GDR training

YLlllllllllll:
### What does this PR do?

Adds an Ascend 910B4-1 launch profile for the public Qwen3.5
GatedDeltaNet NPU backend introduced in #879.

The hidden-state recurrence currently fixes its value tile at `BV=128`.
At Qwen3.5 training shapes that tile exceeds the 910B4-1 unified-buffer
limit. This PR selects `BV=64` for the raw `Ascend910B4` and
`Ascend910B4-1` device names in both forward and backward. Other Ascend
devices retain `BV=128`.

This PR is intentionally limited to the public `triton-ascend` backend. It
does not add private kernel packages, data loaders, or offload changes.

### Checklist Before Starting

- Related PR: #879
- PR title follows `[{modules}] {type}: {description}`.

### Test

- `pytest -q tests/ops/test_gdn_ascend_gate.py` — 19 passed
- `python -m py_compile scripts/npu/validate_910b4_gdn_training.py veomni/ops/kernels/gated_delta_rule/npu_hardware.py`
- `make quality`
- `git diff --check`

The included live gate validates BF16 causal-convolution and GDR
forward/backward against independent short references, then checks
production-length packed execution against per-segment execution. It uses
deliberately unaligned segment boundaries and compares q/k/v/g/beta
gradients. The strict marker additionally requires at least 131072
production tokens and the default tolerances; smaller shapes are diagnostic
only. A 910B4-1 run of the public backend is the remaining hardware
validation step.

### API and Usage Example

Set the public Triton compiler target before importing Triton:

```bash
export TRITON_ASCEND_ARCH=Ascend910B4
python scripts/npu/validate_910b4_gdn_training.py \
  --model-config /path/to/Qwen3.5/config.json \
  --sp-size 8 \
  --tokens 131072
```

The model continues to select the existing `npu` GatedDeltaNet backends:

```yaml
model:
  ops_implementation:
    rms_norm_gated_implementation: npu
    causal_conv1d_implementation: npu
    chunk_gated_delta_rule_implementation: npu
```

### Design & Code Changes

- Add a side-effect-free device-name-to-BV selector with a lazy
  `torch_npu` runtime lookup.
- Use the selected BV in both GDR hidden-state recurrence launches.
- Add CPU-testable B4/B4-1 and non-B4 mapping coverage to the NPU CI
  workflow.
- Add a live 910B4-1 correctness and packed-boundary gate.
[...]

PR #970, opened 2026-07-23
[ops, ci] fix: mask padded GDR dot lanes on Ascend

YLlllllllllll:
### What does this PR do?

Fix the Ascend chunked gated-delta-rule kernel when a packed segment ends
inside a 64-token tile.

Ascend `tl.dot` may produce non-finite values in padded lanes. The existing
arithmetic mask used `dot_product * tril_mask`, but `NaN * 0` remains `NaN`.
Those values entered the triangular solve and corrupted otherwise valid packed
segments. This change uses a boolean `tl.where` mask so diagonal, upper-triangle,
and padding lanes are explicitly written as zero.

### Checklist Before Starting

- Related hardware support: #966
- PR title follows the `[{modules}] {type}: {description}` format.

### Test

- Ascend 910B4-1 manual numerical gate with packed lengths `[31, 42, 54]`:
  - finite-sentinel and NaN-prefilled destinations both fully overwritten;
  - diagonal, upper-triangle, and padding lanes are exactly zero;
  - strict-lower output matches an independent CPU elementwise reference;
  - packed forward, backward (`q/k/v/g/beta`), and direct-wrapper comparisons
    all report zero mismatches.
- `pytest tests/ops/test_gdr_ascend_padding_mask.py tests/ops/test_gdn_ascend_gate.py`
  on a CPU host: `10 passed, 1 skipped`.
- `ruff check` and `ruff format --check` pass for the new test.
- `git diff --check` passes.

The generic `910b-8` CI label does not guarantee a 910B4 device, so the
hardware regression self-skips on other Ascend models. A dedicated 910B4
runner would turn this into a fail-closed automatic hardware gate.

### API and Usage Example

N/A. This is an internal numerical-correctness fix in the existing
`chunk_gated_delta_rule_implementation: npu` path.

### Design & Code Changes

- Replace arithmetic masking of the padded dot tile with `tl.where`.
- Add a first-call regression using the partial packed lengths that exposed
  the bug, two poisoned destination buffers, and a CPU reference that does not
  use matrix multiplication.
- Run the regression in its own NPU pytest process with the 910B4 Triton target
  selected before import.

- [x] Applied targeted lint and formatting checks.
- [x] Added an NPU regression test and CI entry.
- [x] Kept the patch limited to the kernel, regression, and CI wiring.


YLlllllllllll:
@Luosuu All CI checks passed on `fa2d69c` (3 green) — ready to merge when you have a moment.

Luosuu (approved):

YLlllllllllll:

YLlllllllllll:
@cursor[bot] Merge conflicts with upstream main are resolved, and the final reviewed fixes are now on the new head. Please re-review this updated head, especially the previously reported issues.

YLlllllllllll:
@cursor Confirmed against the current head `1d5bbb6`: the sibling `_ascend/triton_core/chunk_scaled_dot_kkt.py` still uses the unguarded `dot_product * tril_mask` expression. It is reachable through the `npu_ascendc` backend, so this remains an actual correctness fix, not a stale comment.

The regression test now exercises both the regular Triton and `triton_core` implementations, but the CI step remains best-effort on the generic `910b-8` label and may self-skip on a non-910B4 device. We will fix the sibling kernel and make the hardware coverage explicit before requesting re-review.

YLlllllllllll:
The Aug 16 leftover is already fixed on current head `1d5bbb6`.

Both `_ascend/triton/chunk_scaled_dot_kkt.py` and `_ascend/triton_core/chunk_scaled_dot_kkt.py` use `tl.where(tril_mask != 0.0, dot_product, 0.0)`. The regression test already launches both backends. `git merge-tree` against current `main` is clean.

Remaining caveat: the NPU CI step is still best-effort. The generic `910b-8` label does not guarantee 910B4, and the test self-skips unless `torch.npu.get_device_name()` is `Ascend910B4`/`Ascend910B4-1` and `TRITON_ASCEND_ARCH=Ascend910B4`. A green job is not proof the numerical gate ran.

Luosuu's APPROVED is still on `fa2d69c`. Re-approval of `1d5bbb6` would unblock merge.

@Luosuu @cursor

PR #976, opened 2026-07-25
[ckpt] fix: revert NPU distributed-save disable

lihanwen7:
## What does this PR do?

Reverts the `and not IS_NPU_AVAILABLE` guard from #919 that disabled the distributed `HuggingFaceStorageWriter` path on NPU, and instead adds `--train.enable_full_determinism True` to the checkpoint test command.

## Why?

#919 was a temporary workaround. The root cause of the intermittent bf16 mismatch under HSDP is a data-source divergence, not an NPU bug:

- **Before torch 2.9**: HF safetensors was generated by loading the DCP checkpoint on rank 0. Both the DCP checkpoint and HF safetensors came from the same source, so zero-tolerance verification always passed.
- **After torch 2.9**: HF safetensors is generated from the live model via distributed save. DCP dedup and HF consolidation may pick different ranks' copies for replicated shards. When NPU floating-point non-determinism causes replicate pairs to have slightly different fp32 values that straddle a bf16 rounding boundary, the two files disagree by 1 ULP and zero-tolerance verification fails.

Using `enable_full_determinism` eliminates the replicate-pair fp32 divergence at the source, making the zero-tolerance comparison valid while keeping the efficient distributed save path on NPU.

## Checklist Before Starting

- Search for relative PRs/issues and link here: #919
- PR title follows `[{modules}] {type}: {description}` format ✅

## Test

Verified on Ascend 910B + CANN 9.0.0: with `enable_full_determinism`, 188/188 HSDP replicate pairs have bit-exact fp32 values, and `test_trainer_saveload_hsdp` passes consistently.

## Checklist Before Submitting

- Read the [Contribute Guide](https://github.com/ByteDance-Seed/VeOmni/blob/main/CONTRIBUTING.md) ✅
- Applied pre-commit checks ✅
- Added/updated documentation: N/A
- Added tests to CI workflow: N/A (fix stabilizes existing CI test)

### DIFF
--- a/tests/checkpoints/utils.py
+++ b/tests/checkpoints/utils.py
@@ -103,6 +103,7 @@ def get_checkpoint_test_command(
         "--train.checkpoint.manager dcp",
         "--train.checkpoint.save_async True",
         f"--train.checkpoint.save_hf_weights {save_hf_weights}",
+        "--train.enable_full_determinism True",
     ]
     # HSDP: split the FSDP dim into (dp_replicate, dp_shard). dp_shard is
     # inferred as dp_size // dp_replicate_size by the argument resolver, and the
--- a/tests/checkpoints/utils.py
+++ b/tests/checkpoints/utils.py
@@ -64,6 +64,7 @@ def get_checkpoint_test_command(
     ep_size,
     save_hf_weights=False,
     dp_replicate_size=None,
+    full_determinism=False,
 ):
     config_path = MODEL_CONFIGS[model_name]["config_path"]
     tokenizer_path = hf_local_or_remote(MODEL_CONFIGS[model_name]["tokenizer_path"])
@@ -103,8 +104,13 @@ def get_checkpoint_test_command(
         "--train.checkpoint.manager dcp",
         "--train.checkpoint.save_async True",
         f"--train.checkpoint.save_hf_weights {save_hf_weights}",
-        "--train.enable_full_determinism True",
     ]
+    # Full determinism is only needed by the HSDP + HF-safetensors case (NPU
+    # replica fp32->bf16 casts must be bit-exact, see #919). Leaving it off by
+    # default keeps the default config (enable_full_determinism: false) covered
+    # and avoids the deterministic-algorithms slowdown on GPU CI.
+    if full_determinism:
+        params.append("--train.enable_full_determinism True")
     # HSDP: split the FSDP dim into (dp_replicate, dp_shard). dp_shard is
     # inferred as dp_size // dp_replicate_size by the argument resolver, and the
     # expert mesh becomes 3D (ep_replicate, ep_fsdp, ep), exercising the
--- a/tests/checkpoints/test_trainer_saveload.py
+++ b/tests/checkpoints/test_trainer_saveload.py
@@ -271,12 +271,23 @@ def _run_trainer_saveload_and_verify(model_name: str, ep_size: int, dp_replicate
     shutil.rmtree(get_output_dir(model_name, ep_size, dp_replicate_size))
 
 
-def _run_trainer_save_hf_safetensor(model_name: str, ep_size: int):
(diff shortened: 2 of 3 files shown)


tests/checkpoints/utils.py
```diff
@@ -103,6 +103,7 @@ def get_checkpoint_test_command(
         "--train.checkpoint.manager dcp",
         "--train.checkpoint.save_async True",
         f"--train.checkpoint.save_hf_weights {save_hf_weights}",
+        "--train.enable_full_determinism True",
```
Crystal-jiang: `get_checkpoint_test_command` is shared by **all** checkpoint test cases (`test_trainer_saveload`, `test_trainer_saveload_hsdp`, `test_trainer_save_hf_safetensor`). Adding `--train.enable_full_determinism True` here unconditionally means:

1. GPU CI cases that never needed determinism will now run with `torch.use_deterministic_algorithms(True)` + cuDNN disabled, which is a real slowdown.
2. The default configuration (`enable_full_determinism: false`) — which is what every shipped `configs/**/*.yaml` actually uses — loses test coverage.

Suggestion: parameterize the flag (e.g. `get_checkpoint_test_command(..., full_determinism=False)`) and enable it only for the HSDP test case that actually needs it.

lihanwen7:
Addressed the review feedback (cursor[bot] + Crystal-jiang). Summary of the changes:

1. **`--train.enable_full_determinism True` is no longer added unconditionally.**
   `get_checkpoint_test_command(..., full_determinism=False)` makes it opt-in; only the HSDP + HF-safetensors case enables it. All other checkpoint tests keep the default config (`enable_full_determinism: false`) covered, and GPU CI avoids the deterministic-algorithms slowdown (`torch.use_deterministic_algorithms` + cuDNN disabled).

2. **Added the missing HSDP + HF-safetensors coverage.**
   New `test_trainer_save_hf_safetensor_hsdp` (`dp_replicate_size=2, full_determinism=True`) exercises the distributed `HuggingFaceStorageWriter` with `save_hf_weights=True` — the exact config the revert re-enabled on NPU, which previously had no test (saveload HSDP ran with `save_hf_weights=False`, and `test_trainer_save_hf_safetensor` never passed `dp_replicate_size`).

3. **The distributed writer on NPU is now gated on full determinism (option A).**
   `save_hf_safetensor` takes `full_determinism` (threaded from `args.train.enable_full_determinism` in all callers) and routes:
   `use_distributed = torch>=2.9 and ckpt_manager=="dcp" and (not IS_NPU_AVAILABLE or full_determinism)`
   So production default configs (`enable_full_determinism: false`) keep the safe legacy rank-0 path, while deterministic runs get the efficient distributed path.

4. **Kept the #919 failure-mode note** in `save_safetensor_utils.py` (shortened version pointing at #919).

The branch has been rebased onto the latest `main`. NPU re-verification (`test_trainer_save_hf_safetensor_hsdp` on Ascend 910B) is pending CI.

lihanwen7:
Correction to my earlier summary: per author decision, the NPU distributed-save path is **not** gated on `enable_full_determinism`. The revert stands as-is (the #919 guard is removed unconditionally) and a short #919 note is kept in `save_safetensor_utils.py` instead.

The changes in this revision are now limited to 3 files:
1. `tests/checkpoints/utils.py` — `--train.enable_full_determinism` is opt-in via `get_checkpoint_test_command(..., full_determinism=False)` (Crystal-jiang + cursor feedback: no GPU CI slowdown, default config stays covered).
2. `test_trainer_saveload.py` — new `test_trainer_save_hf_safetensor_hsdp` (`dp_replicate_size=2, full_determinism=True`) to cover HSDP + `save_hf_weights=True`, the config the revert re-enabled on NPU. Full determinism keeps the zero-tolerance DCP-vs-HF check stable on NPU CI.
3. `veomni/utils/save_safetensor_utils.py` — a 3-line comment pointing at #919 (kept per the review note to not drop the failure-mode note).

PR #949, opened 2026-07-21
[model, dist, config, ci] feat: add DeepSeek-V4 Ulysses sequence parallel

FoolPlayer:
### What does this PR do?

Adds Ulysses sequence parallelism (SP) for DeepSeek-V4 training on the eager/TileLang DSA path.

V4 is not a plain FlashAttention model: it uses MQA (`num_key_value_heads=1`), DSA compressors/indexers that need the full sequence, and 4D HyperConnection residuals. This PR implements a V4-specific SP path:

- Q: local projection/RoPE, then `gather_seq_scatter_heads`
- KV: sequence all-gather while keeping a single head (TileLang MQA contract)
- compressor / indexer inputs: all-gather full hidden / position_ids / q_residual
- sinks sliced to local Q heads; `repeat_kv` uses local Q/K head ratio
- `DeepseekV4Model.forward` rebuilds full-seq masks + packed compression metadata under SP
- explicit error for SP + KV-cache decode (unsupported)
- unit fw/bw SP parity tests and e2e `max_sp_size=2` for deepseek_v4

Generated modeling is regenerated via patchgen (no hand edits under `generated/`).

### Checklist Before Starting

- Search for relative PRs/issues and link here: DeepSeek-V4 support / DSA / mHC / Liger landings on main
- PR title follows `[{modules}] {type}: {description}` format (enforced by [check_pr_title.yml](.github/workflows/check_pr_title.yml))

### Test

- `tests/parallel/ulysses/test_deepseek_v4_ulysses.py` (fw/bw SP1 vs SP2 parity with weight/input broadcast)
- e2e parallel matrix: `deepseek_v4` `max_sp_size` 1 → 2

### API and Usage Example

```bash
# SP=2 training (heads must be divisible by ulysses_size; V4 has 64 Q heads)
NPROC_PER_NODE=4 bash train.sh tasks/train_text.py configs/text/deepseek_v4.yaml \
  --train.accelerator.ulysses_size 2 \
  --train.accelerator.ep_size 1
```

Config note in `configs/text/deepseek_v4.yaml` documents the SP constraint.

### Design & Code Changes

1. `deepseek_v4_gpu_patch_gen_config.py`
   - SP imports / lazy names for Ulysses ops
   - packed compressor/indexer metadata paths (full-seq under SP)
   - `DeepseekV4Attention.forward` Ulysses branch
   - `DeepseekV4Model.forward` full-seq mask + packed compression metadata; reject SP+cache
2. Regenerated:
[...]

PR #981, opened 2026-07-27
[ops, dist, perf] fix: remove DeepSeek-V4 SP recompiles and device syncs

FoolPlayer:
### What does this PR do?

Two performance fixes found while profiling a 61-layer DeepSeek-V4 run with Ulysses
sequence parallelism. Neither shows up as kernel time, but together they cost seconds
per step: TileLang JIT recompiles in the sparse MLA backward, and blocking
device-to-host reads in the Ulysses all-gather.

### Checklist Before Starting

- Search for relative PRs/issues and link here: builds on #949 (DeepSeek-V4 Ulysses
  sequence parallel) and #971 (DeepSeek-V4 sparse indexing with SP)

### Test

- `tests/ops/test_deepseek_v4_kernels.py::test_tilelang_sparse_attention_backward_shares_kernel_across_kv_lengths`
  — three KV lengths (48, 129, 1024) that all round up to a single block must request
  the same `S_kv` from the kernel builder, so TileLang compiles once. Gradients still
  match the reference at cosine similarity > 0.95, including indices deliberately
  placed at the real KV length to confirm they stay masked out of the pad rows.
- `tests/parallel/ulysses/test_all_gather.py::test_all_gather_shapes_stay_on_host`
  — uneven shards gather correctly and shapes come back as plain ints rather than
  device tensors.
- Backward kernel timing on GB200 at `B=1 S=4096 H=16 D=512 KV=4352 topk=640`:
  6.75 ms padded vs 6.71 ms unpadded. Passing `S_kv` in as a runtime value instead
  (making the indirect gather strides dynamic) costs 7.68 ms, so padding is the
  cheaper way to keep the length a compile-time constant.
- All-gather sync count at `world=2, ndim=3`: 8.0 `aten::item` calls per gather
  before, 0 after. For reference, the profiled 61-layer run measured 12786
  `aten::item` and 13418 `cudaStreamSynchronize` calls per step.

### API and Usage Example

> N/A — no public API or config changes.

### Design & Code Changes

- `veomni/ops/kernels/deepseek_v4/tilelang_sparse_mla_bwd.py`: the backward kernels
  specialize on `S_kv`, and the compressor makes `S_kv` drift by a few tokens per
  step, so TileLang recompiled `sparse_mqa_bwd_kernel` and `postprocess_kernel`
  mid-training (over 10 s of JIT stalls inside a single profiled backward). `S_kv` is
  now rounded up to a 1024-token `KV_BLOCK` so the compile-time constant stops moving
  and the variant count stays bounded. Because the rounded length also widens the
  kernel's in-range index check, `topk_idxs` entries at or past the real KV length are
  invalidated before the call, and `dkv` is sliced back to the unpadded length on
  return. `topk` is left unrounded on purpose: the main loop iterates over it, so
  padding it would trade fewer recompiles for proportionally more kernel time.
- `veomni/distributed/sequence_parallel/ulysses.py`: `_all_gather` exchanged shard
  shapes as device tensors and then built `torch.Size` from them, reading one scalar
  back per dimension per rank; `_Gather.forward` read one more per rank for the split
  sizes. Shapes are now exchanged in a single `all_gather_into_tensor` and read back
  with one `tolist()`, then kept as plain ints so callers do not re-sync. Uneven
  shards still work. Output buffers switch from `zeros` to `empty` since
  `all_gather` overwrites them, and the docstring states the equal-rank precondition.

- Applied pre-commit checks
- Added/updated documentation — N/A, no API, config, or workflow surface changed
- If `tasks/` training scripts were moved or renamed: N/A
- Added tests to CI workflow — yes, one regression test per fix (listed above)

### DIFF
--- a/tests/ops/test_deepseek_v4_kernels.py
+++ b/tests/ops/test_deepseek_v4_kernels.py
@@ -256,13 +256,12 @@ def test_tilelang_sparse_attention_backward_shares_kernel_across_kv_lengths(monk
     from veomni.ops.kernels.deepseek_v4 import tilelang_sparse_mla_bwd as sparse_mla_bwd
 
     kv_block = sparse_mla_bwd.KV_BLOCK
-    compiled = []
+    requested_kv_lengths = []
     original_bwd = sparse_mla_bwd.bwd
 
-    def tracking_bwd(*args, **kwargs):
-        kernel = original_bwd(*args, **kwargs)
-        compiled.append(kernel)
-        return kernel
+    def tracking_bwd(B, S, S_kv, *args, **kwargs):
+        requested_kv_lengths.append(S_kv)
+        return original_bwd(B, S, S_kv, *args, **kwargs)
 
     monkeypatch.setattr(sparse_mla_bwd, "bwd", tracking_bwd)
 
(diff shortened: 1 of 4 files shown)


Coach257:
@szl-cursor

Coach257 (approved):

PR #983, opened 2026-07-28
[logging, lora] feat: Extend count_flops to allow Lora config for Qwen 2/3/3.5 models

wyettzeng:
### What does this PR do?

Extend count_flops to allow Lora config for Qwen 2/3/3.5 models

### Checklist Before Starting

### Test

FFT flops count unchanged for:
- 336 different Qwen models, each with 11 batch/sequence settings.
- Run a quick sanity check fo 0.6B qwen model + lora, and mfu is reported
- Added scripts/profile/compare_lora_flops.py, a script that uses Veomni's config loader to load config, and pass it to the flops counter. Tested 10 different Text / Image / MOE models, all output are similar to the following:
image

### API and Usage Example

### Design & Code Changes

- Applied pre-commit checks
- Added/updated documentation

### DIFF
--- a/tests/utils/test_count_flops.py
+++ b/tests/utils/test_count_flops.py
@@ -205,16 +205,22 @@ class TestQwen35LoraFlops:
         )
         return (linear_flops + attention_flops + gdn_flops) / delta_time / 1e12
 
-    @pytest.mark.parametrize(
-        "lora_modules",
-        [
-            ["q_proj", "k_proj", "v_proj", "o_proj"],
-            ["in_proj_qkv", "in_proj_z", "in_proj_b", "in_proj_a", "out_proj"],
-            ["gate_proj", "up_proj", "down_proj"],
-        ],
-    )
-    def test_selected_lora_modules(self, qwen3_5_counter, lora_modules):
+    def test_dense_hybrid_lora_arithmetic(self, qwen3_5_counter):
         batch_seqlens = [12, 5]
+        lora_modules = [
+            "q_proj",
+            "k_proj",
+            "v_proj",
+            "o_proj",
+            "in_proj_qkv",
+            "in_proj_z",
+            "in_proj_b",
+            "in_proj_a",
+            "out_proj",
+            "gate_proj",
+            "up_proj",
+            "down_proj",
+        ]
         flops, _ = qwen3_5_counter.estimate_flops(
             batch_seqlens,
             delta_time=2.0,
@@ -222,20 +228,6 @@ class TestQwen35LoraFlops:
         )
 
         expected = self._expected_flops(qwen3_5_counter.config, batch_seqlens, 2.0, 8, lora_modules)
-        assert flops == pytest.approx(expected, rel=1e-9)
-
-    def test_standalone_text_model_type(self, qwen3_5_counter):
-        text_config = deepcopy(qwen3_5_counter.config.text_config)
-        counter = VeomniFlopsCounter(text_config)
-        lora_modules = ["q_proj", "in_proj_qkv", "gate_proj"]
-
-        flops, _ = counter.estimate_flops(
-            [12, 5],
-            delta_time=2.0,
-            lora_config=_lora_config(8, lora_modules),
-        )
-
-        expected = self._expected_flops(text_config, [12, 5], 2.0, 8, lora_modules)
         assert flops == pytest.approx(expected, rel=1e-9)
 
     def test_vision_flops_follow_input_and_lora_targets(self, qwen3_5_counter):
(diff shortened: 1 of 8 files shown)


veomni/utils/count_flops.py
```diff
@@ -22,6 +22,39 @@
...
+)
+
+_LORA_MODULES_BY_MODEL_TYPE = {
+    "qwen2": _QWEN2_LORA_MODULES,
+    "qwen3": _QWEN2_LORA_MODULES,
+    "qwen3_5": _QWEN3_5_LORA_MODULES,
+    "qwen3_5_text": _QWEN3_5_LORA_MODULES,
+}
+
+
```
xzzWZY: can we move these to `veomni/lora/config.py`？

wyettzeng: done

veomni/utils/count_flops.py
```diff
@@ -961,8 +1061,66 @@ def _estimate_qwen3_5_family_flops(self, tokens_sum, batch_seqlens, delta_time,
...
+                # Qwen3.5-MoE currently has no fused-expert target converter.
+                # Unqualified MLP names therefore match the shared expert's
+                # nn.Linear layers, while routed expert parameters stay frozen.
+                shared_count = num_hidden_layers
+                module_shapes_and_counts.update(
+                    {
+                        "gate_proj": (hidden_size, text_config.shared_expert_intermediate_size, shared_count),
+                        "up_proj": (hidden_size, text_config.shared_expert_intermediate_size, shared_count),
+                        "down_proj": (text_config.shared_expert_intermediate_size, hidden_size, shared_count),
+                    }
```
xzzWZY: Do we support lora for experts which are not shared?

veomni/utils/count_flops.py
```diff
@@ -884,8 +939,45 @@ def _estimate_qwen3_next_flops(self, tokens_sum, batch_seqlens, delta_time):
...
+                "out_proj": (linear_v_size, hidden_size, num_linear_attn_layers),
+                "gate_proj": (hidden_size, self.config.shared_expert_intermediate_size, shared_count),
+                "up_proj": (hidden_size, self.config.shared_expert_intermediate_size, shared_count),
+                "down_proj": (self.config.shared_expert_intermediate_size, hidden_size, shared_count),
+            }
+            lora_N = self._compute_lora_params(lora_rank, lora_modules, module_shapes_and_counts)
+            # The legacy helper omits q_proj's output-gate half. Include the
+            # real frozen projection shape when LoRA arithmetic is requested.
+            gated_q_proj_N = hidden_size * q_size * num_full_attn_layers
+            dense_N_flops = (4 * (base_N + gated_q_proj_N) + 6 * lora_N) * tokens_sum
```
xzzWZY: Why gated_q_proj_N not in base_N and it's not calculated in full finetuning?

veomni/utils/count_flops.py
```diff
@@ -961,8 +1061,66 @@ def _estimate_qwen3_5_family_flops(self, tokens_sum, batch_seqlens, delta_time,
...
+                        "up_proj": (hidden_size, text_config.intermediate_size, num_hidden_layers),
+                        "down_proj": (text_config.intermediate_size, hidden_size, num_hidden_layers),
+                    }
+                )
+            lora_N = self._compute_lora_params(lora_rank, lora_modules, module_shapes_and_counts)
+
+            # The legacy helper ignores q_proj's output-gate half. Include it in
+            # LoRA mode so frozen-base and adapter work use the real projection.
+            gated_q_proj_N = hidden_size * q_size * num_full_attn_layers
+            base_N = mlp_N + attn_linear_N + gated_q_proj_N + lm_head_N
```
xzzWZY: same here, why "gated_q_proj_N" not included in full finetuning?

xzzWZY (approved):

veomni/utils/count_flops.py
wyettzeng: fixed in another PR

veomni/utils/count_flops.py
wyettzeng: fixed in another PR

veomni/utils/count_flops.py
wyettzeng: just add in support for routed experts lora support

veomni/utils/count_flops.py
```diff
@@ -961,8 +1061,66 @@ def _estimate_qwen3_5_family_flops(self, tokens_sum, batch_seqlens, delta_time,
...
+                    {
+                        "gate_proj": (hidden_size, text_config.intermediate_size, num_hidden_layers),
+                        "up_proj": (hidden_size, text_config.intermediate_size, num_hidden_layers),
+                        "down_proj": (text_config.intermediate_size, hidden_size, num_hidden_layers),
+                    }
+                )
+            lora_N = self._compute_lora_params(lora_rank, lora_modules, module_shapes_and_counts)
+
+            # The legacy helper ignores q_proj's output-gate half. Include it in
+            # LoRA mode so frozen-base and adapter work use the real projection.
```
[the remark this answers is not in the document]
wyettzeng: resolved

veomni/utils/count_flops.py
```diff
@@ -15,6 +15,7 @@
 
 from transformers import PretrainedConfig
 
+from ..lora.config import LORA_MODULES_BY_MODEL_TYPE
```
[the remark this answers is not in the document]
wyettzeng: @xzzWZY asked me to put it there, so leaving there for now

veomni/utils/count_flops.py
```diff
@@ -652,7 +679,7 @@ def _estimate_qwen3_vit_flop(self, images_seqlens, config):
         dense_N = patch_embed_N + (mlp_N + attn_linear_N) * depth + deepstack_merger_N + merger_N
 
         # non-attn all_layer & all_token fwd & bwd flops
-        dense_N_flops = 6 * dense_N * tokens_sum
+        dense_N_flops = (2 if frozen else 6) * dense_N * tokens_sum
```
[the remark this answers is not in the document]
wyettzeng: fixed, but it is worth noting that vlm_trainer currently does not support LoRA, so lora calculation here is more so added for future capabilities. Current logic:

input has not image token -> vit flops = 0
input has image token but none of the lora target modules point to vit -> forward only vit
otherwise -> vit lora training

another thing worth pointing out is that our lora currently only support Linear layers and moe experts component (expressed as nn.Embedding), so only Linear layer has lora flops counted. If we add in support for other modules like conv3d later, we will need to update count_flops.

wyettzeng:
Reply to AI review:

1. Wire it into training (this is what makes the feature real) -> done
2. Don't raise from the metering path -> done, for unresolvable lora config we just report 0 flops as standard + log a warning for the user without crashing training
3. Fix the layering / single source of truth -> @xzzWZY asked to be put it there, but open to put it elsewhere if other reviewers would prefer
4. Model semantics that the current shape gets wrong — honour or explicitly document -> fixed, now take in VeOmniLoraConfig
5. missing docs -> fixed

tests/utils/test_count_flops.py
Luosuu, line 1: do we have to add such a large test?

veomni/utils/count_flops.py
```diff
@@ -104,7 +107,114 @@ def _compute_lm_head_params(hidden_size, vocab_size):
...
+            ]
+            if unsupported_patterns:
+                return (
+                    "LoRA FLOPs estimation only supports fused routed-expert target parameters ending in "
+                    f"{supported_suffixes}, got {unsupported_patterns!r}."
+                )
+        return None
+
+    @staticmethod
+    def _compute_routed_moe_lora_flops(config, tokens, lora_config):
```
Luosuu, line 194: I do not get what this is computing for

wyettzeng: For qwen 3.5 and 3 next, there are 2 types of experts:
- routed experts (ex: only topk expert active for each token), represented as nn.Embedding in code
- shared experts, where every token has to pass through it, represented as nn.Linear in code

qwen 3 only has routed experts. 

For veomni, routed expert can have 2 lora implementation: they all share the same AB lora matrix, or each get their own. This function calculates the flops specifically for these. If shared, the we can reuse the gate and up lora for all token, while need to have compute the down lora output for each expert independently as the input to down is no longer the same.

tests/utils/test_count_flops.py
wyettzeng: reduced some, but I would recommend keeping the rest as this pr does add in quite a lot of new extra logic for lora, specifically moe, and vit have their own logic for different settings.

Luosuu (approved):

PR #984, opened 2026-07-28
[ckpt] fix: port DCP consolidation patch to torch 2.12

williamLyh:
### What does this PR do?

PyTorch 2.12 changed the private DCP safetensors consolidation implementation from the mmap-based `_read_tensor_data_mmap` helper to a handle-based `_read_tensor_data` API. The existing VeOmni patch therefore raises `NameError` when its replacement `_process_output_file` runs on torch 2.12.

This PR ports the compatibility patch to torch 2.12 instead of disabling it. This is still required because upstream torch 2.12 continues to:

- use `torch.finfo` for tensor byte-size calculation, which fails for valid integer and boolean tensors; and
- open consolidated output files with `r+b`, which is incompatible with append-only filesystems such as HDFS FUSE.

### Checklist Before Starting

- Searched for related PRs/issues and reviewed the upstream torch 2.12 implementation.
- PR title follows the required `[{modules}] {type}: {description}` format.

### Test

The existing `test_save_state_passed_to_dcp_save` regression test exercises the complete torch 2.12 handle-based path:

1. Save a state dict containing `bfloat16`, `int64`, and `bool` tensors through `HuggingFaceStorageWriter` with consolidation enabled.
2. Consolidate the distributed checkpoint into a single safetensors file.
3. Reload the file and verify the integer and boolean tensors round-trip unchanged.
4. Verify consolidated `int64` metadata has the expected shape and byte size from `Tensor.element_size()`.

PR CI passed the GPU unit/E2E, Ascend, lint, documentation-path, and device API checks.

### API and Usage Example

No public API or user-facing usage changes. `save_hf_weights` continues to apply the compatibility patch automatically.

### Design & Code Changes

- Add torch 2.12 to the supported-version allowlist and allowlist its `_process_output_file` source hash.
- Keep the existing mmap-based implementation for torch 2.9–2.11.
- Add a torch 2.12 implementation that opens input shards once, reads tensor data through the new handle-based `_read_tensor_data` API, and closes all handles in a `finally` block.
- Select the implementation from the internal reader helper exposed by the installed torch version.
- Preserve the two original fixes: `Tensor.element_size()` for non-floating dtypes and append-mode (`ab`) output for HDFS FUSE.
- Update the compatibility documentation to include torch 2.12.

- [x] Applied pre-commit/lint checks
- [x] Updated documentation
- [x] No `tasks/` scripts were moved or renamed
- [x] Existing DCP-to-HF regression coverage validates the changed path

### DIFF
--- a/.agents/knowledge/constraints.md
+++ b/.agents/knowledge/constraints.md
@@ -144,7 +144,7 @@ Core files:
     - Calling checkpoint operations from only rank 0 causes deadlocks.
 
 18. **Distributed HF safetensors consolidation must support non-floating tensors**
-    - PyTorch 2.9–2.11 computes consolidated tensor byte sizes with `torch.finfo`, which crashes for valid integer and boolean buffers such as DeepSeek V4 `tid2eid`.
+    - PyTorch 2.9–2.12 computes consolidated tensor byte sizes with `torch.finfo`, which crashes for valid integer and boolean buffers such as DeepSeek V4 `tid2eid`.
     - `apply_dcp_consolidation_patch()` in `veomni/checkpoint/dcp_consolidation.py` replaces the metadata parser with `Tensor.element_size()` and verifies the upstream private-function source hash before patching.
     - Offline DCP-to-HF conversion may cast `save_dtype` only onto floating tensors; integer and boolean buffers must retain their original dtype, and shard-size planning must use their original element sizes.
     - Do not remove this patch during torch upgrades until the new upstream consolidator is verified with sharded integer-tensor save/load coverage.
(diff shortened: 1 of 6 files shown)


Crystal-jiang:
the PR description still has the unmodified template content. Could you fill in the sections? Specifically it would help reviewers to know:
1. What does this PR do? — torch 2.12 changed _process_output_file to a handle-based _read_tensor_data API, so the existing mmap-based patch raises NameError. This PR ports the patch to 2.12 while keeping the torch.finfo and "ab" mode fixes.
2. Test — how the 2.12 handle path was validated (e.g. int64/bool tensor save → consolidate → reload round-trip)
3. Design & Code Changes — the commit message already covers this well, but a brief note in the description would help.

PR #1015, opened 2026-08-05
[ops, model, config, docs] feat: fused Triton RoPE for DeepSeek-V4

FoolPlayer:
> **Stacked on #1014**, which is itself stacked on #1013. Review the last commit
> only.

### What does this PR do?

V4's RoPE rotates only the trailing `qk_rope_head_dim` slice of each head, with
an interleaved cos/sin layout. The eager implementation runs that as a slice, a
`repeat_interleave`, two elementwise multiplies and a concat — a chain of
separate kernels over the full activation, repeated at every call site: Q, MQA
KV, the inverse rotation on the attention output, the indexer Q, and all three
compressors.

This adds a fused Triton kernel for that layout and selects it through the ops
registry. **1.273 -> 1.182 s/step (+7.1%)** on a 4-layer model at 16K, peak
memory unchanged.

### Checklist Before Starting

- Related: #1013 and #1014 (this stack), #974 (NPU DeepSeek-V4 — this PR merges
  with its generated-module selection and confirms its RoPE conclusion)

### Test

`pytest tests/ops/test_deepseek_v4_kernels.py` passes (44 tests). `make quality`
and `make check-patchgen` pass.

**Numerics.** The kernel is checked against the eager reference across head
counts, rope widths, dtypes and non-contiguous inputs, including the odd-shaped
cases the gate is supposed to reject.

**Dispatch.** Verified end to end rather than only in isolation: on a production
64K trace, the fused symbol appears at every call site with **zero** eager
fallbacks, confirming the gate does not silently opt out at real shapes. GPU
end-to-end overrides in `tests/tools/training_utils.py` now pin `triton` so the
kernel is exercised under FSDP2, Ulysses SP and the TileLang paths rather than
only by the standalone kernel tests.

**Speed.** 1.273 -> 1.182 s/step (+7.1%) over 200 steps on a 4-layer
DeepSeek-V4-Flash-Base at 16K, peak memory unchanged at 123.1 GB.

No NPU hardware was available. On NPU the pin resolves to `eager`, which is what
`configs/text/deepseek_v4_npu.yaml` already specifies, so behavior there is
unchanged.

### API and Usage Example

`configs/text/deepseek_v4.yaml` switches to:

```yaml
rotary_pos_emb_implementation: triton
```

`eager` remains available and unchanged. `liger_kernel` and `npu` were already
unusable for this model and now say so explicitly instead of failing obscurely.

### Design & Code Changes

- **`veomni/ops/kernels/rotary/triton_deepseek_v4.py`** (new): fused
  partial-interleaved RoPE. The nope channels are written through unchanged
  rather than copied separately — a sliced `copy_` measured about 3x slower per
  byte, since a partial-channel copy cannot coalesce.

  The support gate requires `cos`/`sin` on the input's device and a rope slice
  that is non-empty and no wider than the head; anything else falls back to
  eager, because those cases would index out of bounds inside the kernel.

- **`device_patch.py`** (new): maps `rotary_pos_emb` for this model. Replacing
  the module-level `apply_rotary_pos_emb` covers every call site at once.
  `liger_kernel` and `npu` are mapped to `None` — neither implements a
  trailing-slice partial rotary (they produce NaN), and both expect a
  `(q, k, cos, sin)` pair returning two tensors rather than V4's single tensor.
  `None` gives a clean "explicitly disabled" error instead of a wrong-signature
  crash or silent garbage.

  Worth noting: #974's NPU config reached this same conclusion independently and
  pinned `eager` for the same reason, so the two agree.

- **`packed_utils.py`**: the compressors take `apply_rope` as a required keyword
  argument instead of importing the eager reference at module level, so they
  cannot silently keep calling eager while every other call site is fused.

- **`__init__.py`**: applies the device patch to whichever generated module
  #974's `IS_NPU_AVAILABLE` branch selected.

- Docs and `arguments_types.py` metadata updated to list DeepSeek-V4 (and Wan)
  under `triton`, and to state that V4 rejects `liger_kernel` and `npu`.

### Behavior change for configs that omit `rotary_pos_emb_implementation`

DeepSeek-V4 had no `device_patch.py` before this PR, so the field was ignored and
the registry default `liger_kernel` was harmless. It is now rejected at model
registration with "explicitly disabled", because that backend's signature does
not match V4's partial interleaved RoPE. Both in-repo configs pin the field, so
nothing in tree changes -- but a user config that omitted it will now fail where
it previously trained, and has to pin `triton` or `eager`.

- Applied pre-commit checks
- Added/updated documentation: `docs/design/kernel_selection.md`,
  `docs/design/unified_kernel_registry.md`, `docs/usage/arguments.md`
- Added tests to CI workflow: yes, in `tests/ops/test_deepseek_v4_kernels.py`,
  plus the GPU end-to-end override in `tests/tools/training_utils.py`


FoolPlayer:
All three points addressed.

**The blocking one — `apply_rope` at the NPU compressor call sites.** Confirmed exactly as you bound it, and fixed in `3c65ccb79`: both hand-written overrides now pass `apply_rope=apply_rotary_pos_emb`, and `make patchgen` regenerated. On NPU that name resolves to the eager implementation, since `device_patch` disables the NPU rotary backend for V4 — so NPU keeps the behaviour it had before this stack and the Triton fusion stays a GPU-side change. Your diagnosis of why only the Indexer picked the change up (it is reused from the GPU config; the two compressors are forks) is right and is what the fix follows.

**The GPU-free static guard.** Agreed that this bug class has now landed more than once, so it is in the stack at the base, in #1013: `[ci] test: bind generated call sites against their callees`. It is close to the script you sketched, generalized over all 29 generated modules, and it reads source rather than importing — which matters here specifically, because several veomni modules import `torch_npu` at module scope, so an import-based resolver could not check the NPU call sites that keep breaking. It fails on this commit's parent, and it independently caught the `build_block_bias` instance in #1014.

Worth recording why CI could not have: these PRs target a branch rather than `main`, so the Ascend workflow never triggers on them at all. The guard runs in the ordinary unit-test job and needs no accelerator.

**Conflict with `main`.** Resolved at the bottom of the stack as you suggested: the whole stack is rebased onto current `open/main` (which had moved four commits, including #962). The conflict was both sides appending to the end of `tests/ops/test_deepseek_v4_kernels.py` with disjoint content, so both are kept. `git merge-tree open/main HEAD` is now clean.

**Behavior change for unpinned configs.** Called out in the PR description now. Both in-repo configs pin `rotary_pos_emb_implementation`, but a user config that omits it previously ignored the field and now fails at registration with "explicitly disabled".

On the absolute-vs-slice RoPE parity subtlety: that is exactly what the `(headdim - rope_dim) % 2 == 0` term in `_is_supported` exists for, and it was not spelled out in the description. Thanks for deriving it independently.

Issue #101, opened 2025-09-26
When ep_enabled, raise an error during pre_fetch.


imhuim982:
### System Info

If not every layer contains an MoE Layer, the following code will throw a null error:

in torch_parallelize, line 361:
```
    if need_manual_prefetch:
        blocks = [pair[1] for pair in layer_pairs]
        next_blocks = blocks[1:] + [None]
        for current_block, next_block in zip(blocks, next_blocks):
            if next_block is not None:
                prefetch_modules = next_block._fsdp_modules
                assert len(prefetch_modules) > 1, (
                    "Modules to manually prefetch should not be empty when EP is enabled or there are modules to be ignored by mixed precision policy"
                ) <--- assert failed, if not every layer contains an MoE layer
                # prefetch in order of attn, gate, experts
                current_block.set_modules_to_forward_prefetch(list(reversed(prefetch_modules)))

        # configure backward prefetch
        rev_blocks = list(reversed(blocks))
        prev_blocks = rev_blocks[1:] + [None]
        for current_block, prev_block in zip(rev_blocks, prev_blocks):
            if prev_block is not None:
                prefetch_modules = prev_block._fsdp_modules
                current_block.set_modules_to_backward_prefetch(list(reversed(prefetch_modules)))
```

### Information

- [ ] The official example scripts
- [x] My own modified scripts

### Tasks

- [x] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

### Reproduction

1. create a DeepSeekV3 model，with one layer that doesn't contain a MoE layer

### Expected behavior

no error and no assertion error

imhuim982:
If I simply remove the assertation, will it cause EP failure?

Luosuu:
Hi, I think this is fixed in the latest commit. Could you please do a git pull and try again? Thank you!

Simply removing this assertion would not cause errors for EP or FSDP2, but may configure prefetching inproperly which would result in degraded training throughput. If you are fine with slower training, you can safely comment out this assertion.

Luosuu:
Deepseek model should have been fixed as mentioned in https://github.com/ByteDance-Seed/VeOmni/issues/104
Feel free to reopen the issue if you still encounter similar issues.

PR #1043, opened 2026-08-12
[parallel] fix: stabilize nested FSDP2 backward prefetch topology

Chyu0401:
### What does this PR do?

This PR fixes unbounded FSDP2 parameter all-gather retention observed with the following combination:

```text
activation checkpointing
+ nested FSDP2 groups
+ MoE expert parallelism
+ default backward prefetch
```

PyTorch's default FSDP2 backward prefetch infers the next group from the mutable post-forward execution order. Activation checkpoint recomputation appends nested Decoder and Expert groups to that order again. The heuristic can then prefetch an Expert group that is not consumed by the next `pre_backward`, leaving its all-gather result alive until final backward cleanup. On deep MoE models, one unconsumed Expert buffer may accumulate per layer and cause OOM before backward finishes.

This change assigns every nested FSDP group a stable role at parallelization time and builds independent backward prefetch chains for equivalent roles across sequential transformer blocks. For example:

```text
decoder[i] -> decoder[i - 1]
expert[i]  -> expert[i - 1]
```

The first group in each chain uses a self-target to prevent falling back to the dynamic default heuristic. Backward prefetch configuration is also decoupled from `forward_prefetch`, so disabling forward prefetch no longer disables the safe backward topology.

This PR does not change forward/backward math, parameter sharding, gradient reduction, optimizer updates, or activation checkpoint recomputation.

Related work:

- #871 proposes a broader decoupled-trigger FSDP2 approach for MoE recomputation. This PR keeps the existing composable FSDP2 path and stabilizes its manual prefetch topology.
- #101 reported a previous heterogeneous-layer prefetch failure. The role-based chains in this PR avoid assuming that every layer has the same nested-module layout.

### Checklist Before Starting

- Related PRs/issues:
  - https://github.com/ByteDance-Seed/VeOmni/pull/871
  - https://github.com/ByteDance-Seed/VeOmni/issues/101
- PR title follows the required format:
  - `[parallel] fix: stabilize nested FSDP2 backward prefetch topology`

### Test

Focused unit tests:

```bash
python -m pytest -q tests/distributed/test_fsdp2_prefetch.py
```

Result:

```text
7 passed
```

Code quality checks for the changed files:

```bash
ruff check \
  veomni/distributed/torch_parallelize.py \
  tests/distributed/test_fsdp2_prefetch.py

ruff format --check \
  veomni/distributed/torch_parallelize.py \
  tests/distributed/test_fsdp2_prefetch.py
```

Result:

```text
All checks passed.
2 files already formatted.
```

The same FSDP2 diff was validated in an Ascend NPU integration run:

```text
Model:                  Qwen3.5-35B-A3B
Devices:                4 x Ascend 910B
DP shard size:          4
EP size:                2
Sequence length:        4096
Activation checkpoint: enabled
Forward prefetch:       disabled
FSDP offload:           disabled
Activation offload:     disabled
```

The framework configured three static backward chains covering 107 non-root FSDP groups:

```text
FSDP2 manual prefetch configured for 3 backward chain(s) and 107 group(s);
forward prefetch enabled=False.
```

Pending prefetch memory remained bounded during backward:

```text
Decoder prefetch: approximately 0.0697 GiB
Expert prefetch:  approximately 0.7500 GiB
Total:            approximately 0.8197 GiB
```

Backward completed from the last layer to Layer 0. At `backward_post`:

```text
Decoder groups:           40/40 sharded
Expert groups:            40/40 sharded
Other groups:             27/27 sharded
Root group:                 1/1 sharded
pending_all_gather_groups:    []
```

The run subsequently OOMed during the first `AdamW.step()` while lazily creating `exp_avg_sq`. This is a four-device optimizer-state capacity limitation after backward has completed and is unrelated to FSDP prefetch.

Multi-step 397B validation and a stable-step throughput comparison are still in progress.

### API and Usage Example

No new public API or configuration field is introduced.

The existing option keeps its forward-only meaning:

```yaml
train:
  accelerator:
    fsdp_config:
      forward_prefetch: false
```

For models with nested extra-parallel or mixed-precision-ignored FSDP groups, static backward prefetch chains are still configured when forward prefetch is disabled.

### Design & Code Changes

- Assign each nested FSDP group a stable role using:
  - group kind;
  - extra-parallel name;
  - normalized model-stack path;
  - module-relative path.
- Register overlapping nested modules once by object identity.
- Build one backward chain per role in original model traversal order instead of lexicographic FQN order.
- Configure `group[i] -> group[i - 1]` backward prefetch targets.
- Use a self-target for the first group in each chain to suppress fallback to the mutable default heuristic.
- Configure backward prefetch independently from `forward_prefetch`.
- Preserve the existing forward-prefetch behavior when forward prefetch is enabled.
- Add tests covering:
  - every nested FSDP role;
  - forward/backward prefetch decoupling;
  - singleton self-target chains;
  - heterogeneous blocks;
  - multiple model-stack namespaces;
  - original traversal ordering;
  - duplicate module registration.

The static chain assumes that transformer target modules execute sequentially in model traversal order. Models with dynamic layer skipping, repeated module calls, or non-standard interleaving require separate validation.

- [ ] Run `pre-commit run --all-files` before final submission.
- [x] Focused Ruff and unit-test checks pass.
- [x] No public API/configuration documentation update is required.
- [x] No `tasks/` training script was moved or renamed.
- [x] Tests are included under the existing `tests/distributed/` collection; no CI workflow change is required.

### DIFF
--- a/tests/distributed/test_fsdp2_prefetch.py
+++ b/tests/distributed/test_fsdp2_prefetch.py
@@ -1,0 +1,166 @@
+from veomni.distributed.torch_parallelize import (
+    _build_fsdp2_backward_prefetch_chains,
+    _configure_fsdp2_manual_prefetch,
+    _get_fsdp2_prefetch_role,
+    _register_fsdp2_prefetch_module,
+)
+
+
+class _FakeFSDPModule:
+    def __init__(self, name):
+        self.name = name
+        self._fsdp_modules = []
+        self.forward_prefetch = None
+        self.backward_prefetch = None
+
+    def set_modules_to_forward_prefetch(self, modules):
+        self.forward_prefetch = modules
+
+    def set_modules_to_backward_prefetch(self, modules):
+        self.backward_prefetch = modules
+
+
+def _make_nested_blocks(num_blocks=3):
+    blocks = []
+    experts = []
+    for index in range(num_blocks):
+        block = _FakeFSDPModule(f"decoder.{index}")
+        expert = _FakeFSDPModule(f"decoder.{index}.experts")
+        block._fsdp_modules = [expert, block]
+        blocks.append(block)
+        experts.append(expert)
+    return blocks, experts
+
+
+def test_backward_prefetch_configures_every_nested_fsdp_chain():
+    blocks, experts = _make_nested_blocks()
+
+    _configure_fsdp2_manual_prefetch(
+        blocks,
+        {("extra_parallel", "ep", 0): experts, ("target",): blocks},
+        enable_forward_prefetch=False,
+    )
+
+    assert experts[0].backward_prefetch == [experts[0]]
+    assert experts[1].backward_prefetch == [experts[0]]
+    assert experts[2].backward_prefetch == [experts[1]]
+    assert blocks[0].backward_prefetch == [blocks[0]]
+    assert blocks[1].backward_prefetch == [blocks[0]]
+    assert blocks[2].backward_prefetch == [blocks[1]]
+    assert all(block.forward_prefetch is None for block in blocks)
+
+
+def test_forward_prefetch_keeps_existing_next_block_order():
+    blocks, experts = _make_nested_blocks()
+
+    _configure_fsdp2_manual_prefetch(
+        blocks,
+        {("extra_parallel", "ep", 0): experts, ("target",): blocks},
+        enable_forward_prefetch=True,
+    )
+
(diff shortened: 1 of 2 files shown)


Lihui-Gu:
https://github.com/pytorch/pytorch/blob/449b1768410104d3ed79d3bcfe4ba1d65c7f22c0/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py#L460  torch2.10 solved, no changes to VeOmni are required.

Issue #1058, opened 2026-08-18
[BREAKING][omni, task, config, ci] chore: remove the SeedOmni V1 model stack


Coach257:
Part of #1057. **Depends on:** nothing.
## Why

SeedOmni V1's `encoder`/`foundation`/`decoder` triple is superseded by the V2 graph. Deleting it first frees the `veomni/models/seed_omni/` namespace so V2 can land under the same package without a rename dance, and removes ~7k lines from the review surface of every later PR.

V1 has no external users we are aware of: the entry points (`tasks/omni/train_omni_model.py`, `tasks/infer/infer_omni_model.py`) are not referenced by any CI workflow, and `configs/multimodal/omni/seed_omni.yaml` is the only config that drives them.

## Scope

Delete:

- `veomni/models/seed_omni/{auto,configuration_seed_omni,modeling_seed_omni,processing_seed_omni,projector,omni_utils}.py`
- `veomni/models/seed_omni/encoder/` — `base.py`, `janus_siglip/`, `qwen2_vl_vision_model/`, `qwen2_5_vl_vision_model/`, `qwen2_5_omni_vision_model/`, `qwen2_5_omni_audio_model/`
- `veomni/models/seed_omni/foundation/` — `base.py`, `janus_foundation/`, `llama_foundation/`, `qwen2_vl_foundation/`, `qwen2_5_vl_foundation/`, `qwen2_5_omni_foundation/`, `qwen3_moe_foundation/`, `seed_oss_foundation/`
- `veomni/models/seed_omni/decoder/` — `base.py`, `cosmos/`, `instruct_pix2pix/`, `janusvq16/`, `movqgan/`, `ultra_edit/`
- `tasks/omni/train_omni_model.py`, `tasks/infer/infer_omni_model.py`
- `configs/multimodal/omni/seed_omni.yaml`

Clean up the references left behind:

- `veomni/trainer/callbacks/channel_loss_callback.py` — drop the `SeedOmniModel` MRO walk and the `Qwen3MoeFoundationModel` rejection branch
- `tests/trainer/test_channel_loss_callback.py` — drop the matching cases
- `veomni/models/__init__.py`, `veomni/data/multimodal/multimodal_transform.py`, `veomni/data/multimodal/multimodal_chat_template.py`, `veomni/distributed/sequence_parallel/{data,loss}.py`, `tests/special_sanity/check_device_api_usage.py`
[...]

Issue #1057, opened 2026-08-18
[BREAKING] [omni] chore: SeedOmni V2 (omniv2) upstreaming plan


Coach257:
# SeedOmni V2 upstreaming plan

[#1056](https://github.com/ByteDance-Seed/VeOmni/pull/1056) carries the whole `omniv2` branch. That is not reviewable in one pass, so this issue tracks splitting it into a sequence of small PRs against `main`. #1056 stays open as a **draft integration reference** until the last sub-PR lands, then it is closed (not merged).

## What SeedOmni V2 is

SeedOmni V1 (`veomni/models/seed_omni/` on `main` today) models an omni-model as a fixed `encoder → foundation → decoder` triple. Every sub-model is a leaf of one `SeedOmniModel`, and one `BaseTrainer` owns the whole thing: one `ParallelState`, one FSDP config, one optimizer, one checkpoint. That topology cannot express Bagel (mixture-of-transformers + VAE + flow connector), a talker/thinker split, or "freeze the ViT, EP-shard the LLM, DDP the VAE".

V2 replaces the fixed triple with two orthogonal ideas:

**1. The model is a graph, not a pipeline.** `OmniConfig` declares a set of named modules plus a training FSM and one or more generation FSMs
(`veomni/models/seed_omni/graphs/`). `OmniModel` walks the FSM; each node is a `(module, endpoint)` call such as `text_encoder.encode` or `llm.forward`. New capability = new node/edge in YAML, not a new subclass. Graphs are exportable to Mermaid (`scripts/visualize_omni_graph.py`, see `graphs/`).

**2. Every module owns its own runtime.** `ModuleRuntime` (`veomni/models/seed_omni/accelerator/module_runtime.py`) is a per-module
training unit that owns *build → parallelize → weight-load → freeze/LoRA → optimizer → lr-scheduler → checkpoint → grad-clip*, plus its **own**
`ParallelState`/device mesh. `OmniModelRuntime` composes N of them and runs the graph. `OmniTrainer` is then a thin orchestrator: distributed init, dataloader, train loop, callbacks — it never reaches into a module's mesh.

[...]

PR #1086, opened 2026-08-19
[dist, parallel] fix: correct DDP gradient sync under Ulysses SP

### LINKED ISSUE #1059: [dist, parallel] fix: correct DDP gradient sync under Ulysses SP
Part of #1057 . **Depends on:** nothing.

## Why

Two independent DDP bugs, both invisible without SP and both blocking V2 (where a vision tower is routinely DDP-wrapped next to an FSDP2 LLM).

**1. DDP gradients are not synchronized across SP ranks.** DDP wraps with `process_group=dp_group`. Turning on Ulysses shrinks `dp` (`dp_size = world / sp_size`), so at `dp_size == 1` DDP's all-reduce is a no-op while each rank still holds gradients from only its `1/sp` slice of the sequence. The optimizer then steps on a partial gradient — silently, with no shape error. FSDP2 does not have this bug because it already reduces over `fsdp_group` (`dp_sp`).

Fix: in `veomni_clip_grad_norm`, when `dp_mode == "ddp"` and `sp_size > 1`, AVG-all-reduce `p.grad` over `fsdp_group` before clipping. `dp_sp` is exactly FSDP2's effective sync surface (including HSDP, where it already contains `dp_replicate`), so the two modes end up numerically equivalent.

**2. `broadcast_buffers` must be `False` unconditionally.** FSDP2/HSDP performs *no* buffer sync at all (`torch.distributed.fsdp` contains no `_sync_module_states` / `_broadcast_coalesced`; `fully_shard` only places buffers on device). A module's buffer semantics must not change with the `fsdp_mode` a config happened to pick. Nothing is lost: config-derived buffers (SigLIP `position_ids`, static rope tables) are already identical on every rank, and dynamic-rope `inv_freq` is deliberately recomputed from *this* rank's sequence length by `modeling_rope_utils.py` — pushing rank 0's copy would actively corrupt the others. Genuinely replicated mutable state (`nn.BatchNorm*` running stats) is not fixed by `broadcast_buffers=True` either (it discards the other ranks' statistics rather than aggregating them); `SyncBatchNorm` is the real fix. The repo currently has zero plain `nn.BatchNorm*`.

## Scope

- `veomni/distributed/clip_grad_norm.py` — add `_allreduce_ddp_sp_grads()`; call it from the `ddp` branch of `veomni_clip_grad_norm`
- `veomni/distributed/torch_parallelize.py` — DDP wrap passes `broadcast_buffers=False` unconditionally
- `veomni/distributed/parallel_state.py` — `init_parallel_state(name=...)` becomes `Optional[str]`; passing `None` skips registry registration so a per-module caller can hold the returned state without colliding on `"base"`
- `.agents/knowledge/constraints.md` — new constraint documenting both rules

Out of scope (ships with #1064): `veomni_omni_module_clip_grad_norm` and `omni_clip_grad_norm`, which take `ModuleRuntime` objects that do not exist yet.

## CI

New `tests/parallel/test_ddp_sp_grad_sync.py`: a small DDP model at `ulysses_size = world_size`, asserting post-clip gradients match the `sp=1` single-process reference. Wire into the "Run parallel tests" step of `.github/workflows/gpu_unit_tests.yml`.

## Acceptance

- [ ] DDP + SP gradient parity test passes at `sp=2` and `sp=4`
- [ ] Existing `tests/parallel/ulysses/` and `tests/utils/test_extra_parallel_clip_grad_norm.py` unchanged and green
- [ ] Constraint documented in `.agents/knowledge/constraints.md`

Coach257:
### What does this PR do?

Closes #1059.

Under Ulysses sequence parallel, DDP steps the optimizer on a partial gradient. `dp_size = world_size / sp_size`, so a run with `sp_size == world_size` collapses `dp` to one rank and DDP's all-reduce becomes a no-op, while each rank still holds gradients from only its `1 / sp_size` slice of the sequence. Nothing raises and no shape mismatch appears — the run simply trains on the wrong gradient. FSDP2 is unaffected because it reduces over `dp_sp`.

Fixing that exposed a second problem: the DDP path could not start at all under the default `init_device: meta`. Nothing on that path materializes meta parameters or loads weights, so DDP's constructor died on `Tensor.item() cannot be called on meta tensors`. Both are fixed here, since neither is testable without the other.

Follow-up to #1057. Independent of #1058. Reference implementation: #1082.

### Checklist Before Starting

- Related: #1059 (issue), #1057 (context), #1082 (reference), #1058 (independent)
- PR title follows `[{modules}] {type}: {description}`

### Test

Three suites, all wired into both `gpu_unit_tests.yml` and `npu_unit_tests.yml`:

- `tests/parallel/test_ddp_sp_grad_sync.py` (new) — torchrun cases comparing DDP+SP gradients against a single-rank reference over the same global batch: `sp2`, `sp4`, `dp2 x sp2`, `sp4` with gradient accumulation, and `sp2` starting from meta-init. Each asserts parity against the reference *and* that the pre-sync gradient differs from it, so the check cannot pass vacuously.
- `tests/parallel/test_parallel_state_registry.py` (new) — `name=None` skips registration while still setting the ambient state and sharing the topology cache; a later named call with the same topology returns the same object.
- `tests/checkpoints/test_dcp_checkpointer.py` — `TestMaterializeAndLoadDispatch` pins which loader is chosen for each combination of `weights_path` / `should_skip_hf_weight_load` / `is_peft_model`, plus non-persistent buffer preservation. `TestDdpMetaInit` covers random init, resume without reading the HF snapshot, the every-rank read, the ExtraParallel and `ep_sharded_stream_load` refusals, the post-load meta check, and the two cases where a real model must be left untouched.

Not covered by CI: no multi-node run, and no convergence run. The gradient-parity tests are exact (`rtol=0, atol=0` against the reference reduction), so a loss-curve comparison would add little.

### API and Usage Example

`init_parallel_state(name=...)` becomes `Optional[str]`, default unchanged at `"base"`. `None` claims no registry key, for a caller that keeps the returned state itself instead of looking it up later — otherwise it must either collide on `"base"` with the standalone trainers or invent a key nobody reads. Only the registry is opted out of: the state is still topology-cached and still becomes the ambient global if none is set.

`parallelize_model_ddp()` is new and public, mirroring `parallelize_model_fsdp2()`. `build_parallelize_model()` dispatches to it; callers need no change.

`train.init_device` no longer accepts `"cpu"`, and `fsdp_mode: ddp` asserts against it at parse time. This value could not produce a working run on any path: fsdp2 already asserted `meta`; DDP wraps with `device_ids=[local_rank]`, which torch rejects for a CPU-resident module; and with no FSDP wrap nothing materializes or loads at all. The assert lives in `TrainingArguments._validate_accelerator()` rather than in `parallelize_model_ddp()` so that every rank fails together at parse time, instead of rank0 dying in DDP's constructor while the others block in its first collective.

Note that narrowing the `Literal` is not itself an enforcement: the parser only turns a `Literal` into argparse `choices`, which covers the CLI, while a YAML value reaches the dataclass constructor untouched. The explicit assert is what rejects it.

### Design & Code Changes

**1. Gradient sync — `veomni/distributed/clip_grad_norm.py`**

`_allreduce_ddp_sp_grads()` averages the DDP module's gradients over `fsdp_group` (`dp_sp`) before clipping, when `dp_mode == "ddp"` and `sp_size > 1`.

- `dp_sp` is exactly FSDP2's effective sync surface, `dp_replicate` included, so HSDP is covered and the two dp modes become numerically equivalent. Reducing over `sp` alone would be cheaper but silently wrong under `no_sync()`-style accumulation, where DDP has not yet reduced over `dp`.
- Averaging over `dp_sp` after DDP already averaged over `dp` is not a double count: every rank of a `dp` group holds the same value going in, so the second average reproduces the plain mean over all `dp_sp` ranks.
- `SUM` followed by `div_` rather than `ReduceOp.AVG`, which the NPU backend rejects.
- The loop keys on `requires_grad`, zero-filling a missing `grad`, so the set of collectives is identical on every rank even when a parameter goes unused on some of them. Skipping instead would desynchronize and hang.
- One collective per gradient, uncoalesced. Bucketing is the obvious follow-up; left out here to keep the fix reviewable.
- `clip_grad_norm_` returns a DTensor if any gradient is one, and `.item()` on it reads the local shard, so the reported norm is converted with `full_tensor()` — the same thing the fsdp2 path does. The clip itself is already applied globally. This branch is defensive: TP is inert today, since the wrap passes no `parallelize_plan`.

**2. DDP meta-init — `veomni/distributed/torch_parallelize.py`**

The materialize-and-load pass that lived inline in `parallelize_model_fsdp2()` is extracted into `_materialize_and_load_weights()` and called from both paths. It is the single place where the choice between random init, an HF snapshot and a checkpoint resume is made, so a future dp mode owes the same call.

The DDP call is gated on `any(param.is_meta for param in model.parameters())`, not on `init_device`. The flag states an intent the model builder is free to ignore — `tests/data/*` build eagerly while the config still says `meta` — and materializing a model that already holds weights would discard them, or raise on a plain `nn.Module` with no `init_weights`. After the pass, any parameter left on meta is reported by name, rather than surfacing inside DDP's constructor as `Tensor.item() cannot be called on meta tensors`.

`should_skip_hf_weight_load` is honoured here too: a distributed-checkpoint resume is about to overwrite every parameter, so reading the HF snapshot doubles peak memory and the snapshot may not exist. ExtraParallel and `ep_sharded_stream_load` are refused, since only fsdp2 applies the plan that shards expert weights; until now the meta crash was the only thing preventing that combination from loading full tensors into a model whose config says they are sharded.

**3. Buffers — `veomni/distributed/torch_parallelize.py`**

The DDP wrap passes `broadcast_buffers=False` unconditionally. FSDP2 syncs no buffers at all, and a module's buffer semantics must not change with the `fsdp_mode` a config happened to pick. Nothing is lost: rank0's copy is either identical to the other ranks' or, for dynamic-rope `inv_freq`, wrong for them.

**4. Parallel state — `veomni/distributed/parallel_state.py`**

`name` becomes `Optional[str]`; `None` skips both the early registry hit and the two registry writes.

**5. Docs — `.agents/knowledge/constraints.md`, `docs/usage/arguments.md`**

New constraints 7a (DDP gradient sync and buffer policy) and 7b (DDP materialization and weight loading), each recording the reasoning above so the next reader does not re-litigate the `dp_sp` choice or delete the `init_device` assert as dead code.

- [x] Read the Contribute Guide
- [x] Added/updated documentation
- [x] Added tests to CI workflow


FoolPlayer:
@coderabbitai review

FoolPlayer:
@coderabbitai rate limit

tests/parallel/test_ddp_sp_grad_sync.py
```diff
@@ -0,0 +1,262 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
FoolPlayer, line 1: Can we combine these two tests into one file?

FoolPlayer:
@coderabbitai Can we merge the test files instead of adding new ones?

FoolPlayer (approved):

PR #1093, opened 2026-08-21
[optim, trainer] refactor: read Muon hyperparameters from OptimizerConfig inside build_optimizer

Coach257:
Each trainer used to unpack the fifteen muon_* fields itself and pre-resolve the Muon learning rate through a helper that duplicated the policy already implemented in _build_muon_with_adamw. build_optimizer now takes the OptimizerConfig and does that in its muon branch, leaving one owner for the LR policy and one place to extend when another optimizer type grows knobs.

This also fixes tasks/omni/train_omni_model.py, which honoured optimizer.type: muon but silently ignored every muon_* field.

## Summary by CodeRabbit

* **New Features**
  * Optimizer configuration now supports Muon settings directly, including learning-rate inheritance, explicit overrides, weight decay, and validation.
  * Explicit Muon options take precedence over configuration defaults.

* **Documentation**
  * Updated multimodal chat-template examples and clarified tokenizer- versus processor-based usage.
  * Expanded Muon guidance for parallelism requirements, parameter layouts, bias validation, and sharding behavior.


FoolPlayer:
@coderabbitai Can we merge the test files instead of adding new ones?

FoolPlayer (approved):

PR #1016, opened 2026-08-05
[model] fix: DeepSeek V4 cast q to bfloat16 to use tilelang kernel

wuxibin89:
### What does this PR do?

verl VeOmniEngine use torch.autocast in forward_step:
https://github.com/verl-project/verl/blob/main/verl/workers/engine/fsdp/transformer_impl.py#L1482-L1491

torch.autocast `rsqrt` and `sum` to float32, making `q` and `compressed` float32, prevent using tilelang kernel.

This PR made 2 changes:
1. Explicitly cast `q` and `compressed` back to bfloat16.
2. Fail fast if tilelang kernel is specified but not using, instead of silently fallback to eager.
3. Add router replay support

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek-V4 support for MoE router replay across GPU and NPU configurations.
  * Improved expert selection consistency for Top-K and hash-based routers.

* **Bug Fixes**
  * Corrected quantization scale handling for FP4 and non-FP4 expert weights.
  * Improved numerical stability during attention and window compression.
  * Added clear validation errors for unsupported tensor data types in optimized attention operations.
  * Strengthened validation to detect missing or duplicate router replay wiring.

PR #1107, opened 2026-08-24
[model, ops, perf] refactor: build DeepSeek-V4 sparse indices without a dense mask, and fuse its RoPE

FoolPlayer:
### What does this PR do?

Builds DeepSeek-V4's sparse attention indices directly from packed metadata instead of
materializing a dense `[B, 1, S, S_kv]` mask first, and adds a fused Triton RoPE kernel for
the model's decode-side rotary application.

Replaces #1014, which was opened against a base that predates #1012, #1013 and #1016. This
PR is rebuilt on current `main` and resolves the NPU compatibility problem raised in that
review; please close #1014 in favour of this one.

### Checklist Before Starting

- Search for relative PRs/issues and link here: supersedes #1014; builds on #1012 (sparse MLA
  kernel time), #1013 (NPU patchgen imports) and #1016 (bf16 q cast), all already on `main`.

### Test

- `pytest tests/ops/test_deepseek_v4_kernels.py` — 58 passed. The two failures in that file
  (`test_deepseek_v4_generated_attention_dispatch_matches_eager`,
  `test_deepseek_v4_generated_indexer_dispatch_and_position_fallback`) reproduce identically
  on unmodified `origin/main` on the same box, so they are pre-existing and not caused by this
  change. Both assert that an fp32 call falls back to eager, which contradicts `main`'s current
  "tilelang requested but unavailable -> raise" policy; worth a separate fix.
- `make check-patchgen` — all generated files up to date.
- `make quality` — clean.
- `tests/ops` already runs as a whole directory in `gpu_unit_tests.yml`, so the new kernel
  tests are picked up without a workflow change.

### API and Usage Example

New knob for the rotary implementation, defaulting to the existing behaviour:

```yaml
model:
  rope_implementation: triton  # or "eager"
```

### Design & Code Changes

**Mask-free sparse indices.** `DeepseekV4Model.forward` withholds the dense mask exactly when
the packed metadata can validate candidates on its own. Its absence is the signal to take the
mask-free path, so every `O(S^2)` intermediate is skipped: the compressor stops building a
block bias (`build_block_bias=False`), and `build_packed_sparse_attention_indices` derives the
index list from `position_ids`, the sliding window and the compressed candidate ranges.

`CompressedCandidates` replaces the bare top-k index tensor so the packed ranges travel with
the indices rather than being recomputed.

**NPU compatibility (the #1014 blocker).** `DeepseekV4Attention.forward` is shared verbatim
with the NPU patch config, so the new `build_block_bias=` keyword reached NPU compressor
overrides that did not accept it — a `TypeError` on every forward. Both NPU overrides now take
`build_block_bias: bool = True` and skip bias construction when it is false. NPU behaviour is
unchanged: `use_compact_sparse_indices` already requires `q.is_cuda and q.dtype ==
torch.bfloat16` (from #1016), so `mask_free_sparse` is always false there and the argument is
always true. The NPU config's import list also gains
`build_packed_sparse_attention_indices` / `CompressedCandidates`, without which the generated
NPU module referenced an undefined name.

**Dispatch strictness.** `eager_attention_forward` keeps `main`'s "tilelang requested but
declined -> `ValueError`" behaviour. Mask-free callers depend on that refusal for correctness
rather than diagnostics: they withheld the dense mask, so an eager fallback would have nothing
left to enforce causality with. The comment now says so.

**Fused Triton RoPE.** `veomni/ops/kernels/rotary/triton_deepseek_v4.py` adds the kernel, and
`device_patch.py` binds it into the generated module so both the GPU and NPU builds pick the
right implementation at registration time.

- Applied pre-commit checks
- Added/updated documentation
- Added tests to CI workflow (covered by the existing `tests/ops` directory run)

## Summary by CodeRabbit

- **New Features**
  - Added Triton rotary-embedding support for DeepSeek-V4 and Wan models.
  - Improved DeepSeek-V4 packed attention for compatible GPU workloads, including reduced dense-mask creation and packed-sequence handling.
  - Added automatic backend selection with safe fallback for unsupported configurations.
- **Bug Fixes**
  - Improved rotary embedding compatibility across layouts, inverse rotation, CPU execution, and backend validation.
  - Added validation for inconsistent packed attention masks.
- **Documentation**
  - Updated configuration and kernel support documentation for DeepSeek-V4 and Wan.
- **Tests**
  - Expanded coverage for rotary embeddings, packed attention, sparse indices, and fallback behavior.


FoolPlayer:
Both findings addressed in 0e1df0ab7.

**Mask-free gate.** Valid, and fixed — but not with the suggested `attention_mask.all()` inside the boolean chain, because that would change *when* the mask-free path is taken rather than reject a contradictory call. `VeOmni`'s collator guarantees an all-ones mask on this path (`veomni/data/data_collator.py`: SP slices `input_ids` while keeping the full mask, so padding is expressed through `cu_seq_lens_q`), which means a zero here contradicts the caller's own cu-seqlens — the branch above already validates that those span the full sequence, so every position belongs to a sequence. Silently falling back to the dense path would hide that contradiction; raising names it. The alternative you floated — raise whenever a tensor mask coexists with `cu_seq_lens_q` — would reject the standard collator output, since it always passes an all-ones mask.

On the sync cost: the branch already does `cu_seq_lens_q.detach().cpu().tolist()` a few lines up, so reading the mask adds no new class of stall.

**NPU `apply_rope`.** Correct, and already fixed in 7d1a39187 before this review landed — both call sites now pass `apply_rope=apply_rotary_pos_emb`. The Ascend CI failure that caught it is green again.

**Stale doc claim.** Fixed. The "eager-only" sentence now states that `liger_kernel` is rejected for `rotary_pos_emb_implementation` on DeepSeek-V4 and that `triton` / `eager` are the supported values, matching `device_patch.py` (`"liger_kernel": None`).

Coach257 (approved):

PR #1108, opened 2026-08-24
[BREAKING][optim, config, docs] feat: qualify head-split Muon module names instead of a DSA indexer flag

FoolPlayer:
### What does this PR do?

Makes `muon_head_split_modules` entries match a leaf module name **or any dotted path suffix**, and
rejects an entry that selects two projections whose parent modules nest inside one another, naming
the qualified forms to use instead.

This supersedes the first two commits on this branch, which added a
`muon_head_split_exclude_indexer` flag plus a hardcoded `_DSA_INDEXER_CLASS_NAMES` allowlist. Both
are gone.

### Checklist Before Starting

- Search for relative PRs/issues and link here: builds on #1011 (Muon all-to-all on HSDP meshes) and
  the head-split feature it landed with; rebased onto the post-#1111 `muon.py`.
- PR title follows `[BREAKING][{modules}] {type}: {description}` format

### Test

Full local run of the `Run Muon optimizer tests` CI step:

- `pytest tests/optim/test_muon_unit.py` — 87 passed
- `pytest tests/optim/test_optimizer_vlm_param_groups.py` — passed
- `pytest tests/optim/test_muon_fsdp2_smoke.py tests/optim/test_muon_fsdp2_parity.py` — 9 passed
- `make quality` — clean

New cases in `tests/optim/test_muon_unit.py`, all against the real `DeepseekV4Attention` /
`GlmMoeDsaAttention` built from the toy configs rather than synthetic stand-ins:

- bare `[q_b_proj]` on DeepSeek-V4 raises, parametrized over `head_group_size` in `{1, 2, 8, 16}`
- `[self_attn.q_b_proj]`, `[indexer.q_b_proj]`, `[compressor.indexer.q_b_proj]` each select one site
- `[q_b_proj, self_attn.q_b_proj]` and `[q_b_proj, indexer.q_b_proj]` both still raise
- per-layer repeats collapse into one error that counts the other sites
- sibling towers sharing `q_proj` stay selected together, and a qualified entry narrows to one tower
- entries match whole path segments (`q_b_proj` does not select `xq_b_proj`)
- GLM MoE DSA needs no qualification: `[wq_b]` → indexer, `[q_b_proj]` → attention
- the qualified entries survive the production `build_optimizer(optimizer_config=...)` path

### API and Usage Example

```yaml
train:
  optimizer:
    muon_head_group_size: 1
    muon_head_split_modules: [self_attn.q_b_proj]     # MLA only
    # muon_head_split_modules: [self_attn.q_b_proj, indexer.q_b_proj]   # both
```

**BREAKING**: `muon_head_split_modules: [q_b_proj]` was accepted on DeepSeek-V4 and quietly
head-split the DSA indexer alongside the MLA. It now fails at optimizer construction with:

```
muon_head_split_modules entry 'q_b_proj' is ambiguous: it selects
model.layers.3.self_attn.q_b_proj (8 heads) and
model.layers.3.self_attn.compressor.indexer.q_b_proj (64 heads), and the second sits inside the
module holding the first, so the name cannot say which one was meant. Replace it with
'self_attn.q_b_proj' or 'indexer.q_b_proj' -- list both to split both.
```

No in-repo config breaks: the head-split knobs are commented out in `configs/text/deepseek_v4.yaml`
(updated here) and `configs/text/qwen3_moe_muon.yaml` is siblings-only.

### Design & Code Changes

The ambiguity is that DeepSeek-V4's MLA and the DSA indexer *inside* it both name a `q_b_proj`
child, so one leaf name selects two matrices with different head counts (8 and 64). The previous
approach answered it with a default plus a per-model class allowlist. But the two parameters already
differ by path — `self_attn.q_b_proj` vs `self_attn.compressor.indexer.q_b_proj` — so the selector
only had to be able to see the prefix. That removes the config field, removes the model knowledge
from `veomni/optim/`, and stops the optimizer from choosing a side on the user's behalf.

It also covers more of the collision than the allowlist did: within one `DeepseekV4Attention`, the
CSA compressor, the HCA compressor and the indexer all name a `kv_proj` that the MLA also names, and
the allowlist only listed the indexer.

Two details that the checks are shaped around:

- **Per pair of selected sites, not per entry.** The likely reaction to the error is to *add* the
  suggested entry rather than replace the bare one. A per-entry check would then find one site under
  each entry, form no pair, and split both in silence, so `[q_b_proj, self_attn.q_b_proj]` is
  refused as well. Both orders of qualification behave identically; no spelling is accidentally safe.
- **A site is recorded whenever its head layout resolves**, not only when it splits. Otherwise
  `muon_head_group_size: 8` collapses the MLA's 8 heads to a single block, drops it from the check,
  and lets a bare `[q_b_proj]` head-split *only the indexer* — the inversion of what the entry asks
  for, with no error and no warning.

Sibling matches are deliberately still allowed: Qwen2.5-Omni's `Qwen2_5OmniAttention` and
`Qwen2_5OmniAudioAttention` both name a `q_proj`, neither encloses the other, and `[q_proj]` plainly
means both. GLM MoE DSA never collides either — its indexer's up-projection is `wq_b` while the
attention around it uses `q_b_proj` — so the recommended `[wq_b]` keeps working unqualified.

- Applied pre-commit checks
- Added/updated documentation (`docs/usage/arguments.md`, `docs/usage/basic_modules.md` incl. a
  migration note, `configs/text/deepseek_v4.yaml`)
- Added tests to CI workflow (`tests/optim/test_muon_unit.py` is already in `gpu_unit_tests.yml`)

## Summary by CodeRabbit

* **New Features**
  * Muon settings can now be supplied through optimizer configuration, with explicit overrides supported.
  * Improved Muon learning-rate resolution, including inheritance and automatic defaults.

* **Bug Fixes**
  * Muon head splitting now supports qualified dotted module paths and rejects ambiguous nested projection matches with actionable guidance.
  * Removed obsolete indexer-exclusion behavior.

* **Documentation**
  * Updated head-splitting guidance, examples, migration notes, and configuration reference to reflect qualified paths and ambiguity handling.


FoolPlayer:
@coderabbitai full review

Context for the re-review: this branch was rebased, and the earlier passes were diffed against the pre-rebase base `f593cd13`, so the walkthrough attributes changes to this PR that it does not contain — the "Argument and validation cleanup" table (AdamW `betas`, ChunkMBS removal, CPU DDP rejection, and so on) is pre-existing work already in `main`. Against the current base `1d5651fc` this PR touches 7 files, and `veomni/arguments/arguments_types.py` only changes the `muon_head_split_modules` help text.

Note also that the approach changed in `4e7f48c8`: `muon_head_split_exclude_indexer` and the `_DSA_INDEXER_CLASS_NAMES` allowlist were both removed, in favour of dotted path-suffix matching plus rejection of nested selections.

Coach257 (approved):

PR #1109, opened 2026-08-24
[ops] feat: add the TileLang teacher-distribution kernel for the DeepSeek-V4 indexer

FoolPlayer:
### What does this PR do?

Adds the TileLang kernel that computes the DeepSeek-V4 Lightning Indexer's **teacher
distribution**, and lets the existing sparse attention forward hand back the log-sum-exp it
already computes.

Kernel and plumbing only. The KL objective that consumes this arrives in a later PR, so nothing
here changes training behaviour.

### Checklist Before Starting

- Search for relative PRs/issues and link here: builds on #1012 (sparse MLA kernel time),
  already merged.

### Test

- `pytest tests/ops/test_deepseek_v4_indexer_loss.py` — 19 passed. The suite includes an
  independent PyTorch reference for the teacher distribution, `NI > 1` coverage, and the
  zero-row contract.
- `make quality` — clean.
- `tests/ops` already runs as a whole directory in `gpu_unit_tests.yml`, so no workflow change
  is needed.

### API and Usage Example

```python
from veomni.ops.kernels.deepseek_v4 import sparse_attn_tilelang, sparse_mqa_target_fwd

out, lse = sparse_attn_tilelang(q, kv, sink, topk_idxs, sm_scale, return_lse=True)
target = sparse_mqa_target_fwd(q, kv, topk_idxs, lse, sm_scale)
```

`return_lse` defaults to `False`, so every existing call site is unaffected.

### Design & Code Changes

**`sparse_mqa_target_fwd`** produces the per-candidate attention probabilities the indexer is
distilled against, renormalized with the LSE from the attention forward rather than recomputing
a softmax. Its guards `raise` rather than `assert` so they survive `python -O`.

**Optional LSE.** The forward kernel already writes the LSE; returning it costs nothing. It is
marked non-differentiable in the autograd `Function`: the LSE only ever feeds the detached
indexer teacher, and leaving it differentiable would open a second gradient route back into
`q`/`kv`/`sink` and break the decoupling the indexer objective depends on.

- Applied pre-commit checks
- Added/updated documentation
- Added tests to CI workflow (covered by the existing `tests/ops` directory run)

## Summary by CodeRabbit

* **New Features**
  * Sparse attention can optionally return log-sum-exp (LSE) values alongside attention outputs.
  * Added target attention support for compressed sparse indices on compatible CUDA hardware.
  * Improved handling of invalid or padded sparse indices and zero-mass rows.
* **Bug Fixes**
  * Added validation for unsupported head counts, data types, shapes, and configurations.
  * Ensured returned LSE values are not differentiated during backpropagation while attention outputs remain trainable.
* **Tests**
  * Added comprehensive coverage for normalization, masking, warnings, gradients, and kernel correctness.

### DIFF
--- a/tests/ops/test_deepseek_v4_indexer_loss.py
+++ b/tests/ops/test_deepseek_v4_indexer_loss.py
@@ -252,7 +252,7 @@ def test_target_kernel_rejects_more_than_64_heads():
     lse = torch.zeros(1, 2, 128, device="cuda", dtype=torch.float32)
     # Matching on "64" alone would also be satisfied by the head-multiple assert
     # or by any message that happens to mention a shape of 64.
-    with pytest.raises(AssertionError, match="one block owns the head sum"):
+    with pytest.raises(RuntimeError, match="one block owns the head sum"):
         sparse_mqa_target_fwd_interface(q, kv, topk, lse)
 
 
@@ -267,7 +267,7 @@ def test_target_kernel_rejects_head_counts_the_interface_cannot_emit():
     _require_tilelang_cuda()
     from veomni.ops.kernels.deepseek_v4.tilelang_sparse_mla_target import sparse_mqa_target
 
-    with pytest.raises(AssertionError, match=r"padded to one of \(16, 32, 64\)"):
+    with pytest.raises(RuntimeError, match=r"padded to one of \(16, 32, 64\)"):
         sparse_mqa_target(48, 64, 64)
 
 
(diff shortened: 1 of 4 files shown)


FoolPlayer:
Both findings addressed in e900ee621.

**`S_kv > 0` guard.** Valid — the analysis is right that the candidate mask cannot prevent the gather, since it only zeroes the score afterwards. Added as a raise-style guard to match the rest of the interface, so `python -O` cannot strip it, plus a regression test for the empty-kv case.

**Hard-coded `"cuda"`.** Already fixed before this review landed; the tests use `DEVICE = get_device_type()` and `device-api-check` is green.

FoolPlayer:
@coderabbitai review

Coach257 (approved):

PR #1026, opened 2026-08-06
[trainer, perf] fix: apply HSDP all-reduce control to custom trainers

WenzheWang:
### What does this PR do?

PR #781 added `_configure_hsdp_allreduce()` to avoid redundant synchronization over the HSDP replicated dimension during gradient accumulation. `BaseTrainer.train_step()` uses the helper, but the custom train loops in the text, VLM, DPO, and DiT trainers bypassed it.

This PR applies the existing helper in those train loops so replica all-reduce is disabled for the first accumulation micro-step and restored for the final micro-step. The DiT offline-embedding path remains unchanged because it does not execute model forward/backward.

Related: #781

### Checklist Before Starting

- Search for relative PRs/issues and link here: #781
- PR title follows `[{modules}] {type}: {description}` format (enforced by [check_pr_title.yml](.github/workflows/check_pr_title.yml))
  - **Allowed modules:** `agent`, `ci`, `ckpt`, `config`, `data`, `dist`, `docker`, `docs`, `logging`, `lora`, `misc`, `model`, `omni`, `optim`, `ops`, `parallel`, `perf`, `release`, `task`, `trainer`
  - **Allowed types:** `feat`, `fix`, `refactor`, `chore`, `test`
  - Breaking changes: prepend `[BREAKING]` — e.g. `[BREAKING][parallel, model] feat: dynamic batching`

### Test

Current head after rebasing onto `main`:

```text
make quality
All checks passed!
624 files already formatted

pytest tests/trainer/test_step_sync.py tests/distributed/test_torch_compile.py
47 passed, 1 skipped
```

Controlled A-B-B-A validation was performed on the pre-rebase feature commit `469f5c61` based on `main@84bc3601`. The only intervening upstream change before the current head was the unrelated SeedOss documentation PR #1023.

Configuration: 2 x NVIDIA A800 80GB, Qwen3-VL toy model, FSDP2, `dp_replicate_size=2`, `dp_shard_size=1`, gradient accumulation 4, BF16 parameters, FP32 reductions, and dynamic batching enabled.

- Both ranks disabled replica all-reduce at micro-step 0 and restored it at micro-step 3 of every optimizer step.
- Same-rank loss and gradient norm matched exactly across A1/B1/B2/A2.
- Mean median step time decreased from 20.4740 s to 19.8462 s (3.07%); derived throughput increased by 3.16%.
- Peak allocated memory was unchanged; peak reserved memory decreased by 0.287 GiB (1.83%).

[...]

PR #1110, opened 2026-08-24
[trainer, ci] feat: report model aux_metrics without folding them into the loss

FoolPlayer:
### What does this PR do?

Lets a model forward report scalar diagnostics next to the loss without those scalars joining
the backward objective. A forward populates `outputs.aux_metrics`; the trainer reports the
step's mean and never sums it into the scalar handed to `backward()`.

Generic trainer plumbing with no model-specific code. DeepSeek-V4's indexer KL is the first
intended consumer and arrives in a later PR.

### Checklist Before Starting

- Search for relative PRs/issues and link here: no overlapping open PR. Rebased onto the
  current trainer after #941, #938, #1026 and #1091.

### Test

- `pytest tests/trainer/test_aux_metrics.py` — 12 passed.
- `make quality` — clean.
- Wired `tests/trainer/test_aux_metrics.py` into both `gpu_unit_tests.yml` and
  `npu_unit_tests.yml`; the tests are pure CPU.

### API and Usage Example

A forward returns detached 0-d tensors keyed by metric name:

```python
return MoeCausalLMOutputWithPast(loss=loss, ..., aux_metrics={"indexer_kl": kl.detach()})
```

They surface as `training/indexer_kl` in the step metrics. The trainer reads the attribute with
`getattr`, so any output object may carry it and outputs without it are unaffected.

The reported value is the **mean over the step's micro batches**, then averaged over the FSDP
group by `EnvironMeterCallback`. Populate it with a per-micro-batch mean, not a running total:
a metric with sum semantics would come out divided by the gradient-accumulation count. Emit a
key on every micro batch of a step, or on none of them.

### Design & Code Changes

`BaseTrainer.postforward` merges `aux_metrics` into `loss_dict` **after** the sum that builds
the backward scalar, and detaches on the way in — an attached tensor parked in a dict keeps its
whole graph alive for the step. A metric of 1000.0 against a loss of 2.0 is asserted in the
tests so that a leak into the objective cannot pass silently.

`TextTrainer` and `VLMTrainer` do not reuse `BaseTrainer.train_step`; each keeps its own copy of
the accumulation loop. Rather than have all three divide by their own micro-batch count, the
base trainer publishes the count once, so a future trainer cannot forget it. The accumulation
assertion is parameterized over all three loop owners.

The final commit realigns the test fixtures with the trainer as it stands today: the stubs
predated `reduce_global_loss_token`, the fourth `mean_global_loss` argument,
`sync_each_train_step`, and `EnvironMeterCallback.lora_config`.

- Applied pre-commit checks
- Added/updated documentation
- Added tests to CI workflow

## Summary by CodeRabbit

- **New Features**
  - Training reports now include optional auxiliary metrics, averaged across gradient accumulation and distributed workers.
  - Metrics are available under `training/<metric>` for base, text, and vision-language workflows.
  - DiT training now supports configurable data transforms and sample logging.

- **Bug Fixes**
  - Auxiliary metrics remain separate from loss calculations and are safely detached.
  - Conflicting metric names are rejected to prevent reporting ambiguity.

- **Documentation**
  - Added guidance for configuring, averaging, and monitoring auxiliary metrics.

- **Tests**
  - Expanded GPU and NPU coverage for training metrics, checkpointing, and parallel training behavior.

### DIFF
--- a/veomni/trainer/base.py
+++ b/veomni/trainer/base.py
@@ -208,6 +208,22 @@ class VeOmniIter:
         return {}
 
 
+def mean_aux_metrics(total_aux_metrics: Dict[str, float], num_micro_steps: int) -> Dict[str, float]:
+    """Reduce accumulated ``aux_metrics`` to the mean over a step's micro batches.
+
+    Losses may be summed across micro batches because ``mean_global_loss`` has
+    already weighted each by its share of the step's tokens. An auxiliary metric
+    carries no such weight, so it is averaged instead: for a per-token metric that
+    is the step's per-token mean when the micro batches hold equal token counts,
+    and unlike a token-share weighting it assumes nothing about which denominator
+    the metric used. Shared by the trainers that keep their own accumulation loop
+    so none of them can reduce it differently.
+    """
+    if not total_aux_metrics:
+        return {}
+    return {key: value / num_micro_steps for key, value in total_aux_metrics.items()}
+
+
 class BaseTrainer(Stateful, ABC):
     """
     Base trainer class for distributed model training.
@@ -653,19 +669,14 @@ class BaseTrainer(Stateful, ABC):
             callback.on_epoch_end(self.state)
 
     def on_step_begin(self, micro_batches=None, **kwargs):
-        # Every trainer funnels its step here, including the four that keep their
-        # own copy of the accumulation loop, so this is the one place that sees
-        # the step's micro-batch count for all of them. ``postforward`` needs it
-        # to normalize aux metrics; setting it per-loop instead would let the
-        # next trainer forget and silently report those metrics N times too high.
-        if micro_batches is not None:
-            self.num_micro_batches = len(micro_batches)
         for callback in self._callbacks:
             callback.on_step_begin(self.state, micro_batches=micro_batches, **kwargs)
 
-    def on_step_end(self, loss=None, loss_dict=None, grad_norm=None):
+    def on_step_end(self, loss=None, loss_dict=None, grad_norm=None, aux_metrics=None):
         for callback in self._callbacks:
-            callback.on_step_end(self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm)
+            callback.on_step_end(
+                self.state, loss=loss, loss_dict=loss_dict, grad_norm=grad_norm, aux_metrics=aux_metrics
+            )
 
     def preforward(self, micro_batch: List[Dict[str, Any]]) -> Dict[str, Any]:
         """Preprocess micro batches before forward pass.
@@ -691,8 +702,17 @@ class BaseTrainer(Stateful, ABC):
 
     def postforward(
         self, outputs: ModelOutput, micro_batch: Dict[str, torch.Tensor]
-    ) -> tuple[torch.Tensor, dict[str, torch.Tensor]]:
-        """Postprocess model outputs after forward pass."""
+    ) -> tuple[torch.Tensor, dict[str, torch.Tensor], dict[str, torch.Tensor]]:
+        """Postprocess model outputs after forward pass.
+
+        Returns the backward scalar, the losses behind it, and any diagnostics the
+        forward asked to have logged. The diagnostics travel in their own dict
+        because ``loss_dict`` carries a reduction contract they do not share:
+        ``mean_global_loss`` has already scaled each loss by its share of the
+        step's tokens, so summing that dict is the last step of a global token
+        mean. A metric folded in would inherit that summation, and anything later
+        taking ``sum(loss_dict.values())`` would train on it.
+        """
         loss_dict: Dict[str, torch.Tensor] = mean_global_loss(
             outputs.loss,
             self.micro_batch_token_len,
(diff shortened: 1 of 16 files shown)


FoolPlayer:
Both findings addressed in 04400cbbd.

**Loss-key collision.** Valid, and a real gap: the backward scalar stays correct because it is summed before the merge, so nothing diverges — the damage is exactly what you describe, callbacks publishing the auxiliary value as `training/<loss name>`. `postforward` now raises `ValueError` on any overlap, with a regression test.

**Output contract.** Documented in `docs/usage/trainer.md` under "Auxiliary Metrics": detached scalar values, micro-batch averaging (and the resulting "emit on every micro batch of a step, or none" rule), reserved loss names, and why no `*_loss` suffix is needed.

FoolPlayer:
@coderabbitai review

veomni/trainer/base.py
```diff
@@ -730,6 +737,37 @@ def postforward(
...
+            # but callbacks read ``loss_dict``, so they would report the auxiliary
+            # value under ``training/<loss name>`` -- a silently wrong loss curve.
+            collisions = sorted(aux_metrics.keys() & loss_dict.keys())
+            if collisions:
+                raise ValueError(
+                    f"aux_metrics keys {collisions} collide with loss keys "
+                    f"{sorted(loss_dict.keys())}. Loss names are reserved: rename the "
+                    "auxiliary metric so callbacks cannot report it as a loss."
+                )
+            loss_dict.update({key: value.detach() / num_micro_batches for key, value in aux_metrics.items()})
```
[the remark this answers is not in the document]
FoolPlayer: Valid, and the reserved set is larger than the three names listed. Fixed in 9e50d78b8.

Probing the real `on_step_end` showed two more collisions and, more usefully, that the two directions fail differently:

| name | outcome | damage |
| --- | --- | --- |
| `total_loss` | aux wins | real metric overwritten |
| `avg_effective_len` | aux wins | real metric overwritten |
| `avg_sample_seq_len` | aux wins | real metric overwritten |
| `grad_norm` | callback wins | aux dropped silently |
| `lr` | callback wins | aux dropped silently |

`total_loss` is assigned before `update(loss_dict)` so an auxiliary metric replaces it, whereas `grad_norm` and `lr` are assigned after and win instead. `avg_effective_len` / `avg_sample_seq_len` come from `EnvironMeter.step` already prefixed, and the closing `step_env_metrics.update(step_train_metrics)` lets the `training/` values win — so they are in the *corrupting* group, i.e. worse than the two the finding did list. The unprefixed env metrics (`mfu`, `flops_*`, memory) and the `multi_source/` keys are in other namespaces and cannot collide.

One deviation from the suggested placement: the set lives in `trace_callback.py` next to the code that decides the names, not as a literal in `base.py`. Hard-coding it in the trainer would make `base.py` restate callback internals and go stale silently the next time a metric is added to the callback. `tests/trainer/test_aux_metrics.py` now also derives the callback-owned names from what `on_step_end` actually publishes and asserts they equal the reserved set, so that staleness fails a test instead — verified by injecting an unreserved metric into the callback and watching only that test fail.

Docs and per-name regression cases updated as suggested.

veomni/trainer/base.py
```diff
@@ -730,6 +737,37 @@ def postforward(
             getattr(self, "global_micro_batches_token_len", None),
         )
         loss = torch.stack(list(loss_dict.values())).sum()
+        # Diagnostics a forward wants logged next to the losses. Merged *after* the
+        # sum above, because everything in ``loss_dict`` up to that line is the
+        # backward scalar; a metric merged earlier would silently join the objective.
+        # ``getattr`` rather than attribute access: most model outputs in the repo
+        # have no such field.
+        aux_metrics = getattr(outputs, "aux_metrics", None)
```
Coach257: `aux_metrics` should not be merged into `loss_dict`.

`loss_dict` is the backward objective: `mean_global_loss` already scales each term by its token share, so `train_step`'s `+=` is the last step of a global token-mean. Folding a diagnostic into that dict inherits the wrong reduction — hence the `/ num_micro_batches` patch, the reserved-key check, and the risk that a future `sum(loss_dict.values())` silently trains on the metric.

Please keep a separate dict end to end, named `aux_metrics` to match `outputs.aux_metrics`:

- `postforward` → `(loss, loss_dict, aux_metrics)` — extract and `.detach()`, no `/N`
- `train_step` → sum `loss_dict`; **mean** `aux_metrics` over micro-batches
- `on_step_end(..., aux_metrics=None)` — same shape as `grad_norm`, default empty so existing callbacks stay intact
- `EnvironMeterCallback` merges it beside `loss_dict` as `training/<key>`

Do not run these through `mean_global_loss` (`*_metrics` postfix or otherwise). That helper applies token / FSDP / SP scale for gradients, which is not a logging semantic. Still reject keys that collide with `loss_dict`: the wandb `training/` namespace is shared even if the dicts are not.

FoolPlayer: Agreed, and done in e53fa4401 — the separation removes the scaffolding rather than adding to it.

Implemented as proposed:

- `postforward` → `(loss, loss_dict, aux_metrics)`, detached, no `/N`
- `train_step` sums `loss_dict`, averages `aux_metrics`
- `on_step_end(..., aux_metrics=None)`, forwarded by all five trainer wrappers
- `EnvironMeterCallback` merges it beside `loss_dict`, before the prefixing, so it still gets the FSDP-group average
- nothing goes through `mean_global_loss`; collision check kept, against loss keys *and* `RESERVED_TRAINING_METRIC_NAMES`

Two notes on the result.

**The `/N` was load-bearing for a whole commit, which is now gone.** `postforward` reached for `num_micro_batches` via the instance, so the previous commit published it from `on_step_begin` to keep the four trainers with their own accumulation loop in sync. With the reduction moved to where the loop is, `num_micro_steps` is already a local, so that plumbing is reverted — `DiTTrainer` sets its own. The remaining risk was the three loops drifting, so the average goes through a shared `mean_aux_metrics` rather than being written out three times.

**`TextDPOTrainer` already has the bug this design prevents.** `forward_backward_step` puts `chosen_rewards`, `rejected_rewards`, `reward_accuracy` and `reward_margin` into `loss_dict` (`text_dpo_trainer.py:372-378`), and `train_step` accumulates `total_loss_dict[k] += v.item()` with no compensating scale (`:428-432`), so at `gradient_accumulation_steps > 1` a `reward_accuracy` bounded by 1 is reported above it. Out of scope here; I will send it as its own PR.

Test coverage follows the new shape: `postforward` returning the metric outside `loss_dict`, the sum of `loss_dict` staying clean, sum-vs-mean across all three `train_step` owners, partial emission, and the callback publishing `training/<key>` from the separate dict. `make quality` clean; the 16 `test_channel_loss_callback.py` failures on this branch are identical before and after the change (verified by diffing the failure sets) and are pre-existing on `main`.

Coach257 (approved):

PR #1111, opened 2026-08-24
[dist, parallel, ci] feat: context-parallel infrastructure for DeepSeek-V4

FoolPlayer:
### What does this PR do?

Adds the context-parallel (CP) infrastructure: a `cp_size` axis in the parallel state and
arguments, a new `veomni/distributed/context_parallel/` module with the window-ownership
arithmetic and collectives DeepSeek-V4's sparse attention needs, and a model allow-list so an
unsupported model refuses CP instead of silently training something wrong.

Infrastructure only — no model wiring. DeepSeek-V4's attention/compressor/indexer CP path is a
follow-up PR that depends on this one.

### Checklist Before Starting

- Search for relative PRs/issues and link here: builds on #981 (SP recompiles and device syncs),
  already merged.

### Test

- `pytest tests/parallel/context_parallel/` — 19 passed, covering window-ownership arithmetic,
  the collectives, the sharding helpers, the accelerator-config gate and the model gate.
- `make quality` — clean.
- Wired `tests/parallel/context_parallel` into `gpu_unit_tests.yml`.

### API and Usage Example

```yaml
train:
  accelerator:
    cp_size: 4
    ulysses_size: 1  # CP and Ulysses cannot be combined yet
```

`cp_size > 1` on a model that has no CP implementation raises at setup with the model name,
rather than running a silently wrong sharding. The same gate covers the omni encoder/decoder
path.

### Design & Code Changes

**Window ownership.** DeepSeek-V4 compresses KV in fixed-rate windows, so a sequence shard does
not correspond to a window shard. `veomni/distributed/context_parallel/sharding.py` owns that
arithmetic: which windows a rank produces, and where its rows land in global window order.

**Collectives.** `dsa_cp.py` gathers unevenly-sized compressed row counts into global order.
Ranks contribute different row counts because the window boundaries do not divide evenly, so
the gather is count-aware rather than a plain `all_gather`.

**`_Gather.backward` contiguity.** The gradient of a gathered `kv` arrives as a narrowed view of
the concatenation's gradient, which is contiguous only while the leading dims collapse at batch
1. At batch 2 the in-place `all_reduce` failed with "Tensors must be contiguous". The fix buys
contiguity and nothing else.

**Gating.** CP with Ulysses raises `NotImplementedError` rather than asserting, so the refusal
survives `python -O`. The model gate is an explicit allow-list in `veomni/models/auto.py`.

- Applied pre-commit checks
- Added/updated documentation
- Added tests to CI workflow

## Summary by CodeRabbit

- **New Features**
  - Added DeepSeek-V4 context-parallel support, including distributed attention communication, window sharding, halo exchange, and model compatibility checks.
  - Added configurable AdamW optimizer betas, defaulting to `(0.9, 0.95)`.
  - Added a way to check whether distributed parallel state is initialized.

- **Bug Fixes**
  - Improved validation for context-parallel sizes and unsupported combinations with Ulysses parallelism.
  - Fixed distributed gradient handling for non-contiguous tensors.

- **Breaking Changes**
  - Removed multimodal encoder and decoder configuration options from model arguments.


FoolPlayer:
Both findings addressed in 995dc51e3.

**Non-positive `cp_size`.** Valid, and the hole is slightly worse than described. Confirmed by repro:

- `TrainingArguments(accelerator=AcceleratorConfig(cp_size=-1))` with `WORLD_SIZE=1` is *accepted*, deriving `dp_size=-1` **and** `dp_shard_size=-1` — which is precisely the pair that then satisfies both of `ParallelState`'s checks, since the negatives cancel in the product and in `dp_replicate * dp_shard == dp_size`. So the invalid topology survives the full pipeline, not just one gate.
- `cp_size=0` raises a bare `ZeroDivisionError: integer modulo by zero` from the world-size modulo rather than naming the constraint.

Both boundaries now validate `cp_size >= 1` before the topology arithmetic, with parametrized regression tests for `0` and `-1` at each. The `ParallelState` test passes `dp_size=-1, dp_shard_size=-1` deliberately, so the rejection is attributable to the new guard rather than to the product check.

Scoped to `cp_size`, the field this PR introduces. `pp_size` / `ulysses_size` / `tp_size` have the same shape of hole on `main`; unlike `dp_replicate_size` / `dp_shard_size`, where non-positive means "derive it", those are all explicit divisors. Happy to fix them separately.

**Device-count helper.** Already fixed in 9d772bcd8, as noted.

FoolPlayer:
Separately from the review comments: GPU CI on the previous head caught a **real regression from this PR**, fixed in c5b6b91ce.

`tests/lora/test_moe_lora_ep2.py::test_independent_reset_lora_parameters_under_ep_shard` failed with `ValueError: The product of parallel sizes should be equal to the world size.` The chain: `build_foundation_model` → the new `check_context_parallel_supported` → `get_parallel_state()`, which in a process that never installed a parallel state does not return a state but *constructs* a default single-process one. That default's `dp_size=1` contradicts the real world size, so the topology check raises. This test spawns two ranks and calls `build_foundation_model` directly, so the gate — which runs for every model — turned it into a hard failure.

Fixed by adding `is_parallel_state_initialized()` and short-circuiting on it: no state installed means there is no context parallelism to gate.

Worth noting why the unit tests missed this: every test in `test_dsv4_cp_model_gate.py` patches `veomni.models.auto.get_parallel_state`, so none of them ever reached the real accessor's fallback. The new regression test deliberately does not stub it — it sets `_PARALLEL_STATE = None` with a two-rank world and asserts the gate is inert. Verified it fails without the fix.

Local runs after the fix: `tests/parallel/context_parallel/` 24 passed, `tests/lora/` 102 passed / 7 skipped.

FoolPlayer:
Rebased onto latest `main` (a4c872489). CI had stopped running on this PR entirely: it was `CONFLICTING` / `DIRTY`, so GitHub could not build the merge ref and none of the `pull_request` workflows fired — the last two pushes triggered only `Auto Label`, which is `pull_request_target`. Worth knowing as a general failure mode, since the PR page shows stale-but-green checks rather than "not run".

The conflict was #1082 removing the SeedOmni V1 model stack. That deleted `veomni/models/seed_omni/auto.py`, which one commit here modified to gate CP in the omni encoder/decoder construction path. Since that path no longer exists, the commit is obsolete and was dropped along with its five omni tests. The foundation-model gate in `veomni/models/auto.py` is unaffected and still covers every remaining construction path.

Post-rebase verification: `make quality` clean, `tests/parallel/context_parallel/` 20 passed, and the LoRA case that GPU CI failed on passes.

Coach257 (approved):

PR #1114, opened 2026-08-24
[model] fix: port Qwen3.5 linear-attn-mask override to transformers 5.9 Cache API

### LINKED ISSUE #1007: Qwen3.5 / Qwen3.5-MoE generated modeling: _update_linear_attn_maskoverride not ported to the transformers 5.9 call site — any cached forward crashes withAttributeError` on the Cache object
## Summary

The transformers-5.9 regeneration (#774) updated the call site of `_update_linear_attn_mask` to the new upstream signature, but the `method_override` body still implements the pre-5.9 signature. The second argument is now a `Cache` (e.g. `DynamicCache`), while the override body treats it as the old `cache_position` tensor and reads `.shape` on it, so the first cached forward — any `generate()` with `use_cache=True`, including the prefill step — crashes:

```
File ".../veomni/models/transformers/qwen3_5/generated/patched_modeling_qwen3_5_gpu.py", line 1858, in _update_linear_attn_mask
    if cache_position is not None and attention_mask.shape[-1] != cache_position.shape[-1]:
AttributeError: 'DynamicCache' object has no attribute 'shape'
```

Training is unaffected (`past_key_values` is always `None` there), which is presumably why this hasn't been caught.

## Affected files (current `main`)

All four generated variants carry the same mismatch (call site passes `past_key_values`, override body still reads it as `cache_position`):

| file | call site | stale override |
|---|---|---|
| `qwen3_5/generated/patched_modeling_qwen3_5_gpu.py` | L1806 | L1831 / L1858 |
| `qwen3_5/generated/patched_modeling_qwen3_5_npu.py` | L1784 | L1830 / L1857 |
| `qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_gpu.py` | L2001 | L2026 / L2053 |
| `qwen3_5_moe/generated/patched_modeling_qwen3_5_moe_npu.py` | L1986 | L2032 / L2059 |

The root source is the `method_override` for `Qwen3_5TextModel._update_linear_attn_mask` (and the MoE counterpart) in the patchgen configs — the generated files just inherit it.

## Repro (rev 8ca09d7, also current main; transformers 5.9.0)

```python
import torch
from transformers.models.qwen3_5 import Qwen3_5Config
from veomni.models.transformers.qwen3_5.generated.patched_modeling_qwen3_5_gpu import (
    Qwen3_5ForConditionalGeneration,
)

cfg = Qwen3_5Config()
cfg.text_config.hidden_size = 64
cfg.text_config.intermediate_size = 128
cfg.text_config.num_hidden_layers = 2
cfg.text_config.num_attention_heads = 4
cfg.text_config.num_key_value_heads = 2
cfg.text_config.head_dim = 16
cfg.text_config.vocab_size = 512
cfg.text_config.layer_types = ["linear_attention", "full_attention"]
cfg.vision_config.depth = 2
cfg.vision_config.hidden_size = 32
cfg.vision_config.intermediate_size = 64
cfg.vision_config.num_heads = 2
cfg.vision_config.out_hidden_size = 64

model = Qwen3_5ForConditionalGeneration._from_config(cfg, attn_implementation="sdpa").eval()
ids = torch.randint(0, 500, (1, 8))
mask = torch.ones_like(ids)
model.generate(ids, attention_mask=mask, max_new_tokens=4, use_cache=True)
# AttributeError: 'DynamicCache' object has no attribute 'shape'
```

(The crash happens in `Qwen3_5TextModel.forward` before the decoder layers, so it reproduces regardless of the varlen `cu_seq_lens_q` metadata.)

## Expected behavior / suggested fix

Upstream transformers 5.9 implements the cached-forward guard with cache metadata:

```python
if (past_key_values is not None and past_key_values.has_previous_state()) or (...):
    linear_attn_mask = None
```

Porting the override body to the new argument keeps the docstring's sync-free contract intact — `has_previous_state()` only reads host-side cache metadata, no GPU sync — while still dropping the `torch.all(attention_mask == 1)` short-circuit as the override intends:

```python
def _update_linear_attn_mask(self, attention_mask, past_key_values):
    if attention_mask is None:
        return None
    # Cached forward: the 2-D mask spans past + current tokens and would not
    # broadcast against the current-chunk hidden states. Metadata-only check.
    if past_key_values is not None and past_key_values.has_previous_state():
        return None
    return attention_mask
```

Happy to open a PR if useful.

### Information

- [ ] The official example scripts
- [ ] My own modified scripts

### Tasks

- [ ] An officially supported task in the `examples` folder (such as GLUE/SQuAD, ...)
- [ ] My own task or dataset (give details below)

## Repro (rev 8ca09d7, also current main; transformers 5.9.0)

```python
import torch
from transformers.models.qwen3_5 import Qwen3_5Config
from veomni.models.transformers.qwen3_5.generated.patched_modeling_qwen3_5_gpu import (
    Qwen3_5ForConditionalGeneration,
)

cfg = Qwen3_5Config()
cfg.text_config.hidden_size = 64
cfg.text_config.intermediate_size = 128
cfg.text_config.num_hidden_layers = 2
cfg.text_config.num_attention_heads = 4
cfg.text_config.num_key_value_heads = 2
cfg.text_config.head_dim = 16
cfg.text_config.vocab_size = 512
cfg.text_config.layer_types = ["linear_attention", "full_attention"]
cfg.vision_config.depth = 2
cfg.vision_config.hidden_size = 32
cfg.vision_config.intermediate_size = 64
cfg.vision_config.num_heads = 2
cfg.vision_config.out_hidden_size = 64

model = Qwen3_5ForConditionalGeneration._from_config(cfg, attn_implementation="sdpa").eval()
ids = torch.randint(0, 500, (1, 8))
mask = torch.ones_like(ids)
model.generate(ids, attention_mask=mask, max_new_tokens=4, use_cache=True)
# AttributeError: 'DynamicCache' object has no attribute 'shape'
```

(The crash happens in `Qwen3_5TextModel.forward` before the decoder layers, so it reproduces regardless of the varlen `cu_seq_lens_q` metadata.)

## Expected behavior / suggested fix

Upstream transformers 5.9 implements the cached-forward guard with cache metadata:

```python
if (past_key_values is not None and past_key_values.has_previous_state()) or (...):
    linear_attn_mask = None
```

Porting the override body to the new argument keeps the docstring's sync-free contract intact — `has_previous_state()` only reads host-side cache metadata, no GPU sync — while still dropping the `torch.all(attention_mask == 1)` short-circuit as the override intends:

```python
def _update_linear_attn_mask(self, attention_mask, past_key_values):
    if attention_mask is None:
        return None
    # Cached forward: the 2-D mask spans past + current tokens and would not
    # broadcast against the current-chunk hidden states. Metadata-only check.
    if past_key_values is not None and past_key_values.has_previous_state():
        return None
    return attention_mask
```

Eisenhower:
## Summary

Fixes #1007.

The transformers-5.9 regeneration updated the upstream call site of `_update_linear_attn_mask` to pass `past_key_values` (a `Cache`) instead of `cache_position` (a tensor), but the VeOmni `method_override` in `qwen3_5_gpu_patch_gen_config.py` still read the argument as a tensor and called `.shape` on it. Any cached forward — `generate()` with `use_cache=True`, including the prefill step — crashed with:

```
AttributeError: 'DynamicCache' object has no attribute 'shape'
```

Training itself was unaffected, since `past_key_values` is always `None` there — which is presumably why this went uncaught.

## Fix

Ports the override to the same Cache-metadata check upstream now uses (`past_key_values.has_previous_state()`), which is host-side only and preserves the original host-device-sync-avoidance the override exists for (see the updated docstring). `qwen3_5_moe` reuses this same function object (`qwen3_5_text_model_update_linear_attn_mask`), so fixing the one source function and regenerating fixes all 4 affected files (GPU/NPU × qwen3_5/qwen3_5_moe), matching the issue's own root-cause analysis.

## Verification

- Confirmed the exact crash on the pre-fix code with a direct call to the generated `_update_linear_attn_mask`, using a real `transformers.cache_utils.DynamicCache`:
  ```
  AttributeError: 'DynamicCache' object has no attribute 'shape'
  ```
- After the fix, verified correct behavior for all relevant cases (no `past_key_values`, fresh cache with no previous state, cache with previous state, `attention_mask=None`) — no crash, and return values match upstream's `has_previous_state()`-based semantics.
- `patchgen --check` passes (generated `.py`/`.diff` pairs match a fresh regen from the patched source, no drift).
- `ruff check` passes on all touched files.
- Regenerating all 29 patch configs (`patchgen --all --diff`) touches only the 4 files this bug affects — no incidental drift elsewhere.

## Test plan

- [x] Reproduced the exact `AttributeError` from #1007 against the unfixed override
- [x] Verified the fix resolves it and matches upstream `transformers==5.9.0` semantics for `_update_linear_attn_mask`
- [x] `patchgen --check` (drift/CI gate) passes
- [x] `ruff check` passes

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved cached text generation by detecting previous state through cache metadata.
  * Reduced unnecessary GPU synchronization during attention-mask preparation.

* **Bug Fixes**
  * Preserved non-cached attention masks without altering their contents.
  * Improved handling of attention masks during generation with cached states.

### DIFF
--- a/veomni/models/transformers/qwen3_5/generated/patched_modeling_qwen3_5_gpu.diff
+++ b/veomni/models/transformers/qwen3_5/generated/patched_modeling_qwen3_5_gpu.diff
@@ -1138,12 +1138,9 @@
 
 
  class Qwen3_5TextModel(Qwen3_5PreTrainedModel):
-@@ -1233,19 +1828,42 @@
-             past_key_values=past_key_values,
-         )
+@@ -1235,17 +1830,39 @@
 
--    def _update_linear_attn_mask(self, attention_mask, past_key_values):
-+    def _update_linear_attn_mask(self, attention_mask, cache_position):
+     def _update_linear_attn_mask(self, attention_mask, past_key_values):
          """
 -        NOTE: Left-padding is used for linear attention mask.
 -        No need for zeroing states when
@@ -1152,18 +1149,17 @@
 +        Build the attention mask passed to the linear-attention (gated DeltaNet) layers.
 +
 +        Upstream returns ``None`` — disabling the per-token zeroing in ``apply_mask_to_padding_states``
-+        — when ``cache_position[0] > 0`` (a cached forward) or ``torch.all(attention_mask == 1)`` (the
-+        batch has no padding). Both predicates read a 0-D GPU tensor and force an implicit ``.item()``
-+        / host-device sync on *every* forward, which serialises the host against the device in
-+        VeOmni's otherwise sync-free training step.
++        — when ``past_key_values.has_previous_state()`` is true (a cached forward) or
++        ``torch.all(attention_mask == 1)`` (the batch has no padding). The latter reads a 0-D GPU
++        tensor and forces an implicit ``.item()`` / host-device sync on *every* forward, which
++        serialises the host against the device in VeOmni's otherwise sync-free training step.
 +
 +        We keep the cached-forward branch — it is a correctness guard, not just an optimization:
 +        ``apply_mask_to_padding_states`` does ``hidden_states * attention_mask[:, :, None]``, and in a
 +        cached forward the 2-D ``attention_mask`` spans ``past + current`` tokens while ``hidden_states``
 +        only covers the current chunk, so the shapes wouldn't broadcast (or would broadcast wrongly for
-+        a 1-token decode step). But we detect it host-side from tensor shapes — ``attention_mask`` has
-+        ``shape[-1] == past + current`` whereas ``cache_position`` has ``shape[-1] == current`` — rather
-+        than reading ``cache_position[0]``, so no sync.
++        a 1-token decode step). ``has_previous_state()`` only reads host-side Cache metadata, so this
++        branch was never the sync source.
 +
 +        The all-ones short-circuit is the one we drop: returning the all-ones mask makes
 +        ``apply_mask_to_padding_states`` a no-op multiply, so it is equivalent to upstream's ``None``
@@ -1179,8 +1175,8 @@
 +        if attention_mask is None:
 +            return None
 +        # Cached forward (decode / continuation): see docstring — shapes wouldn't line up in
-+        # apply_mask_to_padding_states, and upstream returns None here. Detected from shapes only.
-+        if cache_position is not None and attention_mask.shape[-1] != cache_position.shape[-1]:
++        # apply_mask_to_padding_states, and upstream returns None here. Metadata-only check, no sync.
++        if past_key_values is not None and past_key_values.has_previous_state():
 +            return None
 +        return attention_mask
 +
@@ -1192,7 +1188,7 @@
 
 
  @auto_docstring
-@@ -1441,64 +2059,43 @@
+@@ -1441,64 +2058,43 @@
          self,
          pixel_values: torch.FloatTensor,
          image_grid_thw: torch.LongTensor | None = None,
@@ -1282,7 +1278,7 @@
          return special_image_mask, special_video_mask
 
      def compute_3d_position_ids(
-@@ -1564,13 +2161,18 @@
+@@ -1564,13 +2160,18 @@
          image_grid_thw: torch.LongTensor | None = None,
          video_grid_thw: torch.LongTensor | None = None,
          mm_token_type_ids: torch.IntTensor | None = None,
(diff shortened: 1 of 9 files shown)


Eisenhower:
Closing in favor of #1100 — same root-cause fix (I checked the diff: identical semantics, `past_key_values.has_previous_state()`), opened earlier (2026-08-21, three days before this one) and more complete: it adds a permanent regression test (`tests/models/test_qwen3_5_cache_mask.py`) wired into `gpu_unit_tests.yml`, which this PR doesn't have — I verified the fix with scratch scripts on a dev box instead of committing a lasting test. CI on #1100 is green.

Should have searched for PRs referencing this issue by number in the body (not just the title) before starting — same miss I made on #1059/#1086. Apologies for the duplicate.

PR #1120, opened 2026-08-25
[model, dist] feat: support Ulysses sequence parallelism in MiniMax H…

vvyuervv:
### What does this PR do?

> This PR adds Ulysses sequence parallelism (`ulysses_size > 1`) support to the MiniMax H3.

### Checklist Before Starting

#### Production Training (8 NPUs, 910C)

**Config**: 8-layer validation model (same code path as the 50-layer production model), `global_batch_size=8`, FSDP2, bf16, gradient checkpointing.

**Runs**:
- baseline 4p: dp=4, ulysses=1
- baseline 8p: dp=8, ulysses=1
- SP 8p: dp=4, ulysses=2

> **Data sharding caveat**: the offline loader shards data by dp_size, so different dp configs read different samples (8p baseline step-0 loss 11728 vs 4p 8635 — different sample sets). Precision alignment must compare same dp: baseline 4p (dp=4) vs SP 8p (dp=4).

#### Precision: per-step loss alignment (same dp=4 samples)

| step | baseline 4p | SP 8p | rel diff |
|---|---|---|---|
| 0 | 8635.36 | 8635.36 | bit-identical |
| 1 | 1440.12 | 1440.19 | 0.005% |
| 2 | 6507.35 | 6507.58 | 0.004% |
| 3 | 3183.56 | 3183.59 | 0.001% |
| 4 | 9722.08 | 9720.52 | 0.016% |
| 5 | 7373.04 | 7368.39 | 0.063% |
| 6 | 3243.64 | 3239.39 | 0.131% |
| 7 | 790.20 | 785.84 | 0.552% |
| 8 | 702.06 | 701.69 | 0.053% |
| 9 | 310.07 | 309.76 | 0.100% |

Step 0 bit-identical; subsequent divergence < 0.6% with identical loss trajectory. Sources of divergence: bf16 rounding + different all-reduce / FSDP gradient combination order (would be bit-identical in fp32).

#### Performance: steady-state per-step time (excl. first step and profiler steps)

| Run | steady s/it | notes |
|---|---|---|
| baseline 4p (dp=4) | ~9.0 | 2 samples/rank |
| baseline 8p (dp=8) | ~4.5 | 1 sample/rank, 2× cards |
| SP 8p (dp=4, ulysses=2) | ~5.0 | 2 half-samples/rank |

- SP vs same-card-count baseline 8p: +0.5 s/it (~11% comm overhead), in exchange for per-sample sequence scalability (dp already at 1 sample/rank with 16512-token packed sequences; SP splits the sequence, halving per-rank sequence length)
- SP vs 4p baseline: ~1.8× faster (effect of 2× cards, sub-linear with comm overhead)
- Performance is as expected: SP's benefit is sequence-length scalability + lower memory, not raw speed

#### Memory (per-rank peak allocation, torch memory snapshot)

| Run | peak | vs 8p baseline |
|---|---|---|
| baseline 4p | 33.45 GB | — |
| baseline 8p | 24.43 GB | — |
| SP 8p | 20.53 GB | -16% |

SP ranks hold only their half of the sequence activations; the memory drop matches design.

### API and Usage Example

### Design & Code Changes

- Applied pre-commit checks
- Added/updated documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved distributed attention handling for MiniMax H3, including more consistent gradient behavior and output processing.
  * Updated offline configuration to use the current data-processing settings.

* **Enhancements**
  * Improved DiT data-transform registration for supported online and offline training workflows.
  * Removed an unsupported training configuration restriction, allowing compatible training setups to load successfully.


veomni/models/diffusers/minimax_h3/__init__.py
```diff
@@ -1,3 +1,8 @@
+# Import side effect: registers the dit_offline / dit_online /
+# minimax_h3_online data transforms used by the offline / online training
+# flows (DiTTrainer looks "dit_offline" up at transform-build time).
+from veomni.data.multimodal.dit import data_transform  # noqa: E402,F401
```
Coach257: Move this import registry to `dit_trainer`

veomni/distributed/sequence_parallel/ulysses.py
```diff
@@ -229,6 +229,38 @@ def backward(ctx: Any, grad_output: Tensor) -> Tuple[None, Tensor]:
         )
 
 
+class _GatherSplitNoSum(torch.autograd.Function):
```
Coach257: **`_GatherSplitNoSum` backward looks off by a factor of `sp_world` — suggest reusing the existing `gather_outputs`**

I think the new `_GatherSplitNoSum` makes parameter gradients `sp_world`× too small, and that `_Gather` (via `gather_outputs`) is already the correct primitive for exactly this pattern.

**Why the all-reduce is needed here.** The docstring argues that because the loss is replicated across SP ranks, each rank's incoming grad is already the full `dL/dX`, so summing would double it. The first half is right — but in the replicated case the `all_reduce` in `_Gather.backward` isn't there to complete a partial gradient, it's there to cancel FSDP's averaging over the SP dimension. FSDP2 divides gradients by the size of the reduce mesh, and `dp_shard_sp` (the param-sharding mesh) includes `ulysses` whenever `ulysses_size > 1` — see `veomni/distributed/parallel_state.py:572` and `585-586`. The codebase documents this explicitly in `veomni/utils/loss_utils.py:57-58`: *"fsdp by default divides gradients by its size, so we need to multiply by fsdp_size"*.

Walking through SP=2, dp_shard=1, against an SP=1 single-rank baseline with param grad `G`:

- With `_Gather`: both ranks see the same `g` at the gather; all-reduce gives `2g`; the slice gives rank `r` its `2·g[own rows]`; block param grads are `2·G_r` where `G_0 + G_1 = G`; FSDP averages over the 2-rank mesh → `(2G_0 + 2G_1)/2 = G`. Matches the baseline.
- With `_GatherSplitNoSum`: rank `r` gets `g[own rows]` → `G_r`; FSDP averages → `(G_0 + G_1)/2 = G/2`. In general `G/sp_world`.

**Both existing SP paths already apply a `× sp_world`, just in different places.** The replicated-loss path (all the diffusion models) gets it from the `all_reduce` in `_Gather.backward`, which on identical per-rank grads is exactly `× sp_world`. The sharded-loss path (text) gets it from the explicit `ctx.sp_world_size *` factor in `ReduceLoss.backward`, `veomni/distributed/sequence_parallel/loss.py:62`. The MiniMax H3 offline path has neither — `DiTTrainer.postforward` (`veomni/trainer/dit_trainer.py:461-464`) only divides by `num_micro_batches` and sums, and never goes through `reduce_sequence_parallel_loss`.

**Repo precedent for the identical premise.** Every other DiT here gathers before the loss and uses `gather_outputs`: `wan_t2v/wan_transformer/modeling_wan_transformer.py:341`, `ltx2_3/ltx_transformer/modeling_ltx2_3_transformer.py:176-178`, `qwen_image/qwen_image_transformer/modeling_qwen_image_transformer.py:300`, `transformers/flux/modeling_flux.py:670-671`, `transformers/wan/modeling_wan.py:631`. The Wan comment even states the same premise this PR relies on — *"every rank holds the full sequence so that the loss is identical across SP ranks"* — and still uses the summing backward.

**Suggested change.** Replace the two calls at `minimax_h3_core/minimax_h3_dit.py:604-605` with `gather_outputs(..., gather_dim=0, group=sp_group)` and drop `_GatherSplitNoSum` entirely. That also removes the cross-package import of a private symbol at `minimax_h3_dit.py:14`, and sidesteps a latent trap in the new class: its `forward` allocates recv buffers via `torch.empty_like`, which assumes every rank's local shape is identical, whereas `_all_gather` (`ulysses.py:34-57`) exchanges shapes explicitly. The assumption happens to hold at this callsite thanks to the pad-to-`unit` above, but it wouldn't survive reuse.

**Why CI wouldn't catch this.** `tests/parallel/ulysses/test_all_gather.py::test_grad_aligned` covers the sharded path (gather → `slice_input_tensor` → local loss → `reduce_sequence_parallel_loss`), not "gather → full replicated loss". And the config shipped in this PR sets `ulysses_size: 1` (`configs/dit/minimax_h3_fl2va_offline.yaml:89`), so no default run enters this branch.

**Verification.** I haven't been able to run this, so please sanity-check with one step at `ulysses_size: 1` vs `ulysses_size: 2` on the same data, comparing `grad_norm`. If it's currently off by 2× and aligns after switching to `gather_outputs`, that confirms it. Given SP is now a supported config for this model, a test next to `test_grad_aligned` covering the replicated-loss variant would be worth adding either way.

Coach257 (approved):

PR #1121, opened 2026-08-25
[ops] feat: external ops-backend plugin loader via entry points

zhangxin81:
## Summary

Adds a **generic external ops-backend plugin mechanism**: third-party kernel packages can register **opt-in** backends for any existing op via the standard Python entry-point group `veomni.ops_backends`, with **zero VeOmni-side changes per package**. This generalizes the pattern VeOmni already uses for the external `liger_kernel` package (`BackendSpec.entry` + `requires`): after this one-time mechanism lands, shipping a new external kernel package needs no PR here — `pip install <pkg>` makes its backend names valid values for the matching `*_implementation` config fields (always opt-in; no default ever changes). External training-kernel projects (e.g. the nabla kernels sketched in #1105) could then integrate as versioned installable packages instead of source-tree copies, keeping kernel maintenance with the kernel authors.

This PR contains **no kernels and no reference to any specific third-party package** — only the mechanism, its docs, and CPU-runnable tests.

## What's in the change

| file | change |
|---|---|
| `veomni/ops/config/_plugin_loader.py` | **new** — discovers `veomni.ops_backends` entry points once at `veomni.ops` import (after built-in ops register); validates each plugin's **pure-data** declaration module (`VEOMNI_PLUGIN_API_VERSION` + `VEOMNI_OPS_BACKENDS` payload) fully before registering anything (**atomic per plugin**); translates the payload into `BackendSpec` / `KernelSpec`; any plugin failure (import error, bad payload, name conflict) degrades to a warning and the plugin is skipped |
| `veomni/ops/config/registry.py` | `extend_op_backends()`: merge external backends into a registered op (needed because `register_op` rejects a differing spec for a known op); `_check_requires()`: `requires` generalizes from a closed token whitelist (`liger_kernel`/`torch_npu`/`triton`) to any package name via plain importability — the three known tokens keep their dedicated availability helpers |
| `veomni/ops/__init__.py` | mount: one `load_ops_backend_plugins()` call after built-in registrations |
| `veomni/utils/env.py` | `VEOMNI_OPS_PLUGINS` env default (`0` = kill-switch, restores built-in-only behavior exactly) |
| `veomni/ops/README.md` | "External ops-backend plugins" section: protocol, guarantees, example |
| `tests/ops/test_ops_plugin_loader.py` + `tests/ops/plugin_fixtures/` | **new** — 16 CPU-runnable tests with injected entry points (no real plugin package needed) |

## Design guarantees

- **Opt-in only.** The payload schema has no default/fallback keys; a plugin can make a backend name resolvable but can never touch any op's default.
- **Pure-data declarations.** The declaration module needs no veomni import and no heavy deps; real entry targets are imported lazily at bind time (same as `liger_kernel` entries today).
- **Never fatal.** Plugin problems cannot affect users who didn't install the plugin; users who did get the existing clear errors (`ValueError: ... not a registered backend` / `RuntimeError: package '...' is not installed`).
- **Zero cost when unused.** No plugin installed → the entry-point scan finds nothing and behavior is identical to today. `VEOMNI_OPS_PLUGINS=0` skips even the scan.

## Usage (third-party package side)

```toml
# plugin package's pyproject.toml
[project.entry-points."veomni.ops_backends"]
my_kernels = "my_kernels.veomni_plugin"
```

```python
# my_kernels/veomni_plugin.py — pure data
VEOMNI_PLUGIN_API_VERSION = 1

VEOMNI_OPS_BACKENDS = {
    "ops": {  # merged into OpSpec.backends (config-field dispatch)
        "rms_norm": {
            "my_backend": {"entry": "my_kernels:MyRMSNorm",
                            "requires": ["my_kernels"]},
        },
    },
    "kernels": [  # merged into KERNEL_REGISTRY (OpSlot dispatch)
        {"name": "my_backend", "op_name": "rms_norm", "variant": "standard",
         "factory": "my_kernels:rms_norm", "device_type": "gpu",
         "description": "fused RMSNorm (Triton fwd+bwd)"},
    ],
}
```

```yaml
model:
  ops_implementation:
    rms_norm_implementation: my_backend   # opt-in; default unchanged
```

## Test plan

- `pytest tests/ops/test_ops_plugin_loader.py` — **16/16 passed** (macOS arm64, py3.12, torch 2.13 CPU; no accelerator needed). Covers: happy path (both registries + backend listed in misconfiguration errors + PER_MODEL end-to-end bind + eager untouched), rejection matrix (API-version mismatch, missing/unknown payload keys, unknown op, backend-name conflict with built-ins, kernel-name conflict, atomicity of partially-valid payloads, broken-import isolation, one bad plugin doesn't block others), kill-switch, re-entry safety, `requires` generalization.
- Regression: `tests/ops/test_kernel_registry_sanity.py` **24/24** and `tests/ops/test_eager_kernels_cpu.py` **10/10** unaffected.
- `ruff check` + `ruff format --check` clean.

## Scope & limitations

- Mechanism only — no concrete plugin package is added or referenced; per-package integration is up to those packages.
- Plugin declaration protocol is versioned (`VEOMNI_PLUGIN_API_VERSION`, exact match). Adding optional payload keys later can stay v1-compatible; semantic changes bump the version.
- GLOBAL-scope `side_effect` is supported per `BackendSpec` semantics; per-model-only keys (`replace_forward` / `entry_is_factory` / `target_override`) on a GLOBAL op are rejected at load time instead of being silently ignored.

## Summary by CodeRabbit

* **New Features**
  * Added support for optional external operation-backend plugins.
  * Plugins can contribute configuration backends and hardware-specific kernels.
  * Added validation, conflict detection, safe registration, package checks, and failure isolation.
  * Added an environment setting to disable plugin loading.

* **Documentation**
  * Documented plugin discovery, registration requirements, validation, and configuration.

* **Tests**
  * Added coverage for valid, invalid, conflicting, failing, and partially registered plugins.

### DIFF
--- a/tests/ops/plugin_fixtures/__init__.py
+++ b/tests/ops/plugin_fixtures/__init__.py
@@ -1,0 +1,24 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Fake ops-backend plugin declarations for the loader tests.
+
+Importable as ``plugin_fixtures.*`` (``tests/ops`` is put on ``sys.path`` by
+``test_ops_plugin_loader.py``). Each module mimics one plugin scenario.
+"""
+
+from .kernels import FakeRMSNorm, fake_rms_norm
+
+
+__all__ = ["FakeRMSNorm", "fake_rms_norm"]
--- a/tests/ops/plugin_fixtures/backend_conflict.py
+++ b/tests/ops/plugin_fixtures/backend_conflict.py
@@ -1,0 +1,26 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Invalid plugin: backend name collides with a built-in (``liger_kernel``)."""
+
+VEOMNI_PLUGIN_API_VERSION = 1
+
+VEOMNI_OPS_BACKENDS = {
+    "ops": {
+        "rms_norm": {
+            "liger_kernel": {"entry": "plugin_fixtures.kernels:FakeRMSNorm"},
+        },
+    },
+    "kernels": [],
+}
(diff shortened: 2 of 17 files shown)


Coach257:
Ops will be refactored: #1142

Issue #1061, opened 2026-08-18
[trainer, model] refactor: extract VeOmniModelRuntime from BaseTrainer


Coach257:
Part of #1057. **Depends on:** #1060.

## Why

This is the structural half of the `main` ↔ `omniv2` divergence.

`BaseTrainer` currently mixes two unrelated responsibilities: **model-bound** work (build the model, freeze it, apply LoRA, parallelize, load weights, build the optimizer and lr-scheduler, clip gradients, save/load its checkpoint) and **job-bound** work (distributed init, dataloader, the train loop, callbacks, logging). With one model per job the mix is harmless. With N models per job it is fatal — an omni job needs N of the first and exactly one of the second.

`omniv2` works around this today by composition: `ModuleRuntime` holds a bare `BaseTrainer` in `self.base` and calls into `base._build_model`, `base._freeze_model_module`, `base._setup_lora`, `base._build_parallelized_model` — private methods, on an instance that is never actually trained. That works, but it hard-couples `ModuleRuntime` to `BaseTrainer`'s private surface and leaves two divergent copies of the build sequence.

Extracting a `VeOmniModelRuntime` that owns exactly the model-bound half turns that into plain inheritance: `BaseTrainer` composes one, `ModuleRuntime` subclasses it. Landing it on `main` **before** the V2 code means #1064 ships `ModuleRuntime` in its final shape instead of shipping the `self.base` workaround and refactoring it later.

## Scope

New `veomni/models/model_runtime.py` defining `VeOmniModelRuntime`. It lives under `veomni/models/` rather than `veomni/trainer/` because everything it owns is model-bound, and because its only other subclass — `ModuleRuntime` — already lives at `veomni/models/seed_omni/accelerator/module_runtime.py`; putting the base class in `trainer/` would make `models/` import `trainer/` for the core model build path. `VeOmniModelRuntime` owns:

- `build_model()` — meta-init via the registry-aware loader
- `freeze_model()` / `setup_lora()`
- `build_parallelized_model()` — FSDP2/DDP wrap + weight load, incl. `skip_hf_weight_load` (constraint 26) and the `customized_setup_lora` / `customized_build_parallelize_model` hooks
- `build_optimizer()` / `build_lr_scheduler()` (incl. `_collect_muon_kwargs`)
- `clip_grad_norm()`
- `save_checkpoint()` / `load_checkpoint()` delegation
[...]

PR #1125, opened 2026-08-26
[BREAKING][trainer, model] refactor: extract model-bound logic into VeOmniModelRuntime

Coach257:
### What does this PR do?

Split a training job into two objects. **`VeOmniModelRuntime` owns everything bound to one model. `BaseTrainer` holds one as `trainer.model` and keeps everything bound to the job** (of which there is exactly one).

This is the cut `OmniTrainer` needs: several models in one job cannot inherit several copies of `_build_model` / `_build_optimizer` / clip. DPO in this PR is the first consumer — policy and frozen reference are two runtimes under one trainer.

Part of #1057 / #1061. Depends on #1090 (`AcceleratorConfig` self-resolving and model-scoped) and #1181 (checkpoint step-dir layout), both already on `main`. This PR does **not** change the on-disk checkpoint format; it only binds `ModelCheckpointManager` to the runtime and fans I/O out through the trainer.

#### Scope: trainer before vs after

On `main`, `BaseTrainer` is both the job and the model. `_setup` registers the mesh, then meta-init → freeze/LoRA → assets → data pipeline → FSDP wrap + weight load → optimizer → lr scheduler all live on the trainer. `trainer.model` is the raw `nn.Module`.

| | Before (`main`) | After (this PR) |
| --- | --- | --- |
| Process group, seed, job mesh `"base"` | `BaseTrainer._setup` | `BaseTrainer._setup(args)` — same name, now a staticmethod: it runs before any model exists |
| Meta-init, freeze/LoRA, FSDP/DDP wrap, weight load, optimizer, grad clip | `BaseTrainer._build_*` | **`VeOmniModelRuntime`** |
| Tokenizer / processor / `model_assets` / chat template | `BaseTrainer._build_model_assets` | **`VeOmniModelRuntime._build_model_assets`** (template name is `model.chat_template`) |
| LR scheduler | Trainer, after the dataset sized `train_steps` | Trainer still sizes `train_steps`; **runtime** builds the schedule |
| Data transform, dataset, dataloader | `BaseTrainer` | `BaseTrainer` (job-bound, unchanged) |
| `train_step` / `train()` / callbacks | `BaseTrainer` | `BaseTrainer`. Forward / post-forward / backward wrap `use_parallel_state(self.model.parallel_state)` — not inside `runtime.__call__` |
| `trainer.model` | `nn.Module` | **`VeOmniModelRuntime`** (module at `trainer.model.model`; unshadowed module APIs still forward) |
| Checkpoint **when** | Same callback that also wrote files | `CheckpointCallback` — cadence only. Calls `trainer.save_model_assets()` / `trainer.load()` / `trainer.save_*` / `trainer.wait_for_pending_save()` |
| Checkpoint **what** | Callback reached into trainer fields | Trainer fans out to each runtime (`TextDPOTrainer` → policy only) |
| Checkpoint **how** | Same callback | **`ModelCheckpointManager`** on the runtime (async drain, `empty_cache`, barrier). On-disk layout is #1181's |
| Job cursor (dataloader, rng, meters) | Mixed into the model checkpoint | **`GlobalStateCallback`**, trainer-only; per-rank `trainer_state_rank_{N}.pt`. Does not store model topology |

A model whose build differs subclasses the **runtime**, not the trainer's private `_build_*` list: `VLMModelRuntime` (tower freeze, ViT lr), `DiTModelRuntime` (frozen condition model; embedding-only skips parallelize), `DPOReferenceModelRuntime` (frozen eval replica, no optimizer / checkpoint).

#### Why this split

- **One model = one object with a complete lifecycle.** A job that holds two models (DPO today, Omni next) constructs each the same way and does not fork the trainer.
- **Override the step that differs.** VLM / DiT / DPO-reference changes stay on a runtime subclass. The trainer loop, data pipeline, and callbacks stay shared.
- **Checkpointing scales to N models.** Callbacks do not learn about a second handle. The trainer fans out *what*; each runtime's manager owns *how*. Multi-module assets nest under `model_assets/<module>/`.
- **The runtime is testable without a trainer** (`tests/models/test_model_runtime.py`).

### Checklist Before Starting

- Search for relative PRs/issues and link here: #1057, #1061, #1090 (merged), #1181 (merged), this PR #1125

### Test

Extended existing CI-enumerated tests from `trainer.xxx` to `trainer.model.xxx` (LoRA, dummy-forward, compile, ulysses, saveload, channel-loss, `tests/models/test_models_patch.py`, `tests/checkpoints/test_checkpoint_callback.py`).

Covered by the directory-level `tests/checkpoints/` CI job:

- `tests/checkpoints/test_global_state_callback.py` — per-rank job-state resume

Added under `tests/models/` (not directory-level; **not yet listed** in `gpu_unit_tests.yml` / `npu_unit_tests.yml`):

- `tests/models/test_model_runtime.py` — standalone runtime construction
- `tests/models/test_runtime_chat_template.py` — chat template built inside `_build_model_assets`

Full training-curve parity is not re-measured here; the train loop is intended to be unchanged.

### API and Usage Example

`trainer.model` is a `VeOmniModelRuntime`. Unshadowed `nn.Module` APIs still forward (`trainer.model.config`, `trainer.model.parameters()`, `trainer.model(**batch)`). The raw module is `trainer.model.model`.

```
# before                                      # after
trainer.tokenizer                             trainer.model.tokenizer
trainer.processor                             trainer.model.processor
trainer.optimizer                             trainer.model.optimizer
trainer.lr_scheduler                          trainer.model.lr_scheduler
trainer.model_assets                          trainer.model.model_assets
trainer.save_model_assets()                   trainer.save_model_assets()   # fans out to runtime(s)
trainer.model.save_hf_or_lora(state, assets)  trainer.save_hf_or_lora(state, stage=stage)
data.chat_template                            model.chat_template           # default None, not "default"
```

Custom model build — subclass the runtime, return it from `_build_model_runtime`:

```python
class VLMModelRuntime(VeOmniModelRuntime):
    def _freeze_model_module(self):
        ...
    def _build_optimizer(self, param_groups=None):
        ...

class VLMTrainer:
    def _build_model_runtime(self) -> VLMModelRuntime:
        return VLMModelRuntime(
            self.base.args.model, "base",
            train=self.base.args.train,
        )
```

Frozen DPO reference always copies the policy's `model`. A custom `reference_model` config is not supported.

### Design & Code Changes

- `veomni/models/model_runtime.py` — construct = register this model's mesh (`setup`) + `_build_model` / `_freeze_model_module` / `_build_parallelized_model` / `_build_optimizer` under `use_parallel_state(self.model_name)` + assets + checkpoint handle. `__call__` is a plain forward.
- `veomni/models/checkpoint_manager.py` — runtime-bound DCP / HF / LoRA I/O on #1181's layout. `save_hf_or_lora` drains pending async DCP first. Train-end still `trainer.wait_for_pending_save()` when HF export is skipped.
- `BaseTrainer` drops pass-through properties (`tokenizer`, `processor`, `optimizer`, …). `TextTrainer` / `VLMTrainer` / `DiTTrainer` / `TextDPOTrainer` compose `BaseTrainer` in `self.base` and a runtime in `.model`.
- **Breaking:** `chat_template` moves from `data` to `model`. Old `data.chat_template` is rejected as an unknown key. Default is `None` (was `"default"`).
- `init_parallel_state_from_config` is the production mesh entry; tests still call `_init_parallel_state` for topologies no job config can express.
- `runtime.parallel_state` is a by-name registry lookup, never a stored `ParallelState` object.

- Applied pre-commit checks
- Added/updated documentation (`docs/usage/trainer.md`, `docs/usage/basic_modules.md`, `docs/usage/arguments.md`, `docs/design/local_parallel_state.md`, `.agents/knowledge/architecture.md`)
- Tests: existing CI-enumerated files updated; `tests/checkpoints/` covers the new global-state test. The two `tests/models/test_runtime*.py` files are not wired into the unit workflows yet


.agents/knowledge/architecture.md
```diff
@@ -12,7 +12,7 @@ veomni/
 │   ├── multimodal/     Vision, audio, video preprocessing and chat templates
 │   └── diffusion/      Diffusion model data loading
 ├── distributed/        All parallelism strategies
-│   ├── parallel_state.py   init_parallel_state(), ParallelState, device mesh setup
+│   ├── parallel_state.py   init_parallel_state_from_accelerator(), ParallelState, mesh setup
```
Coach257: rename as `init_parallel_state_from_config`

.agents/knowledge/architecture.md
```diff
@@ -93,27 +97,75 @@ veomni/
...
 ```
 
-`BaseTrainer` provides the composable training skeleton:
-- `build_model()` -> model construction and parallelization
-- `build_dataloader()` -> data pipeline setup
-- `build_optimizer()` / `build_lr_scheduler()` -> optimization
+`VeOmniModelRuntime` contributes the model-bound half. `setup()` runs on construction — registering the mesh, then building over it:
+- `build_model()` -> meta-init through the registry-aware loader
+- `freeze_model()` / `setup_lora()` -> trainable surface
+- `build_parallelized_model()` -> FSDP2/DDP wrap + weight load
```
Coach257: unify name

.github/workflows/gpu_unit_tests.yml
```diff
@@ -162,6 +165,8 @@ jobs:
       - name: Run checkpoint callback unit tests
         run: |
           uv run --frozen pytest -s -x tests/checkpoints/test_checkpoint_callback.py
+          uv run --frozen pytest -s -x tests/checkpoints/test_model_checkpoint_manager.py
```
Coach257: too many ci, merge checkpoint related ci to test_checkpoint_callback. As test_checkpoint_callback calls checkpoint_manager, maybe no need to add a single test.

veomni/trainer/base.py
```diff
@@ -311,172 +295,118 @@ def __init__(self, args: VeOmniArguments):
...
+        if is_nccl_backend(backend) and os.getenv("VEOMNI_DESTROY_NCCL_ON_EXIT", "0") != "1":
+            logger.info_rank0(
+                "Skipping explicit NCCL process-group destroy on normal trainer exit. "
+                "Set VEOMNI_DESTROY_NCCL_ON_EXIT=1 to restore the previous teardown behavior."
             )
+            return
+
+        synchronize()
+        dist.destroy_process_group()
+        clear_parallel_state()
```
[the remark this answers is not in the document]
Coach257: Won't clear the registry while NCCL groups are still alive.

The early return is intentional: skip destroy_process_group unless VEOMNI_DESTROY_NCCL_ON_EXIT=1. Clearing the registry first would drop the Python handles while the groups still exist; a later init in the same process would then try to create a second mesh/group set for the same name. The duplicate-name guard is the safety net for that path. clear_parallel_state() stays after destroy_process_group() on the teardown that actually destroys groups.

veomni/trainer/callbacks/global_state_callback.py
```diff
@@ -0,0 +1,169 @@
...
+        load_path = args.train.checkpoint.load_path
+        if load_path is None:
+            return None
+
+        state_path = global_state_path(load_path, self.rank)
+        if not os.path.exists(state_path):
+            logger.warning_rank0(f"No trainer state at {state_path}; resuming weights only.")
+            return None
+
+        global_state = torch.load(state_path, map_location="cpu", weights_only=False)
```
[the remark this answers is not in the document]
Coach257: weights_only=False is required for this payload.

Global state includes MultiSourceCounterItem and dataloader sampler objects, which torch.load(weights_only=True) rejects. These files are written by this trainer on resume, not loaded from untrusted input. Turning every custom object into primitives across the meter and dataloader serializers is a separate serialization change, not a load-flag flip in this PR.

veomni/trainer/text_dpo_trainer.py
```diff
@@ -129,108 +127,156 @@ class VeOmniDPOArguments(VeOmniArguments):
...
-            init_device=args.model.accelerator.init_device,
-            ops_implementation=args.model.ops_implementation,
-            # The same overrides the policy model is built with. Without this the
-            # reference reads only ``config.json`` and the two architectures can
-            # diverge on anything set under ``model_config`` -- which is where a
-            # model-level training objective such as DeepSeek-V4's
-            # ``dsa_indexer_loss`` is switched off, so the policy would honour the
-            # override and its reference would not.
-            config_kwargs=args.model.model_config,
         )
```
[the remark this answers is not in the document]
Coach257: Addressed. Reference construction now checks ulysses_size * cp_size and dp_size against the policy before the runtime is built.

veomni/models/checkpoint_manager.py
```diff
@@ -179,14 +200,20 @@ def save_dcp(self, state: "TrainerState") -> None:
...
+        clears.
         """
         extra_state = self._extra_state(state)
 
         helper.empty_cache()
         self.checkpointer.save(
             self.config.save_path,
-            {"model": self.trainer.model, "optimizer": self.trainer.optimizer, "extra_state": extra_state},
+            {"model": self.runtime.model, "optimizer": self.runtime.optimizer, "extra_state": extra_state},
             global_steps=state.global_step,
```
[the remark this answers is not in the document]
Coach257: Won't fold checkpoint_subfolder into the DCP save path here.

DCPCheckpointer.save() has no subdir argument. Staging is keyed on the run-root `path` plus `global_steps` — that is the #1179 contract. Passing `<save_path>/global_step_N/<subfolder>` as `path` would give each module/step a fresh staging directory, and a kill mid-write would strand a copy no later save clears.

checkpoint_subfolder is "" on this PR. Omni should add a checkpointer `subdir=` that still keys staging on the run root, then save_dir()/load_dir() and the actual shards will share one layout.

veomni/trainer/callbacks/checkpoint_callback.py
Luosuu, line 1: In description, there are `ModelDcpCallback`/`ModelHfCallback` but I do not see them?

veomni/trainer/base.py
```diff
@@ -266,16 +259,7 @@ class BaseTrainer(Stateful, ABC):
...
-    model_config: PretrainedConfig = PretrainedConfig()
-    tokenizer: PreTrainedTokenizerBase = None
-    processor: ProcessorMixin = None
-    chat_template: ChatTemplate = None
-    model_assets: List[Any] = []
-
-    # Training components
-    optimizer: Optimizer = None
-    lr_scheduler: LRScheduler = None
+    model: VeOmniModelRuntime = None
```
Luosuu, line 262: Maybe `model_runtime: VeOmniModelRuntime` ?

veomni/trainer/base.py
```diff
@@ -310,158 +293,114 @@ def __init__(self, args: VeOmniArguments):
...
-            # Build optimizer and lr scheduler
-            self._build_optimizer()
-            self._build_lr_scheduler()
-            # Build training context
-            self._build_training_context()
-            # Initialize callbacks
-            self._init_callbacks()
-
-    def _setup(self):
+        self.device = self._setup(args)
```
Luosuu, line 296: maybe `_setup_distributed()` as stated in PR description?

veomni/trainer/base.py
Coach257: I'd like to keep `model` the same name it hase on `main`. The runtime forwards unshadowed `nn.Module` APIs, so `trainer.model.config`, `trainer.model.parameters()` and `trainer.model(**batch)` all read exactly as before; that forwarding is what kept the migration off ~100 call sites. Renaming would either keep the forwarding (odd on something explicitly called a runtime) or drop it and break far more than this PR needs to. The raw module stays reachable at `trainer.model.model`.

veomni/trainer/base.py
Coach257: I want to keep. `_setup` as the name it has on `main`, so keeping it means this PR doesn't churn a call site that didn't need to change. What actually changed is the signature: it's now a `staticmethod` taking `args`, because it runs before any model exists and no longer has a built trainer to read from. I'll fix the description to say that instead of promising a rename.

veomni/trainer/callbacks/checkpoint_callback.py
Coach257: These two are merged now. Pr description updated.

veomni/trainer/dit_trainer.py
Coach257, line 138: TODO: These args could be moved to base args.

Coach257:
@coderabbit

FoolPlayer (commented):
Nice split — the runtime/trainer boundary reads clean, and the config + docs migration is thorough. My concern is that a few callers weren't updated for `trainer.model` no longer being an `nn.Module`, and two of them fail silently. Details inline; the channel-loss and DPO `base.model` ones look blocking to me.

veomni/models/model_runtime.py
```diff
@@ -0,0 +1,558 @@
...
+            self._build_optimizer()
+        self._build_model_assets()
+        self.build_checkpoint()
+
+    def __getattr__(self, name: str) -> Any:
+        """Forward unshadowed :class:`torch.nn.Module` APIs to the wrapped model."""
+        model = object.__getattribute__(self, "model")
+        if model is None:
+            raise AttributeError(f"{type(self).__name__} has no attribute {name!r}, and its model is not built yet.")
+        return getattr(model, name)
```
FoolPlayer, line 154: This forwards attribute *reads*, but there is no `__setattr__`, so writes land on the handle — and that silently kills channel loss.

`ChannelLossCallback.on_train_begin` still calls `install(self.trainer.model)` (`channel_loss_callback.py:890`). `_resolve_loss_fn_host` can't unwrap the handle (MRO is just `(VeOmniModelRuntime, object)`, and an FSDP2 root has no `.module`), so it returns the handle itself; `host.loss_function = self._wrapped_loss_fn` (L168) then writes into the handle's `__dict__` and the module keeps its original loss. `_wrap_causal_loss_opslots` also derives candidates from `type(root).__mro__`, so it finds no `veomni_causal_lm_loss` slot either. Both interception paths are dead — but `_installed` is `True` because the *read* of `_original_loss_fn` succeeded, so it logs success and emits no metrics.

Suggest `install(self.trainer.model.model)`, plus an assert after the assignment that the patch actually stuck.

veomni/models/model_runtime.py
```diff
@@ -0,0 +1,558 @@
...
+        from .auto import build_processor
+
+        self.model_assets = [self.model_config]
+
+        path = self.args.tokenizer_path
+        try:
+            loaded = build_processor(path, **(self.args.processor_config or {}))
+        except Exception as e:  # noqa: BLE001 — surfaced later by whoever reads text
+            logger.warning_once(f"{type(self).__name__}: no preprocessor loaded from {path}: {e}.")
+            loaded = None
```
FoolPlayer, line 299: Bare `except Exception` combined with `warning_once` (which is gated on `LOCAL_RANK == 0`, `logging.py:162`) means a typo'd `tokenizer_path` or one transient object-store read leaves `tokenizer` / `processor` / `chat_template` all `None`, the build continues, and the job dies later inside the data transform — with no log line at all on non-zero local ranks. The old text path called `build_tokenizer` unguarded and failed loudly at the right place.

Could this be narrowed to the "this repo ships no preprocessor" case you actually want to tolerate, and re-raise the rest?

veomni/trainer/text_dpo_trainer.py
```diff
@@ -126,111 +124,154 @@ class DPOConfig:
...
+
+        self.base._build_dataset()
+        self.base._build_collate_fn()
+        self.base._build_dataloader()
+        self._build_postforward()
+        self.policy_model._build_lr_scheduler(args.train_steps * args.train.num_train_epochs)
+        self.base._build_training_context()
+        self.base._init_callbacks(self)
+
+        self.reference_model = self._build_reference_model_runtime()
```
FoolPlayer, line 224: `self.base.model` is never assigned here, but `train_step` calls `self.base._reset_async_activation_offload_if_enabled()` (L445), which reads `self.model` on the `BaseTrainer` — still the class default `None` — and passes it to `reset_async_activation_offload`, which does `model.modules()`. So DPO raises `AttributeError: 'NoneType' object has no attribute 'modules'` on the first step whenever `enable_async_activation` is set.

`self.base.model = self.policy_model` fixes it (the `model` property still shadows it for callback reads) and would let L212 reuse `base._build_lr_scheduler()` instead of duplicating the step arithmetic.

veomni/trainer/text_dpo_trainer.py
```diff
@@ -126,111 +124,154 @@ class DPOConfig:
...
+        model_name: str = "reference",
+        *,
+        train,
+        torch_dtype: str = "bfloat16",
+    ):
+        args = copy.deepcopy(args)
+        args.lora_config = {}
+        args.accelerator.fsdp_config.mixed_precision.enable = False
+        args.accelerator.gradient_checkpointing.enable = False
+        args.accelerator.torch_compile.enable = False
```
FoolPlayer, line 160: `offload_config` isn't disabled alongside the other frozen-eval knobs, so the reference now gets async activation offload applied in `_build_parallelized_model` and allocates a second `PinnedBufferPool` — `async_offload.py` notes the per-call host-cache limits add up, so the configured budget silently doubles. The old hand-written call also didn't pass `ep_sharded_stream_load` / `muon_expert_zero_comm` / `fqn_to_index_mapping`, which it now inherits.

Separately, `build_checkpoint()` is never called for the reference, so `checkpoint` stays `None` and `model_assets` resolves to the shared class-level list at `model_runtime.py:123` — which contradicts the base class's own docstring invariant. Harmless today, but the next checkpoint fan-out will reach it.

veomni/trainer/vlm_trainer.py
```diff
@@ -227,36 +179,80 @@ def _freeze_model_module(self):
...
+        self.base.device = self.base._setup(args)  # registers ParallelState("base") before seed
+        self.base.model = self._build_model_runtime()
+
+        # rewrite build_data_transform to support multimodal transform
+        self._build_data_transform()
+
+        self.base._build_dataset()
+
+        # rewrite build_collate_fn to support multimodal collate_fn
+        self._build_collate_fn()
```
FoolPlayer: Order change worth a second look: on `main`, `_build_data_transform` / `_build_collate_fn` ran *before* `_build_parallelized_model`; now the runtime constructor does the whole build including the wrap, so these read through it.

FSDP2 is fine (`fully_shard` is in-place), but under `fsdp_mode: ddp` at >1 rank the module is a `DDP`, which doesn't forward unknown attribute names. `self.base.model.get_position_id_func()` (L256) then raises, and the `getattr(model, "get_extra_collate_infos", None)` / `getattr(model, "get_metadata_collate_func", None)` probes (L270-273) silently return `None`, dropping the omni audio collate rules and the multimodal-metadata precompute hook.

Suggest reading these capability hooks off the unwrapped module, or capturing them during the build while it's still raw.

veomni/trainer/base.py
```diff
@@ -310,158 +293,113 @@ def __init__(self, args: VeOmniArguments):
...
+        # Builds the model: meta-init, freeze, parallelize and optimizer all run
+        # under the mesh the runtime registered, so nothing below needs a scope.
+        self.model = self._build_model_runtime()
+        # build dataset and dataloader
+        self._build_data_transform()
+        self._build_dataset()
+        self._build_collate_fn()
+        self._build_dataloader()
```
FoolPlayer: The data build used to sit inside `with use_parallel_state("base")`, and it does read ambient state — `MainCollator` / `SequenceParallelCollator` capture `sp_enabled` / `sp_size` / `sp_rank` at construction.

It's correct today only by coincidence: `_setup` happens to install `"base"` as the ambient global on a fresh process, and every in-tree runtime is built from the same accelerator config, so the topology cache hands back one object for `"base"` / `"policy"` / `"reference"`. Note `_init_parallel_state` only installs the ambient global when it is `None`. The moment a model carries its own mesh — the stated premise of this refactor — the collator is built over the job mesh while the model runs over its own, which is the silent class constraint §7 exists to prevent.

Wrapping this block in `use_parallel_state(self.model.parallel_state)` is a no-op today and removes the coincidence. FWIW `base_rl_trainer.py:45` already does exactly this for its collators, while `tasks/train_text_rl.py` and `tasks/train_vlm_rl.py` call the same method unscoped.

tests/models/test_model_runtime.py
```diff
@@ -0,0 +1,404 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
FoolPlayer, line 1: `tests/models/` is enumerated file by file, and neither this file nor `test_runtime_chat_template.py` appears in `gpu_unit_tests.yml` or `npu_unit_tests.yml` — so the tests written specifically to guard this new seam never run. Worth adding both.

Same gap explains the channel-loss break above: `tests/trainer/test_channel_loss_callback.py` isn't CI-wired either, and it calls `computer.install(policy_model)` on a real `nn.Module`, never on a `VeOmniModelRuntime`.

Minor, on the module docstring: at one rank `fsdp_enabled` is `False`, so `build_parallelize_model` applies neither wrap and never runs the weight load — "DDP exercises the same runtime sequence at one rank" overstates what's covered.

veomni/models/model_runtime.py
Coach257: Use `self.model.unwrapped_module`. `channel_loss_callback` will be deleted in the future, using the new `op_registry` : 
```
modeling.forward(): loss = veomni_ops("channel_loss", xxx)
```

veomni/trainer/vlm_trainer.py
Coach257: Fixed

veomni/trainer/base.py
Coach257: Fixed

tests/models/test_model_runtime.py
Coach257: Fixed

veomni/trainer/text_dpo_trainer.py
Coach257: Fixed

veomni/models/model_runtime.py
Coach257: Now using `OSError`.

veomni/trainer/text_dpo_trainer.py
Coach257: Now parse `self.policy_model` to `reset_async_activation_offload`.

tasks/train_text_rl.py
```diff
@@ -13,29 +12,20 @@ def __init__(self, args: VeOmniArguments) -> None:
...
+        self._build_data_transform()
+
+        self.base._build_dataset()
+        self.base._build_collate_fn()
+        self.base._build_dataloader()
+        self.base._build_lr_scheduler()
+        self.base._build_training_context(self.base.model)
+        self.base._init_callbacks()
+
+        self.base._build_preforward_postforward()
```
FoolPlayer: The two RL entrypoints are the only paths that didn't get the `use_parallel_state` wrap — `base.py`, `text_trainer.py`, `vlm_trainer.py`, `dit_trainer.py` and `text_dpo_trainer.py` all have it now. Same for `train_vlm_rl.py`.

L28 is the sharper one: `BaseRLTrainer.__init__:45` deliberately wraps `_build_preforward_postforward()` because its collators read the current ParallelState, and this script calls the same method unscoped.

FoolPlayer (commented):
Re-reviewed at `236a30d8`. All seven points from the last round are addressed, and the channel-loss fix is solid — unwrapping at both layers plus the `is not wrapper` guard and a test asserting `"loss_function" not in vars(runtime)` pins exactly the failure mode. I also checked that narrowing the preprocessor catch to `OSError` doesn't break the toy-config path (`tests/toy_config/qwen3_toy` has only `config.json`, `cached_file` raises `OSError`, dir exists → warn), and re-walked every `self.model` read in `BaseTrainer` against the DPO call sites — nothing reachable is left on the `None` default.

Nothing blocking left. A few follow-ups inline, mostly consistency with the fixes you just made.

One that doesn't fit inline (the lines aren't in the diff): `tests/lora/test_moe_lora_trainer.py:276` and `tests/lora/test_lora_trainer_saveload.py:205` still define `_build_model_assets` on a `BaseTrainer` subclass. Nothing calls that hook on a trainer anymore, and it references a `self.model_config` that `BaseTrainer` no longer has — dead code the sweep removed from five sibling files. Both are CI-enumerated.

veomni/trainer/base.py
```diff
@@ -827,34 +680,36 @@ def forward_backward_step(
...
+                model.set_requires_all_reduce(False)
             elif micro_step == num_micro_steps - 1:
-                self.model.set_requires_all_reduce(True)
+                model.set_requires_all_reduce(True)
 
-    def _reset_async_activation_offload_if_enabled(self):
+    def _reset_async_activation_offload_if_enabled(self, model) -> None:
         if _resolve_offload_config(self.args).enable_async_activation:
-            reset_async_activation_offload(self.model)
+            reset_async_activation_offload(model)
```
FoolPlayer, line 735: Now that this takes `model`, it should read the offload config off it too. `_build_training_context` (L467) moved to `model.args.accelerator.offload_config`, but this still gates on `_resolve_offload_config(self.args)` — the job's. Same object today, but per-model config is the whole premise, so leaving one on job args re-creates the split the refactor is removing.

veomni/trainer/base.py
```diff
@@ -515,160 +456,61 @@ def _build_dataloader(self):
...
+        offload_config = getattr(accelerator, "offload_config", None) or OffloadConfig()
 
-        # Async activation offload uses per-module saved_tensors_hooks (applied
-        # before FSDP sharding), so the global fwd/bwd contexts are nullcontext.
         if offload_config.enable_async_activation:
             from contextlib import nullcontext
 
             self.model_fwd_context, self.model_bwd_context = nullcontext(), nullcontext()
             return
+        enable_gc = bool(getattr(getattr(accelerator, "gradient_checkpointing", None), "enable", False))
```
FoolPlayer: `getattr(...) or OffloadConfig()` and `getattr(..., "enable", False)` mean a `model` without `.args` silently gets activation offload **disabled** rather than raising. For a per-model config that's the wrong failure direction — a refactor that drops `.args` would quietly change training behaviour. Could the tolerance be scoped to test stubs, or just let it raise?

veomni/trainer/text_dpo_trainer.py
```diff
@@ -126,111 +124,161 @@ class DPOConfig:
...
-            self._build_postforward()
-            self.base._build_parallelized_model()
-            self.base._build_optimizer()
-            self.base._build_lr_scheduler()
-            self.base._build_training_context()
-            self.base._init_callbacks()
+        self._build_postforward()
+        self.policy_model._build_lr_scheduler(args.train_steps * args.train.num_train_epochs)
+        self.base._build_training_context(self.policy_model)
+        self.base._init_callbacks(self)
```
FoolPlayer, line 222: I traced every `self.model` read in `BaseTrainer` against the DPO call sites — all covered, either overridden here or passed explicitly, so no live bug.

But `self.base.model` stays at the class default `None`, so correctness now rests on ~10 call sites keeping the discipline. The next base method that reads `self.model` breaks DPO only. Either bind `self.base.model = self.policy_model` (the `model` property still shadows it for callbacks), or make `BaseTrainer.model` raise when unset instead of defaulting to `None`.

veomni/models/model_runtime.py
```diff
@@ -0,0 +1,582 @@
...
+    optimizer: Optional["Optimizer"] = None
+    lr_scheduler: Optional["LRScheduler"] = None
+    # Class-level so __getattr__ never fires for them before the build runs.
+    # ``checkpoint`` is only ``None`` during that build: __init__ installs one
+    # unconditionally, so everything downstream may reach through it.
+    checkpoint: Optional["ModelCheckpointManager"] = None
+    tokenizer: Optional[Any] = None
+    processor: Optional[Any] = None
+    chat_template: Optional["ChatTemplate"] = None
+    model_assets: List[Any] = []
```
FoolPlayer: Two small ones here.

L114: `train` shadows `nn.Module.train()`, so `trainer.model.train()` raises `TypeError: 'TrainingArguments' object is not callable` — for the most idiomatic module call there is, and the class docstring promises module APIs forward. No caller hits it today. `train_args` / `job`, or an explicit `def train(self, mode=True)`, would close it.

L124: mutable class default. Every override happens to rebind before appending, so it's safe by convention only — and `DPOReferenceModelRuntime` never calls `_build_model_assets`, so `reference.model_assets` *is* this shared list.

veomni/models/model_runtime.py
```diff
@@ -0,0 +1,582 @@
...
+        """Build this model's device mesh and register it under :attr:`model_name`.
+
+        The process group itself is job-bound and must already be initialised by
+        :meth:`BaseTrainer._setup`; this only derives the model's own
+        mesh from its accelerator config, which is why sibling models in one job
+        can hold different ones.
+        """
+        from ..distributed.parallel_state import init_parallel_state_from_config
+
+        init_parallel_state_from_config(self.args.accelerator, self.model_name)
```
FoolPlayer, line 181: For the single-model trainers this re-registers a name `BaseTrainer._setup` already registered from the same `accelerator`, so `_init_parallel_state` logs `Parallel state 'base' is already registered` on every normal run. A warning on the happy path trains people to ignore warnings — worth demoting to debug for the same-config case, or having `_setup` not register when a runtime will.

veomni/trainer/callbacks/channel_loss_callback.py
```diff
@@ -111,9 +111,15 @@ def _resolve_loss_fn_host(model: torch.nn.Module) -> torch.nn.Module | None:
         """Walk common wrappers to find the module whose forward calls ``loss_function``."""
 
         visited = set()
-        cur: torch.nn.Module | None = model
+        cur: Any = model
         while cur is not None and id(cur) not in visited:
             visited.add(id(cur))
+            if not isinstance(cur, torch.nn.Module):
+                nxt = getattr(cur, "unwrapped_module", None) or getattr(cur, "model", None)
```
FoolPlayer: `or` on an `nn.Module` is a footgun: modules that define `__len__` (an empty `nn.Sequential`/`nn.ModuleList`) are falsy, so a valid `unwrapped_module` could be skipped in favour of `.model`. `nxt = getattr(cur, "unwrapped_module", None)` then `if nxt is None: nxt = getattr(cur, "model", None)` is equivalent and can't misfire.

veomni/trainer/callbacks/trace_callback.py
```diff
@@ -215,7 +215,7 @@ def __init__(self, trainer: "BaseTrainer") -> None:
         self.lora_config = trainer.model.get_lora_config() if hasattr(trainer.model, "get_lora_config") else None
```
FoolPlayer: Not introduced by this PR — under DDP `trainer.model` wrapped a `DDP` before too — but it's the same pattern `unwrapped_module` was just added for, and one site the fix missed. `get_lora_config` lives on `VeOmniLoraModel`, so under DDP + LoRA this `hasattr` is `False` and `EnvironMeter` silently loses the LoRA config (affects reported MFU/FLOPs, not training). `self.trainer.model.unwrapped_module` would fix it in passing.

veomni/models/model_runtime.py
Coach257: Trainer register `base` parallel state as data related use `dp_size` and `sp_size`. ModelRuntime alse register `base` ps as there is only one module currently. In the coming pr, different ps with differen `model_name` will be registered in ModelRuntime.

veomni/trainer/base.py
Coach257: Fixed

veomni/trainer/text_dpo_trainer.py
Coach257: policy model not bind to base.model

veomni/trainer/callbacks/trace_callback.py
```diff
@@ -212,10 +212,12 @@ def __init__(self, trainer: "BaseTrainer") -> None:
         super().__init__(trainer)
 
         args: "VeOmniArguments" = self.trainer.args
-        self.lora_config = trainer.model.get_lora_config() if hasattr(trainer.model, "get_lora_config") else None
+        # LoRA config lives on VeOmniLoraModel, which DDP does not forward.
+        module = trainer.model.unwrapped_module
+        self.lora_config = module.get_lora_config() if hasattr(module, "get_lora_config") else None
```
FoolPlayer: This breaks DiT `offline_embedding`, at trainer construction.

`DiTModelRuntime._build_model` leaves `self.model = None` for that task, and `DiTTrainer.__init__` calls `self.base._init_callbacks()` **unconditionally** (only `_build_lr_scheduler` / `_build_training_context` sit behind the `if training_task != "offline_embedding"`). `_init_callbacks` constructs `EnvironMeterCallback` first, and `unwrapped_module` starts with `if model is None: raise AttributeError(...)`.

It was safe until this commit precisely because the old spelling was `hasattr(trainer.model, "get_lora_config")` — `hasattr` swallowed that `AttributeError` and left `lora_config = None`. Moving the unwrap outside the guard removed the thing that absorbed it. (Also safe on `main`, where `trainer.model` was a bare `None`.)

Nothing covers this: `offline_embedding` has zero hits in `tests/` and `.github/workflows/`, so it would ship broken.

```python
module = getattr(trainer.model, "unwrapped_module", None)
self.lora_config = module.get_lora_config() if hasattr(module, "get_lora_config") else None
```

FWIW this is the only unguarded `unwrapped_module` access — the three in `channel_loss_callback.py` all use `getattr(..., default)`, and the two in `vlm_trainer.py` are on a path where the model always exists.

veomni/trainer/base.py
```diff
@@ -297,6 +288,28 @@ class BaseTrainer(Stateful, ABC):
...
+        except AttributeError:
+            raise RuntimeError(
+                f"{type(self).__name__}.model is unset. Single-model trainers assign it "
+                "during build; a multi-model trainer must pass the runtime into "
+                "base methods explicitly."
+            ) from None
+
+    @model.setter
+    def model(self, value: VeOmniModelRuntime) -> None:
+        self._model = value
```
FoolPlayer, line 304: Good change, but raising `RuntimeError` breaks the `getattr`/`hasattr` contract: both only swallow `AttributeError`. `channel_loss_callback.py:1163` already relies on it — `model = getattr(self.trainer, "model", None)` will now propagate instead of falling through to the `cpu` default.

Not reachable in production today (that helper returns from the `device` branch above it, and every trainer sets `self.base.device`), so this is about the idiom rather than a live break. Raising `AttributeError` with the same message keeps the failure just as loud while leaving `getattr(..., default)` working.

veomni/trainer/base.py
```diff
@@ -215,6 +200,17 @@ def _resolve_offload_config(args) -> OffloadConfig:
...
 
 
+def _resolve_model_offload_config(model) -> OffloadConfig:
+    """Offload knobs live on this model's accelerator, not the job args.
+
+    ``model.args.accelerator`` must exist — a missing handle is a build bug,
+    not a reason to silently disable activation offload.
+    """
+    accelerator = model.args.accelerator
+    config = getattr(accelerator, "offload_config", None)
```
FoolPlayer, line 203: This is the right shape — asserting `model.args.accelerator` rather than defaulting is exactly the failure direction a per-model config wants.

With both call sites moved over, though, the older `_resolve_offload_config` at L196 has no production caller left; the only remaining references are `tests/trainer/test_step_sync.py:8` and `:67`, i.e. a helper kept alive by its own test. Worth dropping both, or noting why it stays.

FoolPlayer (commented):
Re-reviewed at `5e2f1f26`. All eight follow-ups from the last round are addressed, several more thoroughly than I suggested — `_resolve_model_offload_config` giving both call sites one source of truth, the `model` property replacing the `None` default, and the `train` → `train_args` rename (which I checked is complete across `veomni/`, `tasks/` and `tests/`; the remaining `.train` hits are all `VeOmniArguments.train`). The two per-instance/forwarding tests are the right assertions.

One of the fixes introduced a regression, though — the `unwrapped_module` change in `trace_callback.py` breaks DiT `offline_embedding` at trainer construction. Details inline, plus two minor notes.

FoolPlayer (approved):

PR #1134, opened 2026-08-29
[model, ops, config, docs] feat: add the DeepSeek-V4 Lightning Indexer KL objective

FoolPlayer:
### What does this PR do?

Trains the DeepSeek-V4 Lightning Indexer with the DeepSeek-V3.2 eq. (4) sparse KL objective, behind `dsa_indexer_loss` (off by default).

The indexer decides which compressed KV entries CSA attention may attend to, and until now nothing trained it: `DeepseekV4Indexer.forward` returned integer top-k indices, so no gradient reached its 10.75M parameters per CSA layer — 226M across DeepSeek-V4-Flash's 21 CSA layers, frozen while everything beneath them moved. This wires up the objective end to end: the indexer's per-slot scores become the student, the teacher is recomputed in the forward from the TileLang sparse attention's own log-sum-exp, and the per-layer KL is summed, normalised per query token, scaled by `dsa_indexer_loss_coef` and folded into the total loss.

Builds on two merged PRs:
- #1109 — the TileLang teacher-distribution kernel (`sparse_mqa_target_fwd`) this consumes.
- #1110 — `aux_metrics`, which the four indexer metrics are reported through.

Independent of #1131, the DeepSeek-V4 context-parallel model support, and mergeable in either order. The objective refuses both sequence-parallel modes: Ulysses because a rank holds a head shard of a sum the teacher needs whole, and CP because DeepSeek-V4's forward has no context-parallel path on `main` — #1131 is what adds one — so a rank would otherwise score its sequence shard as though it were a whole sequence and the teacher would be built from that attention. The token-weighted reduction CP needs is written and sits behind `sp_enabled`, unreachable under that refusal. Whichever of the two lands second owes the other two things, both recorded in the design doc: flipping the `cp_size` gate, and backward coverage for the indexer's own collectives, which this objective makes reachable from backward for the first time.

### Checklist Before Starting

- Search for relative PRs/issues and link here: #1109 and #1110 (merged, this builds on both), #1131 (CP model support, independent), #1108 (merged; it made head-splitting the indexer something a config has to name explicitly, which is why this PR's Muon note reads the way it does), #1111 (CP infrastructure, merged). Upstream context: [NVIDIA/Megatron-LM#5776](https://github.com/NVIDIA/Megatron-LM/issues/5776) — the teacher-distribution defect this implementation avoids.

### Test

`tests/ops/test_deepseek_v4_indexer_loss.py` — 100 cases, all passing on an SM90 GPU in ~22s, with no skips. Picked up by the existing `pytest -s -x tests/ops` step in `gpu_unit_tests.yml`; the kernel-dependent cases skip cleanly off SM90. The model-level cases build the reference checkpoint's 4-layer schedule from a programmatic config, so nothing depends on a mounted checkpoint.

What the suite pins, beyond the usual:

- **The teacher is the paper's, not Megatron's.** An explicit fp32 reference computes `Z_h` from sink, window and compressed logits, then the head sum, then one L1 normalisation. Perturbing only one head's sink, and separately only the window logits, must move the target — the invariant Megatron's compressed-only teacher violates. Multi-head with differing per-head preferences, since at one head the denominator cancels and the test cannot tell the two apart.
- **The gradient reaches the indexer and stops there.** Asserted on the four projections' gradients under both `use_reentrant` settings, with checkpointing asserted engaged rather than trusted; and no gradient to the main query, KV, sink or `hidden_states`.
- **Every refusal fails when deleted**, and the shared arity gate is held to by all three call sites that read it. The two prerequisites that are fields on the same dataclass as the flag are refused at parse time, before a rank reads 54.8 GB of checkpoint to be told that three lines of YAML disagree.
- **The metrics' aggregations**: the loss keeps the layer sum while the metric is a per-layer mean, and the captured fraction is a ratio of means rather than a mean of ratios. That test carries a guard that fails loudly if the fixture ever stops being able to separate the two, rather than passing vacuously.
- **The rows with nothing to teach** leave both the KL and its reference, so a query whose teacher mass underflowed cannot flatter the captured fraction by contributing zero to one term and a full `log(n_candidates)` to the other.

Beyond CI, a 43-layer DeepSeek-V4-Flash SFT run at `dsa_indexer_loss_coef: 1.0` against three flag-off baselines on bitwise-identical batches, over 375 steps. MFU sits inside the baselines' own ±2.9% spread, i.e. the objective is free on step time. LM loss runs 0.5–1.9% above the baselines, which is not noise — the baselines agree with each other to 0.02% — and the two channels that produce it are documented in `docs/usage/arguments.md`, along with why neither is a gradient leak.

### API and Usage Example

Two new fields on DeepSeek-V4's own config, beside the `output_router_logits` / `router_aux_loss_coef` pair the model's other auxiliary objective is configured through — a training objective is a property of the model, while `ops_implementation` selects kernel backends:

```yaml
model:
  model_config:
    dsa_indexer_loss: true
    dsa_indexer_loss_coef: 1.0             # default
  ops_implementation:
    dsa_indexer_implementation: tilelang   # required by the objective
    dsa_attention_implementation: tilelang # required by the objective
```

Both fields are serialised into the checkpoint's `config.json`, so a flag-on checkpoint asks for the objective again every time it is reloaded. `tasks/infer/*` pins it off, since inference trains nothing and has the DSA stack on eager.

`configs/text/deepseek_v4_indexer_loss.yaml` is a runnable smoke config on the 4-layer checkpoint, and derives `max_seq_len` from `index_topk` and the compression rate rather than inheriting the sibling config's 2048 — which sits exactly on the boundary where the top-k selects everything and the objective silently degenerates to the dense eq. (3).

Four metrics are reported: `training/indexer_kl` (per-layer mean), `training/indexer_kl_uniform` (the zero-information reference it has to be read against), `training/indexer_kl_captured` (`1 − kl / uniform`, formed per micro-batch; the one to watch), and `training/lm_loss_before_indexer_kl`.

Unsupported configurations are refused rather than silently ignored: a non-TileLang indexer or attention (at launch), `ulysses_size > 1` or `cp_size > 1`, a negative or non-finite coefficient, a wrongly typed one (`model_config` is untyped from YAML down, so `dsa_indexer_loss: "false"` is truthy and would otherwise switch the objective *on*), NPU, and a model with no CSA layer to train. No model-type allow-list is needed: the flag is a field of `DeepseekV4Config`, so a config that does not declare it cannot ask for the objective.

### Design & Code Changes

`docs/design/deepseek_v4_indexer_loss.md` records the constraints and why they exist, the decisions that took more than one attempt, and the failure modes that produce a plausible loss curve instead of an error — which is the shape of this whole feature.

- `veomni/models/transformers/deepseek_v4/deepseek_v4_gpu_patch_gen_config.py`: the objective. `DeepseekV4Indexer.forward` detaches its input and returns `index_score`; the CSA compressor routes it out through `CompressedCandidates.indexer_scores`; the attention forward builds the KL against the teacher; the decoder layer and model loop thread it up; `ForCausalLM` reduces, folds in and reports. Three helpers (`_indexer_loss_enabled`, `_builds_indexer_kl`, `_split_indexer_output`) are the single source of the return arity that five call sites have to agree on.
- `veomni/models/transformers/deepseek_v4/deepseek_v4_npu_patch_gen_config.py`: carries the same names into the NPU sibling module so it stays importable and type-checks. The CSA compressor takes `build_indexer_loss` because the call site is shared and refuses it, rather than dropping it — `dsa_indexer_implementation` is a plain `Literal` with no hardware gate, so `tilelang` parses on NPU and the run would otherwise fail several frames later on an internal wiring check.
- `veomni/models/transformers/deepseek_v4/configuration_deepseek_v4.py`: a `DeepseekV4Config` subclass declaring the two fields, registered in `MODEL_CONFIG_REGISTRY`. Hand-written rather than generated because `patchgen` cannot patch a config class: `modify_init` is declared in `patch_spec.py` and unimplemented in the generator, so it silently produces nothing. Declaring them is load-bearing, not cosmetic — `from_pretrained` applies a kwarg only when the attribute already exists, so the primary use case (switching the objective on from YAML against a base checkpoint that never carried the field) drops silently without it.
- `veomni/models/auto.py`: `check_indexer_loss_prerequisites`, called from `build_foundation_model`. It is the only place holding the model config and `OpsImplementationConfig` at once, which the TileLang prerequisites need — the flag is on one dataclass and the two implementations on the other. It runs on the finished config, after `from_dict` has applied the `model_config` overrides, which a `__post_init__` on either dataclass would not: `from_dict` constructs from the on-disk JSON and only then `setattr`s.
- `veomni/utils/model_outputs.py`: `MoeModelOutputWithIndexerKL`, and `aux_metrics` on the causal-LM output.
- `veomni/trainer/text_dpo_trainer.py`: `_build_reference_model` now passes `config_kwargs`, as the other three trainers already did. It was the one build path reading only `config.json`, which did not matter while the flag was global and does now — the policy would honour a `model_config` override and its reference would not.
- `veomni/models/transformers/deepseek_v4/packed_utils.py`: `indexer_scores` on `CompressedCandidates`.

- [x] Applied pre-commit checks (`make quality`, `make check-patchgen`)
- [x] Added/updated documentation (`docs/usage/arguments.md`, `docs/design/deepseek_v4_indexer_loss.md`, `docs/index.md`)
- [x] Added tests to CI workflow — `tests/ops/test_deepseek_v4_indexer_loss.py`, covered by the existing `tests/ops` step

## Summary by CodeRabbit

* **New Features**
  * Added optional DeepSeek-V4 indexer KL auxiliary loss support for training.
  * Added configurable loss enablement and coefficient settings.
  * Added reporting for indexer KL and uniform-reference metrics.
  * Added support for matching model settings between policy and reference models in DPO training.

* **Bug Fixes**
  * Inference now disables the training-only indexer loss automatically.
  * Added validation and clearer errors for incompatible configurations.

* **Documentation**
  * Added design documentation and expanded usage guidance for the indexer loss.


veomni/arguments/arguments_types.py
```diff
@@ -1199,6 +1199,26 @@ class OpsImplementationConfig:
         default="eager",
         metadata={"help": "DeepSeek sparse attention implementation: 'eager', 'flashmla_cudnn', or 'tilelang'."},
     )
+    dsa_indexer_loss: bool = field(
```
Coach257: The indexer KL is a training objective, not a kernel backend. `dsa_indexer_loss` / `dsa_indexer_loss_coef` should not live on `OpsImplementationConfig`.

That dataclass is kernel selection (`*_implementation: eager | tilelang | …`). These two fields are the same shape as MoE’s `output_router_logits` / `router_aux_loss_coef`, which already sit on HF model config and are folded in from `self.config` / `self.router_aux_loss_coef` a few lines above the indexer fold-in. Put them there too, e.g. via `model.model_config`:

```yaml
model:
  model_config:
    dsa_indexer_loss: true
    dsa_indexer_loss_coef: 1.0
  ops_implementation:
    dsa_indexer_implementation: tilelang
    dsa_attention_implementation: tilelang
```

Read them in the patched forward as `self.config.dsa_indexer_loss` (and the coef), same as the router aux path.

That also drops the machinery this placement forced:

- **`check_indexer_loss_supported` / `INDEXER_LOSS_MODEL_TYPES` in `auto.py`.** That allow-list exists only because the flag is on a model-agnostic ops config, so GLM MoE DSA can set it and silently train nothing. A field on `DeepseekV4Config` cannot be set on GLM; the check is unnecessary.
- **`OpsConfigSlot` growing a `default=` and boolean/float values.** Leave it as a string kernel selector. Module-level slots are also shared across every instance of the modeling module (DPO policy/reference); `self.config` is per-model.
- **`arguments_types.py`.** No new fields, no parse-time TileLang co-validation on `OpsImplementationConfig`.

The TileLang requirement belongs in modeling: `_indexer_loss_enabled` already refuses a non-TileLang indexer/attention on the first forward. Keep that (and the Ulysses/CP refusals) keyed off `self.config.dsa_indexer_loss`. Zero coef stays “off” at that same predicate, so Muon does not decay a zero-grad indexer.

Declare the two fields on `DeepseekV4Config` via patchgen (default `False` / `1.0`). Do not rely on HF accepting unknown kwargs.

FoolPlayer:
Done — both fields now live on DeepSeek-V4's own config. Two places where I deviated from the suggestion, and one thing I found while implementing that changed the shape of the fix.

**Not a plain `__post_init__` validator.** `from_dict` constructs the config from the on-disk JSON and *then* `setattr`s the `model_config` overrides, so a bound checked in `__post_init__` reads `config.json` and never the YAML line contradicting it. Validation therefore lives in `check_indexer_loss_prerequisites`, called from `build_foundation_model` on the finished config. That is also the only place that holds the model config and `OpsImplementationConfig` at once, which the TileLang prerequisite check needs — the flag is on one dataclass and the two implementations are on the other, so neither can see the disagreement alone. It still fires before any rank reads a weight (54.8 GB for V4-Flash), which was the point of checking early.

**A hand-written subclass rather than patchgen.** `patchgen` cannot patch a config class: `modify_init` is declared in `patch_spec.py` and unimplemented in the generator, so it silently produces nothing. `veomni/models/transformers/deepseek_v4/configuration_deepseek_v4.py` subclasses upstream and is registered in `MODEL_CONFIG_REGISTRY`, following the `qwen3_omni_moe` precedent.

**Why the declaration is load-bearing.** I expected declaring the fields to be cosmetic and it is not — `from_pretrained` semantics are asymmetric in a way worth recording:

| | undeclared field | declared field |
|---|---|---|
| key in `config.json` | applied (`from_dict` `setattr`s disk keys) | applied |
| passed as a kwarg, key absent from disk | **silently dropped** | applied |

The dropped cell is the primary use case: switching the objective on from YAML against a base checkpoint that never carried the field. Without the subclass a user's `dsa_indexer_loss: true` vanishes between the YAML and the model and the run trains the language-model objective alone, reporting nothing amiss. `test_undeclared_is_not_the_same_as_absent` pins all four cells.

Three consequences of the move that needed handling:

- **Serialisation.** Both fields are written into `config.json`, so a flag-on checkpoint asks for the objective every time it is reloaded. `tasks/infer/infer_text.py` pins every op to eager, so it would have been refused at build with no override available from `InferArguments`. It now passes `dsa_indexer_loss: False` — dropped harmlessly for every other model, per the table above.
- **DPO.** `_build_reference_model` was the one `build_foundation_model` call site not passing `config_kwargs`, which did not matter while the flags were global. It does now: the policy would honour a `model_config` override and its reference would not. Fixed to match the other three trainers.
- **NPU.** `dsa_indexer_implementation` is a plain `Literal` with no hardware gate, so `tilelang` parses on NPU and the gate admits the objective. The NPU compressor took `build_indexer_loss` and dropped it, leaving the run to fail a few frames later on the attention forward's internal wiring check. It now refuses with a message about the actual cause.

Also: `model_config` is an untyped `Dict` the whole way from YAML, so the check validates types. `dsa_indexer_loss: "false"` is truthy and would otherwise switch the objective **on** for someone who wrote it to switch the objective off.

One thing I chose not to do. A config left in the old `ops_implementation` shape parses cleanly and silently trains nothing, because the YAML parser drops keys that are not fields of the dataclass they land in. I documented it in `arguments.md` rather than adding deprecated fields that raise — the flags never landed on `main`, so the only configs in the old shape are ones written against this branch. Happy to add the rejecting shim if you would rather have it refuse.

`make quality` and `patchgen --check` clean; 112 indexer-loss tests and all 289 DeepSeek/patchgen tests pass. Marked `[BREAKING]` for the config move.

FoolPlayer:
Moved. `veomni/models/auto.py` no longer knows the objective exists.

`check_indexer_loss_prerequisites` is now `DeepseekV4Config.validate_build_prerequisites`, and what the builder keeps is a hook:

```python
def check_model_build_prerequisites(config):
    validate = getattr(config, "validate_build_prerequisites", None)
    if callable(validate):
        validate()
```

Two of the three things that check reads are `DeepseekV4Config` fields and the third is which kernels those fields require, so the knowledge was entirely the model's. A model with nothing to refuse still declares nothing — absent is a no-op rather than a required empty method, which `test_a_config_that_cannot_ask_for_the_objective_is_left_alone` pins from one side and `test_the_generic_hook_reaches_the_model_that_implements_it` from the other.

The hook takes no arguments and the method reads the ops singleton itself. The alternative was a signature carrying `OpsImplementationConfig` — and then the next hook needs the parallel state, and the one after that something else, until the builder's hook signature enumerates everything any model might want. Reading the singleton inside keeps the contract at "a config may refuse the run it is about to be built into", which is also why `_indexer_loss_enabled` reads the parallel state rather than receiving it.

What a second model gaining the objective now costs: declare the two fields, implement the method. No edit here.

**The other check in this file.** `check_context_parallel_supported` is the remaining gate of this shape, and it does not move the same way — its allow-list has to refuse models *absent* from it, so a per-model hook would invert the safety property and grant CP to any model implementing nothing. Turning it into a capability the model declares (absent still meaning refused) is the equivalent change, and since it predates this PR I did it separately: **#1153**. That one also can't use a config attribute, because `deepseek_v4` has no VeOmni-owned config class on `main` — this PR is what adds it.

`make quality` and `patchgen --check` clean; 113 indexer-loss tests and 290 DeepSeek/patchgen tests pass. Rationale is in `docs/design/deepseek_v4_indexer_loss.md` under "Decisions that took more than one attempt".

Coach257 (approved):

Issue #1073, opened 2026-08-18
[BREAKING][ops, omni, model] refactor: make op and kernel selection per-module


Coach257:
Part of #1057. **Depends on:** #1064 (needs `ModuleRuntime` to own the registry) and all of Wave 3 (Bagel is the motivating case).

## Why

An Omni graph runs several models in one process, and they do not want the same kernels. Bagel is the concrete case: `bagel_siglip_navit` wants flash attention over its NaViT patch stream, `bagel_qwen2_mot` wants flex attention for training and MagiAttention/flash for inference denoise, `bagel_vae` wants plain SDPA. Today VeOmni cannot express that in the dispatch layer, only in a per-model config field, so modules that need anything else reach around the layer entirely.

**The config surface already promises per-module ops; the implementation does not deliver it.** `OmniModuleRuntimeArguments.ops_implementation` exists (`veomni/omni_arguments/arguments_types.py:100`), `OmniConfig.module_ops_implementation(name)` exists, and shipped configs already use it — `configs/seed_omni/Janus/janus_1.3b/modules_train.yaml` sets `janus_vqvae` to `eager` and `janus_llama` to `flash_attention_2`; `configs/seed_omni/Bagel/bagel_7b_mot/modules_train.yaml` sets `bagel_qwen2_mot` to `flex_attention`. That works **only** for attention, because attention alone is resolved per model instance from `module.config._attn_implementation` at forward time. Everything else in the ops layer is process-global.

**Three layers, one global singleton.**

1. `OpsImplementationConfig` is stored in a module-level singleton (`veomni/ops/config/singleton.py`). `apply_ops_config()` (`veomni/ops/__init__.py:75`) sets it, mutates `transformers.loss.loss_utils.LOSS_MAPPING`, and runs `apply_global_ops()` which `setattr`s module-global function pointers.
2. `build_foundation_model` calls `apply_ops_config()` (`veomni/models/auto.py:172`) — so **every `ModuleRuntime` build overwrites the singleton, and the last module built wins** for cross-entropy, load-balancing loss and fused MoE. `OmniModel._load_modules` has the same pattern (`veomni/models/seed_omni/modeling_omni.py:194`).
[...]

PR #1142, opened 2026-08-31
[BREAKING][ops, model] refactor: replace OpSlot with VeomniOp

Zx55:
### Why

VeOmni previously selected optimized operations through several independent mechanisms, including `OpSlot`, kernel factories, global function replacement, loss mapping mutation, and attention-specific registries.

This made implementation selection difficult to reason about and prevented different model instances from safely selecting different implementations.

This PR consolidates tensor-level operations under `veomni.ops`. Models hold instance-local `VeomniOp` handles. Model-specific argument processing and loss policy remain in `veomni.models`.

### What does this PR do?

The refactor creates one path from implementation registration to instance-local model execution:

```mermaid
flowchart TB
    subgraph Registration["1. Operation registration"]
        direction LR
        Raw["Raw forward/backward pair"]
        Opaque["Opaque eager or vendor wrapper"]
        Register["register_op(...)"]
        Entry["OpEntry<br/>wrapper + optional raw pair"]
        Registry["OP_REGISTRY<br/>(op, variant, implementation, device)"]
        Raw --> Register
        Opaque --> Register
        Register --> Entry
        Entry --> Registry
    end

    subgraph Runtime["2. Model construction and ordinary execution"]
        direction LR
        YAML["model.ops_implementation"]
        Config["OpsImplementationConfig"]
        Model["Model __init__"]
        Handle["VeomniOp(op, variant, impl)"]
        Resolve["Resolve row and<br/>check hardware"]
        Call["entry.wrapper(...)"]
        Kernel["Selected implementation<br/>PyTorch / Liger / Triton / NPU / MLU"]
        YAML --> Config
        Config --> Model
        Model --> Handle
        Handle --> Resolve
        Registry --> Resolve
        Resolve --> Call
        Call --> Kernel
    end

    subgraph Compound["3. Compound operation execution"]
        direction LR
        Outer["Outer raw forward/backward"]
        Inner["resolve_inner_op(...)"]
        RawCall["Call inner raw forward/backward<br/>without nesting an autograd wrapper"]
        State["Flatten and restore SavedState"]
        Outer --> Inner
        Registry --> Inner
        Inner --> RawCall
        RawCall --> State
    end
```

The important properties are:

- **One source of truth:** tensor-level implementations are registered in `OP_REGISTRY`, including an explicit `eager` row.
- **Instance-local selection:** each model constructs its own `VeomniOp`; constructing one model no longer changes another model's implementation.
- **Two registration forms:** an implementation may provide either a raw forward/backward pair or an opaque eager/vendor wrapper.
- **Composable autograd:** compound operations call inner raw pairs directly and preserve their saved state without nesting `Function.apply`.
- **Clear ownership boundary:** tensor math lives in `veomni.ops`; model-specific input adaptation and loss policy live in `veomni.models`.

Process-wide integration is limited to installation work such as registering VeOmni attention names with Transformers. Attention masks, batch-invariant patches, and model-side loss policy are not registered operation rows.

HF-interface families bind attention from `config._attn_implementation`. Families that are not on that path (Wan, MiniMax, Flux, Qwen Image, LTX) bind through `resolve_op_impl("attn_implementation")`.

The complete operator catalog and extension guide are documented in [`veomni/ops/README.md`](veomni/ops/README.md).

The breaking-change map from OpSlot / `kernel_registry` / process-wide bind to `VeomniOp` / `OP_REGISTRY` / instance-local handles is [`docs/design/opslot_to_veomniop.md`](docs/design/opslot_to_veomniop.md). Use that note when updating modeling, YAML (`model.ops_implementation`), or out-of-tree kernels.

### Operator surface differences from `main`

Except for the differences below, the refactored operations are intended to preserve the tensor contracts, numerical behavior, gradients, and hardware support of `main`. Moving an existing implementation into the new registry is not considered a behavior change.

| Area | Difference from `main` | Behavior change |
| --- | --- | --- |
| RMSNorm `unweighted` | No `npu` implementation is registered. NPU consumers use `eager`, because the previous NPU-labelled path was PyTorch math and did not call a distinct `torch_npu` kernel. | No numerical change; removes a hardware alias. |
| RMSNorm `offset` | The `qwen3_5` variant is renamed to `offset`. Rows are `offset/{eager,liger_kernel,npu}`. Consumers are Qwen3.5, Qwen3.5-MoE, Gemma3, and Qwen4-Exp. | Naming only; offset-1 Gemma-style math is unchanged. |
| RMSNorm `group_size` | Every RMSNorm variant accepts optional `group_size`. Fused weighted rows delegate grouped calls to eager. Only Qwen4-Exp passes a non-`None` value (`hc_norm`, `norm_key`, `norm_query`, `norm_conv`). | Contract extension. Qwen4 grouped math matches `main`'s local reshape; other callers omit it. |
| Cross-entropy | Exposes `standard/{eager,chunk_loss,liger_kernel}`. The previous `npu` entry was an alias of `chunk_loss`, so NPU selects the canonical `chunk_loss` or `eager` name. Causal and sequence-classification policy remains model-side. | No kernel-math change; removes an alias and separates model policy. |
| Vision RoPE | There is no separate `rope_vision` family. Rank-3 vision and text full RoPE both use `rope/full`. `rotary_pos_emb_vision_implementation` remains temporarily as an independent configuration choice. | No numerical change; consolidates equivalent tensor operations. |
| Causal Conv1D | Public default is `activation="silu"`, matching the model-path normal operation. The previous default was `None`. | Default-only; an explicit activation still wins. |
| MoE implementation names | Registered names use the same `fused_triton`, `fused_quack`, `fused_npu`, and `fused_mlu` values exposed by configuration. | Naming only; fused implementations are unchanged. |
| Attention adapters | VeOmni names drop `_with_sp`: `veomni_flash_attention_*`, `veomni_flex_attention`, `veomni_magi_attention`. | Naming only. |

A hardware-specific implementation name is only retained when it represents a real hardware-specific implementation. For example, MoE `fused_npu` remains because it uses `torch_npu` grouped GEMM.

These rows are new first-class registry surfaces. They are not silent replacements of an existing `main` op name:

| Area | Difference from `main` | Consumers |
| --- | --- | --- |
| LayerNorm | New `layer_norm/standard/{eager,apex}`. On `main` this lived only inside async Ulysses Apex/native calls. | Nested by async Ulysses QKV when `norm_type="layernorm"`. No model constructs `VeomniOp("layer_norm", ...)` directly. Flux and Qwen2.5-Omni still use `nn.LayerNorm`. |
| Partial RoPE | `rope/partial` adds `liger_kernel` (and an explicit `eager` row). | Qwen3.5, Qwen3.5-MoE, Qwen4-Exp via `rotary_pos_emb_implementation`. |
| RoPE variants | New `rope/interleave/eager` and `rope/mrope/eager`. | Interleave: DeepSeek-V3, GLM-MoE-DSA. MRoPE: Qwen2-VL, Qwen2.5-VL, Qwen2.5-Omni. |
| GeGLU | New `swiglu_mlp/geglu/{eager,liger_kernel}`. | Gemma3. |
| Sage / SDPA adapters | New `veomni_sage_attention` and `veomni_sdpa`. Config rewrites `sage_attention` / `sdpa` to those names. | Sage: Wan (`sageattention` remap) plus HF-path rewrite. SDPA: HF-path rewrite; Qwen Image, LTX, and Flux treat `sdpa` and `veomni_sdpa` as the same set. |

Correctness fixes discovered during migration are intended to restore `main` semantics and are therefore not listed as intentional differences.

### Validation

The ops suite covers registry contracts, public signatures, forward and backward parity, hardware requirements, and optional dependency handling. The models suite covers construction, instance-local selection, family parity, and model-facing operator integration.

Trainer composition, `ModelArguments` path fallback, and SP `ReduceLoss` live with their owners rather than `tests/models`.

```bash
pytest tests/ops tests/models tests/arguments \
  tests/trainer/test_trainer_model_runtime.py \
  tests/distributed/test_reduce_loss.py -q
make quality
```

Latest local GPU result on 8× H20:

```text
1592 passed, 69 skipped
```

Hardware-specific tests are skipped when their required NPU, MLU, Magi, or optional vendor runtime is unavailable.

### Usage

Models construct an operation handle once and call it directly:

```python
from veomni.ops import VeomniOp
from veomni.ops.config import resolve_op_impl


self.veomni_rms_norm = VeomniOp(
    "rms_norm",
    "standard",
    resolve_op_impl("rms_norm_implementation"),
)

hidden_states = self.veomni_rms_norm(
    hidden_states,
    self.weight,
    eps=self.variance_epsilon,
)
```

The exact registered and currently available implementations can be inspected through the registry:

```python
from veomni.ops import OP_REGISTRY


OP_REGISTRY.list_registered("rms_norm", "standard")
OP_REGISTRY.list_available("rms_norm", "standard")
OP_REGISTRY.list_entries("rms_norm", "standard")
```


Coach257:
#1073

tests/kernels/attention_cases.py
```diff
@@ -0,0 +1,215 @@
...
+from torch.nn.attention import SDPBackend, sdpa_kernel
+from torch.nn.attention.flex_attention import BlockMask, create_block_mask
+
+from veomni.kernels.mask import MagiAttentionMask
+
+
+_BAGEL_MODES = ("causal", "noise", "full", "causal")
+
+
+def bagel_span_splits(sequence_length: int) -> list[int]:
```
Coach257: rm `bagel` prefix

configs/multimodal/qwen3_5_moe/qwen3_5_moe_vl.yaml
```diff
@@ -2,7 +2,7 @@ model:
   model_path: Qwen/Qwen3.5-35B-A3B
   ops_implementation:
     attn_implementation: flash_attention_2
-    moe_implementation: fused_triton
+    moe_implementation: triton
```
Coach257, line 5: 1. cuda: eager/fused_triton/fused_quack
3. npu: eager/fused_npu
4. mlu: eager/fused_triton/fused_mlu

tests/kernels/mhc/test_mhc.py
```diff
@@ -0,0 +1,267 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
Coach257, line 1: Mkdirs: `tests/kernels/base/test_kernel_entry`; `tests/kernels/attention/magi|fa`;`tests/kernels/mhc/xxx`

veomni/kernels/_kernels/async_ulysses/o_proj/standard/eager.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2025 Bytedance Ltd. and/or its affiliates
```
Coach257, line 1: Maybe remove `_kernels` directory level.

Zx55: Unify naming so terms stay consistent across the registry, api and CLI, and keep the previous behavior as much as possible so users do not have to relearn it.

1. `veomni/kernels/_kernels/[kernel1]/...` -> `veomni/ops/kernels/[kernels1]/...` (follow previous behavior)
2. `KERNEL_REGISTRY` -> `OP_REGISTRY`
3. `register_kernel` -> `register_op`
4. `veomni/ops/batch_invariant_ops/...` -> `veomni/ops/batch_invariant_helpers.py` (not register)

tests/kernels/mhc/test_mhc.py
Zx55: Done. `tests/kernels/` now mirrors the kernel families:
- `base/` for registry / `VeomniKernel`
- `attention/` with `magi/` and `flash/`
- other families (`mhc`, `dsa`, `moe_experts`, …) each have their own dir `tol.py` stays at the `tests/kernels/` root because `models_kernel` tests import it.

See [cee30de67](https://github.com/ByteDance-Seed/VeOmni/pull/1142/changes/cee30de6734af9bbcbc367bce7e46712c967179b)

tests/kernels/attention_cases.py
Zx55: Renamed. The mixed-visibility case is now `2d_mask` (`dense_2d_mask` / `magi_2d_mask` / `flex_2d_mask`) instead of `bagel_*`.

See [cee30de67](https://github.com/ByteDance-Seed/VeOmni/pull/1142/changes/cee30de6734af9bbcbc367bce7e46712c967179b).

veomni/models_kernel/transformers/qwen3_5_moe/qwen3_5_moe_gpu_patch_gen_config.py
```diff
@@ -0,0 +1,1115 @@
...
+    key_states = self.k_norm(self.k_proj(hidden_states).view(hidden_shape)).transpose(1, 2)
+    value_states = self.v_proj(hidden_states).view(hidden_shape).transpose(1, 2)
+
+    cos, sin = position_embeddings
+    query_states, key_states = apply_rotary_pos_emb(query_states, key_states, cos, sin)
+
+    if past_key_values is not None:
+        key_states, value_states = past_key_values.update(key_states, value_states, self.layer_idx)
+
+    attn_output, attn_weights = attention_kernel()(
```
Coach257: **`attention` / `moe` selection should come from this module’s config, not `kernel_utils`.**

This file already constructs local handles for RMSNorm (`self.veomni_rms_norm` in `__init__`, call it in `forward`). Attention and MoE do not follow that, and they also do not follow HuggingFace.

HuggingFace binds the impl on the model config at construct time (`config._attn_implementation`, `config._experts_implementation`) and the forward looks that string up. It does not read a process-global singleton.

Here:

1. **`Qwen3_5MoeAttention.forward`** calls `attention_kernel()` every step. That helper intern-looks up `get_kernels_config().attn_implementation`. `build_foundation_model` already wrote the same name onto `self.config._attn_implementation`, and the flash adapter reads `module.config._attn_implementation` again. Two sources of truth; they can disagree.

2. **`PatchedQwen3_5MoeExperts.__init__`** uses `resolve_moe_impl()` (same singleton). The `Qwen3_5MoeModel.__init__` patch that copies `config._moe_implementation` onto `text_config` is leftover from `@use_experts_implementation` — Experts never read it. HF’s field is `_experts_implementation` anyway.

Please make modeling own the choice, same as RMSNorm / HF:

- Attention: in `__init__` (or wherever `config` is available),  
  `self.veomni_attn = VeomniKernel("attention", "standard", self.config._attn_implementation)`  
  and in `forward` only `self.veomni_attn(...)`.  
  Alternatively keep upstream `ALL_ATTENTION_FUNCTIONS.get_interface(self.config._attn_implementation, eager_attention_forward)` — the `attention` registry rows already forward to that table.
- MoE: write the YAML impl onto `config` / `text_config` at construct, then  
  `self.veomni_moe = VeomniKernel("moe_experts", "standard", self.config._moe_implementation)`  
  (or drop the unused propagate if Experts keep reading the singleton — don’t leave both).
- `attention_kernel()` / `resolve_moe_impl()` should not be the selection source from modeling. `get_kernels_config()` can stay at construct/validate time only.

Do not add `if implementation == ...` in `forward`. The handle (or `get_interface`) is the lookup.

configs/multimodal/qwen3_5_moe/qwen3_5_moe_vl.yaml
Zx55: Resolved. Renamed `triton`/`quack`/`npu`/`mlu` to `fused_triton`/`fused_quack`/`fused_npu`/`fused_mlu`.

veomni/models_kernel/transformers/qwen3_5_moe/qwen3_5_moe_gpu_patch_gen_config.py
Zx55: Done. Attention binds `self.veomni_attn` from `config._attn_implementation` in `__init__`; MoE binds `self.veomni_moe` at construct. `forward` only calls the handles. `attention_kernel()` / `resolve_moe_impl()` are gone.

PR #1176, opened 2026-09-10
[ckpt] fix: keep the staged-checkpoint copy off the training backend

TimYangst:
## What

A staged save (`stage_dir`) writes the checkpoint to node-local disk and copies it to `output_dir` afterwards. Its collectives — preparing the staging directory, then the promotion's phases — ran on the default (training) process group, so every rank that is not copying sat in an NCCL collective for the copy's whole duration. They now run on a dedicated gloo group, created at the start of a staged save, whose timeout is the new `save_timeout_seconds`. The same setting applies to the per-slot gloo groups `save_async` writes use.

## Why

On a destination far slower than local disk the copy takes minutes: 302.6 GiB through a fuse HDFS mount measured 11m09s, and staged saves on that job ran 1m33s–8m58s. NCCL's collective timeout is 600s and its watchdog **aborts the process**; gloo raises instead, and its deadline is configurable. Nothing about the copy itself changes — only which process group the ranks wait on while it runs.

## Design

- **One gloo group for staging**, created before `_prepare_stage_dir`, so every rank reaches it before any of them can fail in the write. `dist_utils.any_rank_failed` takes an optional `group`; a gloo group reduces on CPU. Callers that pass no group are unchanged.
- **Cleanup in a `finally`.** The promotion keeps its phases — clear the destination, copy the data, copy the markers — and frees the staged copy in a `finally` rather than an `always` phase. A gloo timeout raises out of a phase where NCCL aborted the process, and as a phase the cleanup would be skipped, stranding a model-plus-optimizer-sized copy on the scratch disk. A phase whose own work failed before the group did keeps that error as the cause, so the log shows more than the connection closing.
- **A failed promotion publishes no marker.** The markers are copied before the phase that agrees on them, so the coordinator retracts every one of them on any failing exit, not only when the copy itself raised. Otherwise a failure of the agreement — a peer gone, the group timed out — leaves them standing over a save that never finished, and a step being rewritten still carries the previous run's manifest until its cursor files are rewritten: manifest plus marker is exactly what resume reads as complete.
- **`save_async` slots get the same deadline.** Each slot's gloo group is created with `save_timeout_seconds`, since an async write also has the ranks not writing wait on its group for the whole duration.
- **Draining an async save agrees on that save's group.** `dist_utils.raise_if_any_rank_failed` takes a `group` as well, and both drains — the one before a slot is reused and the one at train end — reduce on the group the write itself ran on. The ranks whose write finished first wait out the others there rather than on the training backend. One agreement per drained slot, after every slot has been drained.
- **Groups are cached for the life of the process**, like the async slot groups already were. The timeout takes effect when a group is created, and a group broken by a timeout is not rebuilt: nothing catches a failed save, so it has already stopped the run.
- **Validation at startup.** `CheckpointConfig.__post_init__` rejects a non-positive or non-integer `save_timeout_seconds` — the YAML parser passes scalars through untouched and `bool` is an `int` subclass, so `true` would otherwise be a one-second timeout — and warns when the value is set while neither `stage_dir` nor `save_async` is enabled, since those two are the only paths that create a group for it to bound. It also rejects `stage_dir` together with `save_async` there: the checkpointer already did, but not until the first checkpoint was due, which is `save_steps` steps into a run.

`docs/usage/checkpoint.md` notes that a multi-node staged save now needs working Gloo networking, which otherwise only shows at the first checkpoint.

Unchanged: the unstaged path, `stage_dir` still being rejected together with `save_async`, and the marker and manifest handling from #1181.

## Tests

- A staged `save` hands the stage group to both staging steps, the preparation reduces on the group it is handed, and the timeout reaches both DCP writes; `ModelCheckpointManager` forwards `save_timeout_seconds`.
- Every role runs the same three promotion collectives and raises when any rank's copy or marker copy fails; a failure of the group itself still frees the scratch disk, and one that lands after the markers were copied retracts them while leaving the data for the next save to clear.
- An async slot's group is created with the timeout; both drains agree on that slot's group; `any_rank_failed` runs on the group it is handed and, for a gloo group, on CPU without asking for the accelerator's device.
- `save_timeout_seconds` rejects non-positive, non-integer and `bool` values, takes a numpy integer as the integer it is, warns when nothing would use it, and stays quiet when either path is enabled; unset defers to gloo. `stage_dir` with `save_async` fails at construction.
- Four real ranks promote a staged checkpoint over a gloo group: it completes, one rank's failed copy fails every rank without publishing a marker, and a rank that outlives the group timeout makes its peers fail rather than hang, again without a marker and with the staged copy freed.

## Summary by CodeRabbit

- **New Features**
  - Added configurable collective timeouts for staged and asynchronous checkpoint saves.
  - Checkpoint saves now support dedicated communication groups for improved reliability and failure handling.
  - Added validation for timeout values and incompatible staging/asynchronous save settings.

- **Bug Fixes**
  - Improved cleanup and error propagation when distributed checkpoint promotion or asynchronous saves fail.

- **Documentation**
  - Expanded checkpoint configuration and save behavior documentation, including timeout defaults, staging requirements, and networking considerations.


veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -733,18 +771,33 @@ def save(
...
+                # nothing to gain by also backgrounding it, and DCP's stager would
+                # hold a full copy of every rank's shard in host memory to do it.
+                # With stage_dir it is the promotion that runs in the background.
+                save_async=save_async and stage_path is None,
                 save_to_lowest_rank=save_to_lowest_rank,
+                timeout_seconds=save_async_timeout_seconds,
             )
+            if stage_path is not None:
+                if save_async:
+                    cls._promote_in_background(stage_path, checkpoint_dir, save_async_timeout_seconds)
```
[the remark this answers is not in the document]
TimYangst: Fixed in db7b0beb. `_prepare_stage_dir` now retracts the destination's old `.metadata` before the save writes anything into the checkpoint directory, `extra_state` included, so the window is closed for both the inline and the background promotion (it existed on the inline staged path as well). The promotion itself is down to two phases: copy the data, then publish the new marker once every rank's data landed. Covered by `test_the_old_marker_is_retracted_before_extra_state_is_written[False|True]`.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -833,13 +898,108 @@ def wait_for_pending_save(cls) -> None:
...
+                # creation.
+                executor.submit(int).result()
+                cls._promotion_executor = executor
+        except BaseException as e:  # noqa: BLE001 - raised once every rank has agreed
+            error = e
+        if _any_rank_failed(error is not None):
+            raise error or RuntimeError("another rank could not start a checkpoint promotion")
+
+        try:
+            future = cls._promotion_executor.submit(_promote_staged_checkpoint, stage_path, final_path, group)
```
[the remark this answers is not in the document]
TimYangst: Addressed in db7b0beb by going the other way: the cross-rank agreement before the handoff is removed rather than extended. A rank whose thread start or `submit` fails raises out of `save()` like any other rank-local failure in the training loop, and the job is torn down. Its peers' promotions stop at their first reduction and fail on the group timeout without publishing `.metadata`, and the staged copy is freed in a `finally`. Gating the worker behind a second agreement round would add a collective for a failure that is fatal anyway.

Coach257:
After #1181

TimYangst:
> After #1181

Rebased

TimYangst:
@coderabbitai resume

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -833,13 +838,29 @@ def wait_for_pending_save(cls) -> None:
...
+        return cls._checkpoint_process_group
+
     @classmethod
     def execute_save(
         cls,
         save_state: Dict[str, Any],
         storage_writer: FileSystemWriter,
         save_async: bool,
         save_to_lowest_rank: bool = False,
+        timeout_seconds: Optional[int] = None,
```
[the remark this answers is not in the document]
TimYangst: Leaving this as is. The value comes from `train.checkpoint.save_timeout_seconds`, which is fixed for a run, and every `ModelCheckpointManager` of a job reads the same config object — a mismatch can only come from a caller driving `DistributedCheckpointer.save` directly. The caching is documented on the helper (renamed `_get_stage_process_group` in the rebase onto #1181), including why a group broken by a timeout is not rebuilt: nothing catches a failed save, so it has already stopped the run.

veomni/arguments/arguments_types.py
```diff
@@ -855,6 +867,15 @@ class CheckpointConfig:
         metadata={"help": "Save the huggingface format weights to the last checkpoint dir."},
     )
 
+    def __post_init__(self):
```
Coach257, line 870: Add warning as `save_timeout_seconds` only used for `stage_dir` and `save_async` open.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -1063,6 +1090,19 @@ def _drain_slot(cls, slot: str) -> None:
             error = e
         raise_if_any_rank_failed(error, f"the previous async DCP save ({slot})")
```
Coach257: parse `group` here, same as `any_rank_failed`.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -717,6 +731,8 @@ class DistributedCheckpointer(CheckpointerBase):
     # first, which is what draining before each save used to do.
     _save_futures: Dict[str, Any] = {}
     _async_process_groups: Dict[str, Any] = {}
+    # The gloo group a staged save's own collectives run on, created on first use.
+    _stage_process_group: Optional[Any] = None
```
Coach257, line 765: `_stage_process_group` and `_async_process_groups` are cached for the life of the process with no invalidation hook. Unreachable today — `destroy_distributed()` early-returns on NCCL and nothing re-initialises afterwards — so this is a note for whoever adds in-process reuse, not a request

veomni/arguments/arguments_types.py
TimYangst: Done in 95ab6693 — `CheckpointConfig.__post_init__` now warns when `save_timeout_seconds` is set while neither `stage_dir` nor `save_async` is on, since it only bounds the gloo groups those two create.

veomni/checkpoint/dcp_checkpointer.py
TimYangst: Good catch — done in 95ab6693. `raise_if_any_rank_failed` takes a `group` now, and both drains pass the slot's own gloo group: the ranks whose write finished first waited out the rest on the training backend, which is exactly what this PR moves off it.

Coach257 (approved):

veomni/checkpoint/dcp_checkpointer.py
TimYangst: Agreed, and thanks for spelling out why it's unreachable. 

Not adding an invalidation hook in this PR, for two reasons. The only meaningful trigger would be a second `init_distributed()` in the same process, which has no entry point today, so the hook would have no caller and nothing to test against. And "rebuild after a timeout" is not a safe policy either: `new_group` is itself collective, and a group broken by a timeout leaves ranks failing at different points, so they would hang in the rebuild — the run is already over at that point, since nothing catches a failed save.

veomni/arguments/arguments_types.py
```diff
@@ -855,6 +867,24 @@ class CheckpointConfig:
...
+            return
+
+        # ``bool`` is an ``int`` subclass and the parser passes YAML through
+        # untouched: a stray ``true`` would be a one-second timeout.
+        if type(self.save_timeout_seconds) is not int or self.save_timeout_seconds <= 0:
+            raise ValueError(f"save_timeout_seconds must be a positive integer, got {self.save_timeout_seconds!r}.")
+
+        # It only bounds the gloo groups those two paths create; a direct
+        # synchronous save has none and keeps the training backend's timeout.
+        if self.stage_dir is None and not self.save_async:
```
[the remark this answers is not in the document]
TimYangst: Fixed in 165ed64d — switched to truthiness, since an empty `stage_dir` is what the checkpointer reads as unset. The warning test is parametrized over both.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -679,19 +696,16 @@ def copy_markers() -> None:
...
-        shutil.rmtree(stage_path, ignore_errors=True)
-
-    _promotion_phase(state, clear_destination, participates=is_coordinator)
-    _promotion_phase(state, copy_this_nodes_files, participates=is_node_leader)
-    _promotion_phase(state, copy_markers, participates=is_coordinator)
-    _promotion_phase(state, drop_staged_copy, participates=is_node_leader, always=True)
+    try:
+        _promotion_phase(state, clear_destination, participates=is_coordinator, group=group)
+        _promotion_phase(state, copy_this_nodes_files, participates=is_node_leader, group=group)
+        _promotion_phase(state, copy_markers, participates=is_coordinator, group=group)
```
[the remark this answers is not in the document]
TimYangst: Real, and fixed in 165ed64d. The coordinator now retracts the markers on any failing exit rather than only when the copy itself raised, so a failure of the closing agreement cannot publish them. Worth noting why it matters: `GlobalStateCallback` removes the previous run's `checkpoint_manifest.json` only after the model save, so a step being rewritten still carries it, and manifest plus marker is what resume reads as complete. The data is left in place — without a marker the destination reads as incomplete, and the next save clears it. Test added: `test_a_group_failure_after_the_markers_retracts_them` fails the third reduction after `copy_markers` succeeds.

PR #1181, opened 2026-09-11
[BREAKING][ckpt, lora, docs] feat: split the checkpoint step directory by state owner

Coach257:
### What does this PR do?

Reshapes a checkpoint step directory so each artifact sits where its owner puts it, and makes one writer serve both a single-model job and a multi-module (SeedOmni V2) one.

```
checkpoints/global_step_{N}/
├── checkpoint_manifest.json      # completion marker, rank 0 writes it last
├── model/[<module>/]
│   ├── ckpt/                     # weights, DCP
│   ├── optimizer/                # optimizer state, DCP
│   └── lr_scheduler.pt           # replicated pickle
├── loader/rank_{R}.pt            # dataloader / sampler cursor
├── extra_state/rank_{R}.pt       # global_step, RNG, meters
├── hf_ckpt/[<module>/]           # full-model export, not resume
└── lora_ckpt/[<module>/]         # LoRA adapter export, not resume
```

Four things change:

- **Weights and optimizer become separate DCP directories.** This buys file separation, not read selectivity — reading the weights alone always worked by omitting `optimizer` from the load state, since `dcp.load` reads only the keys it is asked for. What a single directory made impossible was *moving* them apart afterwards, because both land in the same `__{i}_{rank}.distcp` files. Optimizer state is ~2x the weights under Adam and useless outside the run, so shipping, archiving or converting a checkpoint wants the weights alone.
- **The job cursor splits into `loader/` and `extra_state/`.** An Energon / multisource-sampler state is large and is the part a job may want to drop on its own: resuming weights onto a different dataset means keeping `extra_state/` and discarding `loader/`.
- **`checkpoint_manifest.json` is a step-level completion marker.** DCP's `.metadata` marks one *directory* complete, and there are now at least two per module plus per-rank cursor files DCP knows nothing about, so no single `.metadata` can stand for the step. Rank 0 writes the manifest after every module's DCP save has returned and every rank's files are down; resume discovery accepts a step only when it is present.
- **All path construction moves into `veomni/checkpoint/layout.py`.** `module=""` collapses `model/<module>/` to `model/`, so single-model and V2 are one code path differing only in a subfolder name.

Also fixes a latent path bug: `ModelCheckpointManager` built its own `save_dir` while `DistributedCheckpointer` appended `global_step_{N}` itself and ignored `checkpoint_subfolder`, so any non-empty subfolder would have made write and read disagree. Both now call the same helpers. `checkpoint_subfolder` is renamed `module_name`, which is all it ever meant.

Contract: `docs/usage/checkpoint.md`.

Related: #1179 (manager / global_state split), #1125 (runtime branch this was extracted from).

**Breaking.** New saves are not readable by an older VeOmni. The other direction is covered: resume reads three older shapes (0.1.12 mixed `extra_state`, 0.2.x flat `trainer_state_rank_{R}.pt`, and a fused DCP directory) through the throwaway loader in `veomni/checkpoint/legacy_v0_1_12.py`. Two API signatures change — see **API and Usage Example**.

### Checklist Before Starting

- Search for relative PRs/issues and link here: #1179, #1125

### Test

Extended existing CI-enumerated tests under `tests/checkpoints/` and `tests/lora/` (both directories are enumerated in the GPU and NPU unit workflows). No new files, so nothing needs wiring into the workflows.

New in `tests/checkpoints/test_dcp_checkpointer.py`:

- `TestCheckpointLayoutRoundTrip` — `test_weights_and_optimizer_are_separable_on_disk` (asserts every metadata key under `model/ckpt/` starts with `model`, so nobody can quietly fuse them back), `test_resume_restores_weights_optimizer_and_scheduler`, `test_modules_of_one_job_do_not_share_a_directory`, `test_resume_reads_a_pre_split_checkpoint` (legacy fused directory).
- `TestWaitForPendingSave` — `test_drains_every_slot_before_raising` and `test_concurrent_slots_get_their_own_group_and_are_not_drained`, which pin the two-slot async design below.
- `test_save_without_optimizer_writes_only_the_weights`; `test_load_passes_partial_planner_to_dcp` now asserts two `dcp.load` calls.

`tests/checkpoints/test_checkpoint_callback.py`: `test_save_lora_writes_the_adapter_to_its_own_export_dir`, `test_save_waits_for_pending_dcp` (asserts `extra_state/rank_0.pt`, `loader/rank_0.pt` and the manifest all exist), plus the three 0.1.12 / 0.2.x legacy-resume cases.

Ran the full `tests/checkpoints` suite (124 tests, including the 8-GPU `save_async=True` end-to-end runs) and `tests/lora` save/load; all pass. `make quality` passes. The on-disk tree from a real run was diffed against the doc.

### API and Usage Example

Two signature changes, both breaking for direct callers:

- `DistributedCheckpointer.save()` / `.load()` drop `storage_writer` / `storage_reader`. With two output directories per call a single writer has no well-defined meaning. Nothing in the repo passed them outside one unit test.
- `ModelCheckpointManager.checkpoint_subfolder` → `module_name`.

Saving is unchanged at the call site — one call still lays down weights, optimizer and scheduler:

```python
checkpointer.save(
    save_path,
    {"model": model, "optimizer": optimizer, "lr_scheduler": lr_scheduler},
    global_steps=step,
    module="",  # "" for a single model; the module name for a V2 job
)
```

Derive paths instead of hardcoding them:

```python
from veomni.checkpoint.layout import step_dir, weights_dir, optimizer_dir

step = step_dir(args.train.checkpoint.save_path, global_step)
weights_dir(step)                 # <step>/model/ckpt
weights_dir(step, "janus_llama")  # <step>/model/janus_llama/ckpt
optimizer_dir(step)               # <step>/model/optimizer
```

`scripts/merge_dcp_to_hf.py --load-dir` now takes either a step directory (resolves to `model/ckpt`, writes the export to `<step>/hf_ckpt`) or a DCP directory directly, which is what a pre-split checkpoint needs:

```bash
python3 scripts/merge_dcp_to_hf.py --load-dir ./exp/run/checkpoints/global_step_500
```

Resuming a LoRA adapter is unchanged, but the adapter moved to its own directory:

```yaml
model:
  lora_config:
    lora_adapter: ./exp/my_run/checkpoints/global_step_500/lora_ckpt
```

### Design & Code Changes

- **`veomni/checkpoint/layout.py` (new)** — every directory and file name, plus `write_manifest` / `read_manifest`. Single source of truth for both the save and the load side.
- **`DistributedCheckpointer`** — `save` issues two `execute_save` calls (weights, optimizer) and one scheduler write, all through layout helpers; `load` reads the two directories and falls back to a fused legacy directory when `model/ckpt/` is absent. Staging promotion now handles a `.metadata` per nested directory and publishes all of them last, so a reader sees either the previous complete checkpoint or none.
- **Async saves stay concurrent.** `dcp.async_save` hands its process group to a background thread that runs the whole save, collectives included, so two saves sharing one group would have two threads issuing collectives on it at once. Each slot therefore gets its own future and its own Gloo group, and `execute_save` drains only its own slot. `wait_for_pending_save` drains every slot even when one raises — a failed weights write must not leave the optimizer write running into the next step's collectives.
- **`GlobalStateCallback`** — splits its payload into `loader/rank_{R}.pt` and `extra_state/rank_{R}.pt`, then rank 0 publishes the manifest after a barrier. A missing `loader/` file on resume is a warning, not an error.
- **`ModelCheckpointManager`** — `save_dir` / `load_dir` / `hf_export_dir` / `lora_export_dir` all derive from layout; LoRA exports to `lora_ckpt/` instead of sharing `hf_ckpt/`, because a future LoRA merge writes both for the same step and a reader must tell them apart by path.
- **`veomni/utils/checkpoint_utils.py`** — resume discovery accepts a step on `checkpoint_manifest.json`, falling back to a step-root `.metadata` for pre-split checkpoints.
- **`veomni/checkpoint/legacy_v0_1_12.py`** — now covers all three older shapes. Deleting this file and its imports drops legacy resume; the `extra_state/` name collision between eras is resolved on the file name inside (`rank_{R}.pt` vs `extra_state_rank_{R}.pt`), and that rule is written down in the module.
- **Docs** — `docs/usage/checkpoint.md` rewritten as the layout contract, including the V2 nesting; LoRA, trainer and conversion pages updated.

- Applied pre-commit checks
- Added/updated documentation
- Tests: extended an existing CI-enumerated test, **or** added a new test and wired it into `gpu_unit_tests.yml` (and `npu_unit_tests.yml` where applicable), **or** explained in **Test** why none is needed. Note that CI enumerates test files individually — a new file outside `tests/data/` and `tests/ops/` does not run until it is listed. See [.agents/knowledge/testing.md](.agents/knowledge/testing.md).

## Summary by CodeRabbit

- **New Features**
  - Checkpoints now use a documented hierarchical layout with separate model, optimizer, scheduler, loader, and extra-state data, plus optional Hugging Face and LoRA exports.
  - Checkpoint completion is tracked with a manifest, including compatibility with legacy layouts.
  - Scheduler state is saved and restored, including from legacy VeOmni 0.1.12 checkpoints.
  - LoRA adapters are stored in a dedicated directory within each checkpoint step.
  - Checkpoint conversion accepts complete step directories.

- **Documentation**
  - Expanded checkpoint layout, conversion, resume, and configuration guidance.

### DIFF
--- a/docs/usage/checkpoint.md
+++ b/docs/usage/checkpoint.md
@@ -2,8 +2,9 @@
 
 This page is the on-disk contract for a VeOmni training run: which directories
 are written, what each file holds, and how that differs from the previous
-layout. Resume reads **this** layout only. Older `extra_state/` pickles are
-not loaded.
+layout. New writes use **this** layout. Resume also reads VeOmni 0.1.12
+`extra_state/` pickles through a throwaway loader (see
+[Legacy resume (VeOmni 0.1.12)](#legacy-resume-veomni-0112)).
 
 `train.checkpoint.output_dir` is the run root. Every per-step artifact — DCP
 shards, scheduler, job cursor, HuggingFace export, and LoRA adapter — lives
@@ -75,7 +76,8 @@ part of the published checkpoint.
 ## Previous layout
 
 Checkpoints written before this layout used a per-rank `extra_state` pickle
-instead of `lr_scheduler.pt`. That path is **not** read on resume.
+instead of `lr_scheduler.pt`. Resume still reads those pickles; see
+[Legacy resume (VeOmni 0.1.12)](#legacy-resume-veomni-0112).
 
 ```
 <output_dir>/
@@ -100,10 +102,31 @@ it contained changed once:
 | `CheckpointerCallback` (before the manager / global_state split) | Mixed bag: `lr_scheduler.state_dict()`, and the job cursor (`global_step`, dataloader, RNG, meters). Rank-local cursor is why the file was per-rank. Changing world size required a matching `extra_state_rank_{R}.pt`. |
 | After the split, before this layout | **Only** `lr_scheduler.state_dict()`. The job cursor moved to `trainer_state_rank_{R}.pt` beside the shards. The per-rank `extra_state` files were leftover: the scheduler is replicated, so every rank wrote the same pickle. |
 
-There is no loader for `extra_state/`. A directory that only has
-`extra_state_rank_*.pt` fails resume if a scheduler is expected: the load
-raises rather than silently restarting the LR at step 0. Re-run from the
-HuggingFace / adapter export, or convert the pickle yourself.
+## Legacy resume (VeOmni 0.1.12)
+
+Temporary. New saves never write `extra_state/`. This loader exists so a
+0.1.12 (and later extra_state-era) checkpoint can still resume; it will be
+removed.
+
+**Where:** `veomni/checkpoint/legacy_v0_1_12.py`. The only call sites are:
+
+- `DistributedCheckpointer._load_lr_scheduler` — if `lr_scheduler.pt` is
+  missing, load `extra_state/extra_state_rank_{R}.pt` (`lr_scheduler` key).
+- `GlobalStateCallback.load_global_state` — if `trainer_state_rank_{R}.pt` is
+  missing, load the same pickle's job cursor (`global_step`, dataloader, RNG,
+  meters) when `global_step` is present.
+
+**How to drop it:** delete `veomni/checkpoint/legacy_v0_1_12.py` and the two
+imports that load it (search for `legacy_v0_1_12`). After that, a directory
+with only `extra_state_rank_*.pt` fails resume if a scheduler is expected.
+
+Current files always win: `lr_scheduler.pt` is not mixed with extra_state, and
+`trainer_state_rank_{R}.pt` is not mixed with a 0.1.12 job cursor.
+
+0.1.12 LoRA export lived at `<output_dir>/global_step_{N}/` (sibling of
+`checkpoints/`). That path is an inference artifact; DCP resume does not read
+it. Point `PeftModel.from_pretrained` at the old directory if you need the
+adapter.
 
 ## Related pages
 
--- a/docs/usage/checkpoint.md
+++ b/docs/usage/checkpoint.md
@@ -1,0 +1,109 @@
+# Checkpoint layout
+
+This page is the on-disk contract for a VeOmni training run: which directories
+are written, what each file holds, and how that differs from the previous
+layout. Resume reads **this** layout only. Older `extra_state/` pickles are
+not loaded.
+
+`train.checkpoint.output_dir` is the run root. Every per-step artifact — DCP
+shards, scheduler, job cursor, HuggingFace export, and LoRA adapter — lives
+under `output_dir/checkpoints/global_step_{N}/` (`train.checkpoint.save_path`).
+Tokenizer / processor / config sidecars are written once to `model_assets/`.
+
+## Current layout
+
+```
+<output_dir>/                          # train.checkpoint.output_dir
+├── checkpoints/                       # train.checkpoint.save_path
+│   └── global_step_{N}/
+│       ├── .metadata                  # DCP completion marker. A reader treats
+│       │                              # this directory as a complete checkpoint
+│       │                              # only when this file is present.
+│       ├── __{i}_{rank}.distcp        # PyTorch DCP shards. Together they hold
+│       │                              # the model weights and the optimizer
+│       │                              # state. LoRA / PEFT runs store only
+│       │                              # trainable adapter tensors here; the
+│       │                              # frozen base is reloaded from
+│       │                              # model.model_path.
+│       ├── lr_scheduler.pt            # One pickle: lr_scheduler.state_dict().
+│       │                              # Replicated across ranks; rank 0 writes
+│       │                              # it. Every rank reads this same file
+│       │                              # on resume.
+│       ├── trainer_state_rank_{R}.pt  # Per-rank job cursor. Rank R writes and
+│       │                              # rank R restores. Contents:
+│       │                              #   global_step
+│       │                              #   train_dataloader   (sampler / iterator)
+│       │                              #   environ_meter
+│       │                              #   channel_loss_callback
+│       │                              #   torch_rng_state
+│       ├── adapter_config.json        # LoRA / PEFT export (adapter runs only).
+│       ├── adapter_model.safetensors  # Inference artifact; not used to resume
+│       │                              # the optimizer. Same directory as DCP.
+│       └── hf_ckpt/                   # Optional full-model HuggingFace export
+│           ├── model*.safetensors     #   (train.checkpoint.save_hf_weights).
+│           ├── model.safetensors.index.json
+│           ├── config.json
+│           └── tokenizer files
+└── model_assets/                      # Tokenizer / processor / config, written
+                                       # once at train start (rank 0).
+```
+
+A multi-module job (Omni) nests the DCP shards, `lr_scheduler.pt`, LoRA adapter
+files, and `hf_ckpt/` one level deeper under the module name
+(`ModelCheckpointManager.checkpoint_subfolder`). `trainer_state_rank_{R}.pt`
+stays at `checkpoints/global_step_{N}/` — there is one job cursor, not one per
+module.
+
+`train.checkpoint.stage_dir`, when set, is scratch: the shards and
+`lr_scheduler.pt` are written there first and copied to `global_step_{N}/`
+only after DCP finishes. The staging directory is not part of the published
+checkpoint.
+
+### Who writes what
+
+| File | Writer | Why |
+|------|--------|-----|
+| `__*.distcp`, `.metadata` | All ranks, via DCP | Collective; shards are rank-local. |
+| `lr_scheduler.pt` | Rank 0 | Scheduler state is replicated. Peers still join the save reduction so a failed write cannot leave them inside `dcp.save`. |
+| `trainer_state_rank_{R}.pt` | Rank R | Dataloader cursor, RNG, and meters are rank-local. Restoring rank 0's cursor on every rank would replay rank 0's shard. |
+| `model_assets/` | Rank 0 | Config / tokenizer dump. |
+| `hf_ckpt/`, LoRA adapter | Rank 0 after a DCP gather | Export, not resume. |
+
+## Previous layout
+
+Checkpoints written before this layout used a per-rank `extra_state` pickle
+instead of `lr_scheduler.pt`. That path is **not** read on resume.
+
+```
+<output_dir>/
+├── checkpoints/
+│   └── global_step_{N}/
+│       ├── .metadata
+│       ├── __{i}_{rank}.distcp        # Model + optimizer (same role as today).
+│       ├── extra_state/
+│       │   └── extra_state_rank_{R}.pt
+│       └── hf_ckpt/                   # Optional HF export (same role as today).
+├── model_assets/
+└── global_step_{N}/                   # LoRA adapter used to live here, as a
+    ├── adapter_config.json            # sibling of checkpoints/. It now sits in
+    └── adapter_model.bin              # checkpoints/global_step_{N}/ instead.
+```
+
+`extra_state/extra_state_rank_{R}.pt` was a per-rank `torch.save` pickle. What
+it contained changed once:
+
+| Era | Contents of `extra_state_rank_{R}.pt` |
+|-----|----------------------------------------|
+| `CheckpointerCallback` (before the manager / global_state split) | Mixed bag: `lr_scheduler.state_dict()`, and the job cursor (`global_step`, dataloader, RNG, meters). Rank-local cursor is why the file was per-rank. Changing world size required a matching `extra_state_rank_{R}.pt`. |
+| After the split, before this layout | **Only** `lr_scheduler.state_dict()`. The job cursor moved to `trainer_state_rank_{R}.pt` beside the shards. The per-rank `extra_state` files were leftover: the scheduler is replicated, so every rank wrote the same pickle. |
+
+There is no loader for `extra_state/`. A directory that only has
+`extra_state_rank_*.pt` will skip the scheduler (and, for the mixed-bag era,
+will not restore the job cursor from that pickle either). Re-run from the
+HuggingFace / adapter export, or convert the pickle yourself.
+
+## Related pages
+
+- [Checkpoint conversion](checkpoint_conversion.md) — DCP shards → HuggingFace safetensors (`scripts/merge_dcp_to_hf.py`).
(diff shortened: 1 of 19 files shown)


veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -724,7 +745,17 @@ def save(
...
+        # copied with the shards, before ``.metadata`` is published: writing it
+        # into a still-valid destination would pair a new scheduler with the
+        # previous model and optimizer.
         stage_path = _prepare_stage_dir(stage_dir, path) if stage_dir else None
+        if stage_path is not None:
+            cls._save_lr_scheduler(checkpoint_dir=stage_path, state=state)
+        elif save_async:
+            # Async ``execute_save`` returns before DCP finishes, so the sidecar
+            # has to go out first. ``stage_dir`` is rejected with ``save_async``.
+            cls._save_lr_scheduler(checkpoint_dir=checkpoint_dir, state=state)
```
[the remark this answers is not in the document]
Coach257: Not taking the unstaged-always-stage + wait-for-future change.

save_async exists so save() can return before DCP finishes and overlap
the next step. Waiting on save_future.result() inside save() makes that
path synchronous.

Atomic publish is already stage_dir: write shards + lr_scheduler.pt on
local scratch, then _promote_staged_checkpoint copies data before
.metadata. A sibling temp on the destination filesystem would double
writes on the slow store that stage_dir is meant to avoid. stage_dir
and save_async stay incompatible for the same reason (save() returns
while the staged copy is still in flight).

On the sync unstaged path the sidecar is written only after dcp.save
returns, so a failed DCP cannot refresh lr_scheduler.pt (covered by
test_failed_unstaged_overwrite_does_not_change_previous_scheduler). A
missing sidecar on resume raises instead of silently resetting the LR.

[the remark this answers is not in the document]
Coach257: The remaining cases assume an unstaged save overwrites an already-published `global_step_{N}/`. That is not the training path: each save writes a new `global_step_{N}/` with a new N, so a failed step cannot mutate a previously published directory.
Resume only treats a step directory as complete when `.metadata` is present (`dcp_get_last_iteration`). Unstaged order is now `lr_scheduler.pt` first, then DCP; `.metadata` is DCP’s completion marker and is written last. A failed save of the new step has no marker and is skipped; the previous step is unchanged.
`save()` still does not wait on `save_future.result()`. `stage_dir` stays the opt-in path for writing on local scratch and copying to a slow destination, with `.metadata` published last.

docs/usage/checkpoint.md
```diff
@@ -0,0 +1,135 @@
...
+## Current layout
+
+```
+<output_dir>/                          # train.checkpoint.output_dir
+├── checkpoints/                       # train.checkpoint.save_path
+│   └── global_step_{N}/
+│       ├── .metadata                  # DCP completion marker. A reader treats
+│       │                              # this directory as a complete checkpoint
+│       │                              # only when this file is present.
+│       ├── __{i}_{rank}.distcp        # PyTorch DCP shards. Together they hold
```
FoolPlayer: Maybe we can seperate as models (for llm/vit/dit...) / optimizer / lora (with lora config and lora ckpt) / hf_ckpt (config/tokenizer files) / extra_state (like lr_schedule, per rank states...)

Luosuu (approved):

docs/usage/checkpoint.md
Coach257: Done.

veomni/checkpoint/layout.py
```diff
@@ -0,0 +1,187 @@
...
+        "modules": list(modules or []),
+    }
+    path = manifest_path(step_root)
+    os.makedirs(step_root, exist_ok=True)
+    # Write-then-rename: a reader must never see a half-written marker, since the
+    # whole point of the file is that its presence means "complete".
+    tmp_path = f"{path}.tmp"
+    with open(tmp_path, "w") as handle:
+        json.dump(payload, handle, indent=2)
+    os.replace(tmp_path, path)
```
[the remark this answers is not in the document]
Coach257: Took the first option: documented the contract, not adding the abstraction.

Checkpoint output has always been POSIX-only and this PR doesn't change that. On
`main` today every writer in the step directory is already a plain filesystem
call — `os.makedirs` + `shutil.copyfile` in staged promotion, `torch.save` for
the per-rank state, and DCP's own `FileSystemWriter`, which has no remote
backend. `write_manifest` joins a set of writers that was already uniformly
local; it doesn't make a previously remote-safe path local.

The read side is the genuinely asymmetric part, and it also predates this PR:
`veomni/utils/checkpoint_utils.py` imports `exists` / `isdir` / `listdir` from
`hdfs_io`, so resume *discovery* will happily walk an `hdfs://` tree that nothing
can write to. That's worth fixing, but it's a pre-existing inconsistency in a
module this PR touches only to add the manifest check. Routing all checkpoint I/O
through one filesystem abstraction would also still be incomplete while DCP
itself only writes POSIX, so it belongs in its own change.

`docs/usage/checkpoint.md` now states the contract: `train.checkpoint.output_dir`
must be a POSIX path. An HDFS FUSE mount (`/mnt/hdfs/...`) qualifies and is how
people actually checkpoint to remote storage; an `hdfs://` URL does not, and
unlike a model path it is not staged to local first.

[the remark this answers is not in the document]
Coach257: Not doing this in this PR, and reverting the documentation claim that went with it.

The behaviour you describe is real but predates this PR and is unchanged by it.
Every writer under a step directory is already a plain filesystem call on `main` —
`os.makedirs` + `shutil.copyfile` in staged promotion, `torch.save` for the
per-rank state, and DCP's `FileSystemWriter`, which has no remote backend.
`write_manifest` joins a set of writers that was already uniformly local; it
neither introduces the limitation nor widens it.

Enforcing a new startup failure for that pre-existing condition does not belong in
a layout change. `veomni/arguments/parser.py` already treats an `hdfs://`
`output_dir` as legal (`save_args` stages the YAML locally and uploads it), and
`helper.convert_hdfs_fuse_path` is a stub in this repo, so rejecting the value
here could contradict a path that is not visible in open source. That is a call
for a change that is actually about the storage contract, where the read side
(`checkpoint_utils.py` importing `exists` / `isdir` / `listdir` from `hdfs_io`)
can be fixed in the same breath.

I have removed the POSIX-path paragraph I added earlier, so this PR makes no claim
it does not enforce.

Coach257:
> Review

Not acting on this, because the precondition cannot be produced by this codebase.

The only caller of `DistributedCheckpointer.save()` is
`ModelCheckpointManager.save_dcp`, and it always passes all three artifacts
(`model`, `optimizer`, `lr_scheduler`) off the trainer. The `if "optimizer" in
state and state["optimizer"] is not None` branch is therefore always taken for a
training job, so a staged save that omits the optimizer or the scheduler is not a
state this repo reaches. The scenario needs the same `global_step_{N}` written
twice *with different composition*; a restart that resumes and re-reaches N writes
the same composition, so every marker and file is replaced.

The proposed cleanup would also apply to the staged path only, and the unstaged
path has the identical property — writing into an existing `model/` does not
remove what it did not write. Adding it on one side makes `stage_dir` change the
overwrite semantics of a checkpoint, which is worse than the uniform behaviour we
have now. If we want "a save replaces the step's model directory" as a contract,
it has to hold for both paths and it is not a quick win.

Worth noting for the residual case: leftover `__{i}_{rank}.distcp` files from a
different world size are dead weight rather than stale state, since DCP reads only
what `.metadata` lists, and `.metadata` is always republished.

Coach257:
> scripts/merge_dcp_to_hf.py (1)
> > `88-88`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > **Point LoRA conversion guidance at the actual adapter export.**
> > Current checkpoint saves write `adapter_config.json` under `global_step_*/lora_ckpt/`, but this warning points users to the step root. Following it makes `--adapter-config-path` target a nonexistent file. Include `lora_ckpt/` here and in the option help.

Already fixed in the commit after the review that raised this. Both places now
point at `lora_ckpt/`:

- the no-`--adapter-config-path` warning →
  `output_dir/checkpoints/global_step_*/lora_ckpt/`
- the `--adapter-config-path` help →
  `<output_dir>/checkpoints/global_step_*/lora_ckpt/adapter_config.json`

This review ran on 384bff8; the fix landed after it.

docs/usage/checkpoint.md
```diff
@@ -0,0 +1,241 @@
+# Checkpoint layout
```
FoolPlayer, line 1: Maybe simplify this file, rename the tile with just `Path hold`... not a question. And remove some unnecessary doc, just telling the ckpt structure is fine.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -565,53 +562,89 @@ def _promote_staged_checkpoint(stage_path: str, final_path: str) -> None:
...
+                if filename == metadata_name:
+                    rels.append(os.path.relpath(os.path.join(dirpath, filename), stage_path))
+        return sorted(rels)
+
     def drop_stale_marker() -> None:
         """Stop advertising the previous checkpoint before overwriting its shards."""
         os.makedirs(final_path, exist_ok=True)
-        if os.path.exists(final_metadata):
-            os.remove(final_metadata)
+        for rel in staged_markers():
```
FoolPlayer: [P2] Please invalidate the step-level manifest before overwriting a published checkpoint. This only removes nested `.metadata` files; if staged copying fails, the old manifest survives and `load_path: auto` still selects the incomplete step. Republish it only after the whole step succeeds.

veomni/checkpoint/dcp_checkpointer.py
```diff
@@ -772,66 +847,132 @@ def load(
...
+            return
+        rank = dist.get_rank() if dist.is_initialized() else 0
+        futures = cls._save_futures
+        cls._save_futures = {}
+        error: Optional[BaseException] = None
+        for slot, future in futures.items():
+            try:
+                logger.info(f"[RANK {rank}] waiting for pending DCP save ({slot}) to end...")
+                future.result()
+            except Exception as e:  # noqa: BLE001 - re-raised once every rank has agreed
```
FoolPlayer: [P2] PyTorch's `CheckpointException` inherits directly from `BaseException`, so this catch misses real DCP failures. If the weights future raises it, the optimizer future is never drained and its reference is lost because `_save_futures` was already cleared. Please catch it explicitly (or catch `BaseException`), drain every slot, then propagate; cover this exception type in the test.

veomni/utils/checkpoint_utils.py
```diff
@@ -52,11 +63,17 @@ def _validate_dcp_checkpoint_entry(checkpoints_dir: str, entry: str):
...
         return None
 
-    metadata_path = os.path.join(checkpoint_path, ".metadata")
-    if not exists(metadata_path):
-        return None
+    # Imported here rather than at module scope: ``veomni.checkpoint.layout``
+    # imports this module for the step-directory prefix.
+    from ..checkpoint.layout import MANIFEST_FILENAME
+
+    if exists(os.path.join(checkpoint_path, MANIFEST_FILENAME)):
```
FoolPlayer: [P2] Export-triggered DCP saves never get this manifest unless they also hit the regular DCP cadence. For example, with `save_steps=100`, `save_epochs=0` and HF export at train end (step 150), `_prepare_export()` saves DCP at 150, but auto-resume falls back to 100. Please also finalize job state and publish the manifest for DCP saves triggered by HF/LoRA export.

Coach257: Done. Published through `on_train_end`.

veomni/checkpoint/dcp_checkpointer.py
Coach257: Fixed

veomni/checkpoint/dcp_checkpointer.py
Coach257: Done. Add `cls._unpublish_step(checkpoint_dir)`

docs/usage/checkpoint.md
Coach257: Done.

veomni/checkpoint/dcp_checkpointer.py
Coach257: For `torch.save / torch.load / os / shutil` use `Exception`.

PR #1185, opened 2026-09-12
fix: use base model when building DPO reference

### LINKED ISSUE #1129: Question: TextDPOTrainer initialization when used from the pinned verl-das stack
Hello VeOmni and verl-das maintainers,

We are integrating the text DPO trainer through the `verl-das` stack on a Hygon BW1000 environment and would like to confirm the intended initialization path.

Our source revisions are:

  - `verl-das`:  `54d9a34ecc09afd9a141eaff1d0f9b967ef1f53d`
  - `third_party/VeOmni`:  `cbb3e012936912fd9ce063241e1fb77e8d564d2f`
  - `third_party/verl`: pinned by the  above `verl-das` checkout

The VeOmni source is used from:

```text
  third_party/VeOmni/tasks/
  train_text_dpo.py
  third_party/VeOmni/veomni/trainer/
  text_dpo_trainer.py
```

The official VeOmni documentation shows:

```bash
bash train.sh tasks/train_text_dpo.py configs/text/qwen3_dpo.yaml
```

In our verl-das integration, the launcher invokes the same official entrypoint through the underlying distributed command:

```bash
torchrun \
    --standalone \
    --nnodes=1 \
    --nproc-per-node=4 \
    /opt/verl-das/third_party/VeOmni/
    tasks/train_text_dpo.py \
    /workspace/.../effective.yaml
```

The launcher also sets the distributed environment variables normally exported by VeOmni/train.sh, including:

```bash
  TOKENIZERS_PARALLELISM=false
  TORCH_NCCL_AVOID_RECORD_STREAMS=1
  NCCL_DEBUG=WARN
```

During trainer initialization, the process reaches TextDPOTrainer._build_reference_model() and reports:

```bash
  AttributeError: 'TextDPOTrainer' object has no attribute 'model'
```

  The relevant code appears to be:

```python
  class TextDPOTrainer:
      def __init__(self, args):
          self.base =
          BaseTrainer.__new__(BaseTrainer
          )
          ...
```

Later, _build_reference_model() contains:

```python
  if hasattr(self.model, "get_parallel_plan"):
      ...
```

Elsewhere in the same trainer, the policy model appears to be accessed as self.base.model.

Could you please clarify the expected behavior for this composition pattern?

  1. When VeOmni DPO is invoked through the pinned verl-das stack, is there an additional initialization step that     should provide self.model?

  2. Is the direct torchrun invocation above expected to be behaviorally equivalent to invoking VeOmni/train.sh?

  3. Should integrations expose the policy model as self.model, or should this reference-model path use    self.base.model?

  4. Is there a recommended verl-das or VeOmni revision containing the intended initialization pattern?

  We may be missing an expected integration convention between verl-das and VeOmni, so we would appreciate guidance on the supported invocation and object composition contract. We have not modified the vendored VeOmni source locally.

OnePunchMonk:
Fixes #1129.

TextDPOTrainer stores the policy model on `self.base.model`, but `_build_reference_model()` dereferenced the nonexistent `self.model` while reading the parallel plan. This caused DPO initialization to fail before reference-model parallelization.

Changes:
- read `get_parallel_plan()` from `self.base.model`;
- add a focused regression test.

Validation:
- pre-fix Modal run: reproduced `AttributeError: TextDPOTrainer has no attribute model`;
- patched Modal run: `1 passed`;
- `git diff --check`: passed.

The Modal test used the repository source mounted into a clean Python 3.12 environment.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected reference model setup to use the policy model’s parallel configuration.
  - Ensured the appropriate CPU-loaded parameter setting is preserved when building the reference model.

- **Tests**
  - Added coverage verifying that the reference model receives the expected parallelization setting.

### DIFF
--- a/tests/trainer/test_text_dpo_trainer_reference_model.py
+++ b/tests/trainer/test_text_dpo_trainer_reference_model.py
@@ -1,0 +1,44 @@
+from types import SimpleNamespace
+from unittest.mock import Mock
+
+import pytest
+
+from veomni.trainer import text_dpo_trainer
+
+
+def _args():
+    fsdp_config = SimpleNamespace(
+        reshard_after_forward=False,
+        forward_prefetch=False,
+        offload=False,
+        offload_pin_memory=False,
+        max_load_broadcast_size=None,
+    )
+    accelerator = SimpleNamespace(init_device="cpu", fsdp_config=fsdp_config)
+    model = SimpleNamespace(
+        config_path="config",
+        model_path="weights",
+        accelerator=accelerator,
+        ops_implementation=SimpleNamespace(),
+        model_config={},
+        basic_modules=[],
+        broadcast_model_weights_from_rank0=False,
+    )
+    return SimpleNamespace(model=model, dpo_config=SimpleNamespace(refer_model_precision="float32"))
+
+
+def test_build_reference_model_reads_parallel_plan_from_base_model(monkeypatch):
+    trainer = text_dpo_trainer.TextDPOTrainer.__new__(text_dpo_trainer.TextDPOTrainer)
+    policy = Mock()
+    policy.get_parallel_plan.return_value.cpu_load_param_name = "policy_param"
+    trainer.base = SimpleNamespace(args=_args(), model=policy)
+
+    reference = Mock(_no_split_modules=[])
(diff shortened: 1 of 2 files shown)


Coach257:
Thanks! This bug will be fixed in #1125.

PR #1187, opened 2026-09-14
[agent, model] feat: enhance veomni-new-model skill and onboard SeedVR2

ruchu127:
### What does this PR do?

Extends `/veomni-new-model` to onboard existing models from Transformers, Diffusers, other frameworks, or standalone repositories, and demonstrates the workflow with SeedVR2-3B inference and supervised fine-tuning through VeOmni's existing DiTTrainer.

Related: #1030. This is a scoped implementation, not a claim that every issue acceptance criterion is complete: exact stochastic save/resume equivalence remains unresolved.

### Checklist Before Starting

- Related issue: #1030 (generic model migration skill and SeedVR2-3B example). The source is [ByteDance-Seed/SeedVR](https://github.com/ByteDance-Seed/SeedVR/tree/e4de8c24441a67e1b7df56abea10645059bb1185).
- PR title follows the required `[agent, model] feat: ...` format.
- Base: `ByteDance-Seed/VeOmni:main`; head: `ruchu127/VeOmni:feat/seedvr2-onboarding`. The 500-step experiment used `acb80ea8252b4f5d02b22c32ae34f0911a02cd9c`, incorporating upstream `b8a3edc`. Follow-up `4655ea6` fixes VAE checkpoint validation and documentation; the focused checks below were rerun after that fix.

### Test

**CI coverage:** extended the existing `tests/models/test_model_registry.py` with the SeedVR2 configuration/model case. This file is already enumerated in both `gpu_unit_tests.yml` and `npu_unit_tests.yml`. Added `tests/checkpoints/test_seedvr2_state_dict.py` under the directory already run wholesale by both workflows; no workflow change is necessary. The numerical scripts under `scripts/validation/` are manual reproducers. No full CI-suite pass is claimed.

Checks run after integrating Transformers 5.16.1:

- `make quality`: passed (Ruff check and formatting, 632 files).
- `pytest tests/checkpoints/test_seedvr2_state_dict.py tests/models/test_model_registry.py -k seedvr2 -q`: 19 passed. Covers missing/unexpected VAE convolution keys, non-strict mismatch reporting, complete state round trips, 2D-to-3D inflation and model registration. Before the fix, 9 rejection cases failed as expected.
- Complete pinned official VAE loaded successfully with strict validation after the fix (250,648,227 parameters).
- `python scripts/validation/validate_seedvr2.py`: finite forward/backward, an optimizer update, gradient checkpoint on/off agreement (maximum gradient difference `9.31e-10`), and exact tiny-model HF save/reload output agreement.
- Skill frontmatter validation: passed.

Earlier migration validation included strict loading of all 635 official DiT state tensors, loading the complete pinned VAE, exact converted-tensor equality, a fixed-noise full-weight NPU output check, and tiny source-reference comparisons. VAE missing/unexpected convolution-key rejection is now covered by the regression tests above. The reference comparison substitutes native attention/normalization and does not establish parity with original CUDA kernels; see `docs/examples/seedvr2.md` for the pinned revisions and tolerances.

#### Public-video, full-parameter training: 500 steps

Dataset follows the existing Wan example in `docs/examples/wan2.1_I2V_1.3B.md`: [Wild-Heart/Tom-and-Jerry-VideoGeneration-Dataset](https://huggingface.co/datasets/Wild-Heart/Tom-and-Jerry-VideoGeneration-Dataset), pinned revision `fc9066b00cb690f44779c5202572b8e052b5ced2`.

- Selected 512 videos after SHA256 deduplication; 46 duplicate files were excluded during initial selection and replacement. Sampling seeds: 1030 / 1031.
- Nine frames per video, temporal stride 2, seeded start frame, HQ resolution 128×192. LQ uses bicubic antialiased 2× downsampling and upsampling. Every pair has nonzero pixel MSE; all encoded latent shapes are `3×16×24×16`.
- Frozen official VAE encodes paired latents once; fresh noise is sampled during training. The 512-video subset is repeated across 2,000 sample presentations (500 global batches), not 2,000 unique videos.
- Fresh start from the converted official SeedVR2-3B checkpoint; all 32 NaDiT layers and **3,391,475,776 backbone parameters** trainable. VAE/fixed text conditioning frozen.
- Four Ascend NPUs; FSDP2, BF16 mixed precision, FP32 reduction, gradient checkpointing; global batch 4, micro batch 1; AdamW, constant LR `1e-5`, clipping threshold `1.0`, seed 666.
- **500/500 steps completed**, all recorded metrics finite. Training plus final checkpoint save took approximately 591 seconds (excluding download/preprocessing/startup). The step-500 DCP checkpoint was saved successfully. This timing is not a throughput benchmark.

| Metric | Step 1 | Step 500 | Last 20-step mean |
|---|---:|---:|---:|
| Loss (supervised velocity MSE) | 0.616222 | 0.044888 | 0.051933 |
| grad_norm (global L2, before clipping) | 7.179111 | 0.258007 | 0.259967 |

All 500 raw observations are shown without smoothing; the y-axes are logarithmic. Loss is averaged over four equal-sized rank batches. grad_norm is the FSDP global norm before clipping.

**Loss**

image

**Gradient norm**

image

**Limitations:** this is training-set optimization on a small cartoon-video subset with simple synthetic degradation, not held-out restoration quality, full-resolution convergence, or a reproduction of SeedVR2's unpublished adversarial post-training. No PSNR/SSIM/FVD or original-CUDA performance claim is made. Checkpoint writing and tiny-model reload passed, but fresh-process stochastic training resume did **not** reproduce the uninterrupted next-step loss; exact resume is not claimed. The inference CLI currently loads the original official weights, not the trained DCP checkpoint.

### API and Usage Example

```bash
# Convert official weights into the registered VeOmni model format.
python scripts/model_conversion/convert_seedvr2.py \
  --source /path/to/SeedVR2-3B --output /path/to/SeedVR2-3B-VeOmni

# Encode aligned LQ/HQ pairs (JSONL keys: video and target).
python scripts/model_conversion/prepare_seedvr2_pairs.py \
  --manifest pairs.jsonl --weights /path/to/SeedVR2-3B \
  --output output/seedvr2_pairs --device npu:0 --height 128 --width 192

# Run the existing trainer; provide enough samples/shards for all ranks.
torchrun --standalone --nproc-per-node=4 tasks/train_dit.py configs/dit/seedvr2.yaml \
  --model.model_path=/path/to/SeedVR2-3B-VeOmni \
  --model.config_path=/path/to/SeedVR2-3B-VeOmni/config.json \
  --model.condition_model_path=/path/to/SeedVR2-3B-VeOmni/condition \
  --data.train_path=output/seedvr2_pairs \
  --train.global_batch_size=4 --train.max_steps=500 \
  --train.checkpoint.save_steps=500
```

Source the installed CANN environment on Ascend first. See `docs/examples/seedvr2.md` for official weight checksums, conditioning paths, image/video inference, and supported boundaries. The public-video experiment used offline paired embeddings and a metric callback around the standard trainer; the generic commands above are not an automatic downloader/reproducer for that exact subset.

### Design & Code Changes

- One onboarding entry point: inspect actual source implementation to select the modeling route, then share configuration, data, training, checkpoint and validation phases. Existing Transformers patchgen responsibilities remain delegated to `/veomni-patchgen-model`.
- Add reusable migration references, checklist/report/config/conversion templates, and source-inspection tooling. Preserve upstream's requirement to ask before resolving checkpoint-key conflicts.
- Place all DiT integrations under `veomni/models/diffusers/<model_name>/` without requiring independent backbones to adopt Diffusers APIs.
- Add SeedVR2's NaDiT, separate frozen VAE conditioning, config/model registry entries, strict conversion and paired-data preparation, training config, inference example, registry test and validation scripts.
- Preserve the official DiT state layout with only the approved `dit.` prefix. Use native PyTorch attention/normalization in place of CUDA-only source dependencies, and make output modulation compatible with gradient checkpoint recomputation.
- Reuse the existing DiTTrainer and checkpoint manager. No SeedVR-specific training loop or shared-trainer RNG workaround is introduced. Sequence/context parallelism and 7B weights are outside the supported scope.
- Keep VAE convolution inflation compatible with strict missing/unexpected-key validation; add regression tests, correct the single-device FSDP2/meta recipe, and use English-only comments.

- Read the [Contribute Guide](https://github.com/ByteDance-Seed/VeOmni/blob/main/CONTRIBUTING.md).
- Applied `make quality` (Ruff lint/format checks). A full `pre-commit run --all-files` is not claimed.
- Added/updated model documentation, README and onboarding guidance.
- No existing `tasks/` training scripts were moved or renamed.
- Extended an existing CI-enumerated model-registry test; manual numerical and training checks are detailed above.
- Exact stochastic resume and held-out quality evaluation remain explicitly unverified/unsupported rather than marked as passing.

## Summary by CodeRabbit

* **New Features**
  * Added SeedVR2-3B support for image and video restoration.
  * Added supervised paired-training configuration and preprocessing tools.
  * Added checkpoint conversion with checksum and validation reporting.
  * Added model registration, inference support, and reusable video VAE components.
  * Added a generalized external-model migration workflow supporting Transformers, Diffusers, and other frameworks.

* **Documentation**
  * Added SeedVR2 usage, training, conversion, and validation guidance.
  * Added migration checklists, configuration templates, and integration report templates.
  * Updated supported-model and agent workflow documentation.

* **Tests**
  * Added SeedVR2 registry, checkpoint, numerical parity, training, and save/load validation coverage.


veomni/models/diffusers/seedvr2/core/nablocks/attention/mmattn.py
```diff
@@ -0,0 +1,260 @@
...
+                txt_k_repeat = rearrange(txt_k_repeat, "l (h d) -> l h d", h=num_h)
+
+                vid_q, vid_k, txt_q, txt_k = self.rope(
+                    vid_q, vid_k, window_shape, txt_q_repeat, txt_k_repeat, txt_shape_repeat, cache_win
+                )
+            else:
+                vid_q, vid_k = self.rope(vid_q, vid_k, window_shape, cache_win)
+
+        out = self.attn(
+            q=concat_win(vid_q, txt_q).bfloat16(),
```
FoolPlayer, line 240: [P2] RoPE has already repeated text Q/K per window, but `concat_win` indexes them using the original text lengths. In a packed batch, an earlier sample with multiple windows therefore makes later samples use its text Q/K. Reproduced with the toy model: the second sample differs from independent execution by ~0.0039. Please fix the indexing and add a packed-versus-independent regression test.

veomni/models/diffusers/seedvr2/inference.py
```diff
@@ -0,0 +1,92 @@
...
+    def __call__(self, frames, target_area=720 * 1280, seed=666):
+        """Restore T,C,H,W frames; offload VAE and DiT between stages."""
+        # Seed both the sampled VAE posterior and the subsequent DiT noise.
+        torch.manual_seed(seed)
+        video = preprocess_video(frames, target_area)
+        self.model.to("cpu")
+        self.condition.to(device=self.device, dtype=self.dtype)
+        latents = self.condition.encode_video(video)
+        text = self.condition.text_embedding
+        self.condition.to("cpu")
```
FoolPlayer, line 86: [P2] `.to("cpu")` leaves causal-convolution `memory` tensors on the accelerator because they are plain attributes. With a 9-frame tiny VAE, 16 CUDA caches remained after encode/offload and 41 after decode/offload, increasing memory pressure during DiT and subsequent calls. Please clear these caches before offloading, and provide a before/after test on the same clip (>5 frames) comparing live accelerator memory after each offload and peak memory, while confirming output equivalence.

veomni/models/diffusers/seedvr2/core/nablocks/attention/mmattn.py
ruchu127: `concat_win` used per-sample text lengths while q/k were already repeated per window, so later samples read the earlier sample's text (v was un-repeated as well). The interleave now uses per-window lengths and q/k/v each get one copy per window.

Toy config, window (2,2,2): new test `test_seedvr2_window_attention.py` went from 915/9216(9.9%) mismatched, max |Δ| 0.0535 to passing at atol/rtol 1e-3; packed DiT (3,8,8)+(1,4,4) was max |Δ| 0.0039184 (your ~0.0039), first sample bit-identical, now == per-sample; b=1 is bit-identical and `compare_seedvr2_source.py` compares per sample (0.0 / 7.45e-9).

veomni/models/diffusers/seedvr2/inference.py
ruchu127: `memory` is a plain attribute, so `to("cpu")` left it on the accelerator;`clear_causal_cache()` / `SeedVR2ConditionModel.clear_vae_cache()` now runs before both offloads.

Ascend 910, sliced tiny VAE (split_size=4, 9-frame 3x64x64): encode 14 caches = 1,038,400 B,decode 20 = 2,039,808 B; offloading with vs without clearing differs by 1,045,504 B / 2,050,048 B in both live and peak memory, and outputs stay equal (atol 2e-3; NPU jitters ~9.4e-4 across offload cycles). A no-op clear fails the new test, which is wired into both the GPU and NPU unit workflows.

PR #1189, opened 2026-09-15
[ckpt, trainer] fix: persist device and condition-model RNG across DiT resume

Feng0w0:
### What does this PR do?

Fixes a correctness hole in checkpoint resume for diffusion (DiT) trainers.

DiT condition models sample the noise and timestep ids used to build training targets. For Wan and Qwen-Image this comes from a per-run `torch.Generator`;for LTX2 / MiniMax-H3  it comes from the accelerator's default RNG.Neither is model state, and the job-level checkpoint (`GlobalStateCallback`) previously saved and restored only the CPU generator (`torch.get_rng_state()`).

On resume, the condition model is rebuilt fresh — and re-seeded — in a new process, so the resumed run **replayed the noise stream from the start of the run** at the first resumed step instead of continuing it. Training continued but was not bitwise-equivalent to an uninterrupted run. Verified by running the real `DiTTrainer` twice against a dummy dataset with dropout active (`p=0.5`):without the fix the resumed run diverges from the uninterrupted one at the first resumed step; with it the streams line up exactly.

This adds three pieces:

- the device-default RNG snapshot (`get_device_rng_state` / `set_device_rng_state`);
- a `rng_state_dict` / `load_rng_state_dict` protocol on the Wan and Qwen-Image condition models that own a `torch.Generator`;
- wiring in `GlobalStateCallback` to persist both across the resume, plus a rank-0 warning when a condition model owns a `generator` but does not expose `rng_state_dict` (so a future generator-backed model cannot diverge silently).

### Checklist Before Starting

- Related issues/PRs: none found — this closes a gap where the resume-path was never exercised (`tests/train_scripts/train_dit_test.py` sets the condition model to `None` and skips `process_condition` entirely).
- PR title follows `[{modules}] {type}: {description}`.

### Test

- Extended an existing CI-enumerated test: new cases in `tests/checkpoints/test_checkpoint_callback.py` (already listed in the `gpu_unit_tests.yml` / `npu_unit_tests.yml` `tests/checkpoints/` runs) cover the new `device_rng_state` / `condition_model_rng_state` fields, the noise-stream round trip, backward compatibility with checkpoints written before the
  fields existed, and the new warning.
- Added a new test file `tests/checkpoints/test_dit_resume_align.py` that runs the real `DiTTrainer` twice (uninterrupted vs. resumed from a mid-run checkpoint) under FSDP2 + DCP on the same dummy dataset and asserts the noise/timestep signatures match on the resumed steps. It lives in `tests/checkpoints/`, which both unit workflows enumerate wholesale, and skips on CPU. Dropout is kept on by default (`VEOMNI_ALIGN_DROPOUT_P=0.5`) to guard the ordering-symmetry invariant. Verified with 2 NPUs: **1 passed**.
- `make quality` passes for all changed files.

### API and Usage Example

No user-facing config changes. New optional protocol on diffusion condition models:

- `rng_state_dict() -> dict[str, torch.Tensor]`
- `load_rng_state_dict(state: dict[str, torch.Tensor]) -> None`

A condition model that draws noise/timesteps from a per-run `torch.Generator` should implement both so the stream is restored on resume. Models that sample from the accelerator's default RNG need no change — `device_rng_state` covers
them.

### Design & Code Changes

- `veomni/utils/device.py`: add `get_device_rng_state()` / `set_device_rng_state()`, accelerator-agnostic (CUDA / NPU / MLU via `get_torch_device()`), returning `None` / no-op where the device namespace does not expose the accessor.
- `veomni/trainer/callbacks/global_state_callback.py`: persist `device_rng_state` and `condition_model_rng_state` in the job-level checkpoint; restore both on load with backward compatibility (`.get`); warn if a `generator`-owning condition model exposes no `rng_state_dict`.
- `veomni/models/diffusers/{wan_t2v,qwen_image}/*_condition/modeling_*.py`: add `rng_state_dict` / `load_rng_state_dict` for the per-run generator.
- `veomni/trainer/dit_trainer.py`: expose the condition model on the reused `BaseTrainer` instance so `GlobalStateCallback` can reach it.

- Applied pre-commit checks
- Added/updated documentation
- Tests: extended an existing CI-enumerated test, **or** added a new test and wired it into `gpu_unit_tests.yml` (and `npu_unit_tests.yml` where applicable), **or** explained in **Test** why none is needed. Note that CI enumerates test files individually — a new file outside `tests/data/` and `tests/ops/` does not run until it is listed. See [.agents/knowledge/testing.md](.agents/knowledge/testing.md).

## Summary by CodeRabbit

- **New Features**
  - Checkpoint resumption now preserves CPU, accelerator, and supported condition-model random-number streams.
  - Supported DiT condition models can restore noise and timestep generation for consistent resumed runs.
  - Checkpoints remain compatible when device or model-specific random-state restoration is unavailable.
  - Additional checkpoint state is stored and restored alongside model data.

- **Tests**
  - Added distributed accelerator coverage verifying deterministic DiT checkpoint-resume behavior.

- **Documentation**
  - Updated testing guidance for resume and noise-stream equivalence.


veomni/trainer/callbacks/global_state_callback.py
```diff
@@ -145,6 +184,19 @@ def load_global_state(self) -> Optional[Dict[str, Any]]:
 
         self.trainer.environ_meter.load_state_dict(global_state["environ_meter"])
         torch.set_rng_state(global_state["torch_rng_state"])
+        # ``.get`` keeps checkpoints written before these fields existed loadable;
+        # the device namespace may also not support RNG state at all.
+        set_device_rng_state(global_state.get("device_rng_state"))
```
Coach257, line 293: Move model-related rng_state to `checkpoint_callback` and save in extra_states like:
├── model/[<module>/]
│   ├── ckpt/                     # weights, DCP
│   ├── optimizer/                # optimizer state, DCP
│   └── extra_state/          # lr scheduler & rng state

Feng0w0: The condition_model_rng_state has been moved to the checkpoint_manager.py

veomni/models/checkpoint_manager.py
```diff
@@ -136,44 +136,87 @@ def load_dir(self) -> Optional[str]:
...
         return self.config.load_path
 
+    def _extra_state(self, state: "TrainerState") -> Dict[str, Any]:
+        """Model-bound state to store beside the weights.
+
+        The condition model's noise/timestep generator and the lr scheduler are
+        model-bound state, persisted here for the checkpointer to write per rank.
+        """
+        lr_scheduler = self.trainer.lr_scheduler
+        condition_model = getattr(self.trainer, "condition_model", None)
```
Coach257: Use a more general way: 
`extra_state = self.trainer.extra_state()` and `return_dict.update(extra_state)` 
In `dit_trainer`: 
```
def extra_state():
    return self.condition_model.rng_state`
```
Move model-bound extra state to trainer function.

Coach257 (approved):

veomni/models/checkpoint_manager.py
Feng0w0: done

PR #1190, opened 2026-09-16
[parallel, ckpt] fix: register ExtraParallel specs for aliased parameter names

JorgenWan:
### What does this PR do?

Fixes silent expert-weight corruption on DCP resume for any model that registers one
ExtraParallel-sharded parameter under more than one name (a shared encoder/decoder trunk).

`ParallelPlan.apply` walks the deduplicated `model.named_parameters()`, so such a model gets a
`fqn2spec_info` keyed by whichever name came first. State dicts are *not* deduplicated — they carry
every name — and `_apply_extra_parallel_dim` silently skips any name it cannot find in that map. The
alias was therefore written to DCP with the ExtraParallel dimension still collapsed: an
`[E/ep, ...]` local shard recorded as if it were the whole `[E, ...]` tensor.

That is not a cosmetic duplicate key. Both names reach `set_model_state_dict` for a single parameter object, and their state-dict entries share storage, so the truncated alias is written over the same memory as the correctly assembled tensor. The checkpoint on disk is wrong regardless of how that resolves: with 8 experts at `ep=2` the alias records a `[4, 16, 32]` local shard as the whole `[8, 16, 32]` parameter.

### Checklist Before Starting

- Search for relative PRs/issues and link here: no existing issue found for aliased
  ExtraParallel parameters in DCP.
- PR title follows `[{modules}] {type}: {description}` format — `[parallel, ckpt] fix: ...`
- Not a breaking change: the fix only adds entries to `fqn2spec_info` for names that previously had
  none, so checkpoints of models without aliased parameters are byte-identical.

### Test

Added `tests/checkpoints/test_dcp_aliased_extra_parallel.py`. It is picked up by CI without a
workflow change, because `gpu_unit_tests.yml` and `npu_unit_tests.yml` both run
`pytest tests/checkpoints/` as a whole directory rather than enumerating files.

The test builds a toy model whose `self.encoder is self.decoder`, so one expert parameter is
reachable under two names, and runs it under `ep_size=2`. It asserts three things:

1. both the canonical and the aliased FQN are registered in `fqn2spec_info`;
2. the saved DCP metadata records the full global expert dimension for the aliased key, not the
   per-rank shard size;
3. expert values survive a save/load round-trip bit-for-bit.

Verified it fails on `main` and passes with this change:

```
$ pytest tests/checkpoints/test_dcp_aliased_extra_parallel.py -q
2 passed in 31.31s
```

Also ran locally: `ruff check` / `ruff format --check`, and
`python tests/special_sanity/check_device_api_usage.py -d tests -d veomni`.

Beyond the unit test, the bug was originally found on a 26B shared-trunk MoE model where the
exported HF checkpoint was unloadable because the aliased expert tensors had the per-rank shape;
after this change the export has the correct global shapes and loads.

### API and Usage Example

No API change. `ParallelPlan.apply` keeps its signature and return type; it only populates
`fqn2spec_info` more completely.

### Design & Code Changes

`veomni/distributed/parallel_plan.py` — after the existing per-parameter sharding loops, walk
`model.named_parameters(remove_duplicate=False)` and, for any name still missing from
`fqn2spec_info`, reuse the `spec_info` already resolved for that parameter object. A spec describes
the parameter's sharding, not the name it is reached by, so an alias is correct to reuse it. This
keeps the canonical name's spec authoritative and only fills in the names `named_parameters()` hid.

- Read the Contribute Guide
- Applied pre-commit checks (`ruff check`, `ruff format`)
- No documentation change needed: this is a correctness fix behind an existing API, with no new
  config field or workflow.
- Tests: added a new test file under `tests/checkpoints/`, a path CI already runs as a directory.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed checkpoint save and load behavior for models with shared parameters registered under multiple names.
  - Preserved full expert dimensions and parameter values during distributed checkpoint round-trips when ExtraParallel is enabled.

- **Tests**
  - Added coverage for aliased parameters across multiple ExtraParallel configurations and distributed ranks.


Coach257 (approved):

tests/checkpoints/test_dcp_aliased_extra_parallel.py
```diff
@@ -0,0 +1,181 @@
+"""DCP round-trip for a parameter that is registered under more than one name.
```
Coach257, line 1: Add this to ci pipeline

JorgenWan: it seems that it is already in the ci pipeline, as `uv run --frozen pytest -s -x tests/checkpoints/` runs all test in `tests/checkpoints/`.

PR #1193, opened 2026-09-16
[ckpt] feat: let the HF export plan exclude caller-named keys

JorgenWan:
### What does this PR do?

Adds an optional `drop_hf_keys` argument to `_get_sharding_plan`, letting the
caller name HF keys that should be left out of the export plan. Without the
argument nothing changes.

### Why

`_get_sharding_plan` reads the DCP metadata and plans one shard entry per key it
finds. That is all the information it has — the metadata records each name
independently, so two names that happen to reach the same tensor look exactly
like two separate tensors.

Tied weights are that case. `nn.Module.state_dict()` reports a shared tensor
under every name it is reachable by, and DCP does not deduplicate, so the
checkpoint holds one entry per name. An export that plans them all writes the
same bytes more than once, and `model.safetensors.index.json` then advertises the
inflated total. The plan itself cannot detect this; only the caller, which holds
the model, knows which names alias which.

Why a drop-list rather than a keep-list: the DCP key space and the model's key
space are only conventionally aligned. A wrong `--model-assets-dir`, or a
different wrapper depth, would make a keep-list silently drop real weights. An
over-long drop-list at worst exports a tensor that was meant to be pruned.

### Scope

There is no in-tree caller yet — the alias computation is model-specific (it
depends on how a given model ties its parameters) and belongs with whichever
exporter needs it. This PR contributes only the mechanism; `scripts/merge_dcp_to_hf.py`
is untouched.

### Tests

`tests/checkpoints/test_dcp_checkpointer.py::TestShardingPlanDropHfKeys` — three
single-process CPU cases over a two-tensor DCP checkpoint keyed the way a tied
checkpoint holds it:

- the named key is absent from the plan
- `total_size` follows the plan, so the index does not advertise bytes nobody wrote
- a key the checkpoint does not have removes nothing

Each fails on `main` with `TypeError: _get_sharding_plan() got an unexpected
keyword argument 'drop_hf_keys'`.

### Checks

- `pytest tests/checkpoints/ -q` → 12 failed, 169 passed, 2 xfailed. The 12
  failures are all in `test_trainer_saveload.py` and reproduce identically on a
  clean `upstream/main` worktree (verified commit-by-commit, same 12 ids).
- `ruff check` / `ruff format --check` clean on both changed files.
- `python tests/special_sanity/check_device_api_usage.py -d tasks -d veomni -d tests` passes.

### Checklist

- [x] Read the [Contributing Guide](../blob/main/CONTRIBUTING.md).
- [x] Tests added for the new behavior.
- [x] No API breakage — the new argument defaults to `None`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Checkpoint sharding now correctly excludes specified Hugging Face–normalized duplicate keys, preventing them from affecting shard contents and reported sizes.
  * Unmatched keys continue to leave sharding results unchanged.

* **Tests**
  * Added coverage for excluding duplicate normalized keys and preserving behavior when no matching keys are found.


scripts/merge_dcp_to_hf.py
```diff
@@ -40,6 +40,55 @@ def _detect_lora(all_hf_keys: Sequence[str]) -> bool:
     return any(_is_lora_key(k) for k in all_hf_keys)
 
 
+def _aliased_hf_keys(model_assets_dir: str) -> Optional[set[str]]:
```
Coach257: No change on this script.

JorgenWan: OK

Luosuu (approved):

Issue #1198, opened 2026-09-16
RFC: Diffusion Remove-Padding with MiniMax H3 as the First Consumer


NancyFyong:
## Feature request

Propose a DiT-wide, opt-in `model.use_remove_padding` interface for fixed-microbatch cross-sample packing, with native **MiniMax H3 FL2VA offline training** as the first working example. The full RFC follows below.

**This is a shared capability interface, not automatic support for every model.** Only semantically invalid padding may be removed; valid zero-valued latents, conditioning/reference tokens, and learned register tokens must be preserved. Each model/backend/parallelism combination needs explicit support and validation.

## Motivation

VeOmni already supports LLM/VLM packing, but the DiT path bypasses the common collator and forces single-sample microbatches. H3's core accepts packed boundaries, yet its condition preparation and prediction/loss reconstruction remain single-sample. The proposal connects these pieces without changing sample weighting or importing causal-LM assumptions into diffusion.

## Related work / duplicate check

Open issues and PRs were searched for remove-padding, diffusion packing, packing, and MiniMax H3. No open proposal implementing this same DiT/H3 capability was found.

- [#212](https://github.com/ByteDance-Seed/VeOmni/issues/212), closed: generic packing enable/disable usage, not native DiT multi-sample execution.
- [#402](https://github.com/ByteDance-Seed/VeOmni/issues/402), closed: padding packed inputs for compiler shape stability, not removing per-sample diffusion padding.
- [#1189](https://github.com/ByteDance-Seed/VeOmni/pull/1189), open at inspection: DiT device/condition RNG checkpoint restoration. Coordinate future save/resume tests with this work rather than duplicate the RNG fix.
- Related downstream work is distinguished in Section 11 below.

## Your contribution

This issue contributes a source-grounded design, H3 integration plan, and proposed numerical/performance acceptance criteria. No implementation or model/GPU validation is being submitted, and no speedup is claimed. AI assistance was used for source inspection and drafting; maintainer feedback is requested before follow-up implementation.

---

**Status:** Draft for discussion; no implementation or performance claim.
**Inspected baseline:** VeOmni `aa8bb43c6ba457afde4b9bd35cceaf74823604cc` (main).
**Scope:** Native VeOmni DiT training and model forward, not a verl-omni engine patch.

## 1. Summary and proposed decisions

Introduce a DiT-wide, opt-in `model.use_remove_padding: false` execution setting.
[...]

PR #1201, opened 2026-09-16
[trainer] fix: honor DiT microbatch size

NancyFyong:
### What does this PR do?

Honors configured DiT `train.micro_batch_size` while preserving fixed-sample batching and the existing condition/model interfaces. Following @Coach257's review, this replaces the previously proposed shared remove-padding protocol, rather than adding a second Trainer API.

Related: RFC #1198 and the separate MiniMax H3 consumer #1204. This PR does not implement packing or close the RFC. Models remain responsible for batching/packing and unsupported configurations.

### Checklist Before Starting

- Searched open PRs for `DiT microbatch` and `1198 in:body`, and checked RFC #1198. The related work is this PR and #1204; this updates the existing PR rather than creating a duplicate.
- PR title follows the `[{modules}] {type}: {description}` format.

### Test

**Added a new test and wired it into CI:** `tests/trainer/test_dit_microbatch.py` replaces the API-only suite and is explicitly registered in both `gpu_unit_tests.yml` and `npu_unit_tests.yml`. The new tests use CPU fixtures, not accelerator execution.

The initial 24-case regression suite reported **18 failed, 6 passed** before the microbatch fix. The final **30 cases pass**, covering:

- `M=1/2/4`, accumulation `K`, and mocked DP sizes 1/2 through the production dataloader/collator.
- Offline/online `DiTTrainer.train_step` with unequal sample lengths, runtime-root forwards, real CPU AdamW updates/moments and LR scheduling against an equal-sample reference. Distributed lifecycle hooks are mocked.
- Embedding with `M=1/2/4`, `K=1`, and one parquet row per sample.
- Non-divisible embedding datasets through the real `MappingDataset`: all mocked DP ranks together save each original ID exactly once. This reproduced five failures before fixing padding, then three duplicate/missing-ID failures before forwarding `shuffle=False`.

Expanded validation:

```bash
CUDA_VISIBLE_DEVICES='' PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH="$PWD" \
  python -m pytest -q -ra \
  tests/trainer/test_dit_microbatch.py \
  tests/trainer/test_step_sync.py \
  tests/trainer/test_channel_loss_callback.py \
  tests/models/test_model_runtime.py \
  tests/models/test_runtime_chat_template.py \
  tests/trainer/test_aux_metrics.py \
  tests/checkpoints/test_checkpoint_callback.py
```

**185 passed, 5 skipped.** Skips require an accelerator for FSDP2 meta initialization. `make quality`, changed-file pre-commit, agent-doc paths, doc task paths, device API checks and `git diff --check` pass.

Environment: Python 3.11, Torch 2.11.0+cu130, Transformers 5.16.1 and Diffusers 0.37.0; not a complete frozen-lockfile installation. **No GPU/NPU, real distributed optimizer, convergence or performance validation is claimed.** Historical H3 tests using the removed protocol do not validate this revision.

### API and Usage Example

No new public API or packing switch. For a model supporting `M=2`, with DP size `D=4`, this uses `K=2`:

```yaml
train:
  training_task: offline_training
  dyn_bsz: false
  micro_batch_size: 2
  global_batch_size: 16
```

Training keeps `G = M × D × K`. Embedding overrides `G = M × D`, keeping `K=1` while preserving `M`. Existing recipe settings remain `M=1`.

### Design & Code Changes

- Stop forcing `micro_batch_size=1`; retain `dyn_bsz=false` and `dataloader_batch_size=G/D`.
- Keep embedding to one microbatch per step without forcing one sample. Forward `data.shuffle` and pad an index view without overwriting `MappingDataset`'s source length, preventing padded repeats from displacing real samples in saved embeddings.
- Remove the proposed `use_remove_padding`, `prepare_samples`, `DiffusionSample`, `DiffusionBatchOutput`, `sample_inputs` protocol and Trainer capability/execution checks.
- Preserve `get_condition? → process_condition(**batch) → model(**batch) → outputs.loss / K`, the dict-of-lists collator and sample-mean scalar loss contract. Models own packing, multimodal layouts, SP/CP and capability rejection.
- Replace interface documentation with `docs/usage/dit_microbatching.md`; update architecture guidance and the stale LTX recipe comment.

#1204 still needs migration to the existing condition/forward contract; its H3 modeling and 64-row-alignment review changes are separate and are not claimed complete here.

AI assistance was used for implementation, validation and drafting. Independent static review identified the embedding shuffle/padding issue; after the fix and regression, follow-up review returned **safe** with no findings. This is not accelerator validation.

- Read the Contribute Guide.
- Applied pre-commit checks and `make quality`.
- Updated documentation.
- No task scripts moved or renamed; doc task-path validation passed.
- Added and registered the new test in both GPU and NPU unit workflows; scope and limitations are listed above.

## Summary by CodeRabbit

- **New Features**
  - DiT training now honors the configured fixed micro-batch size while keeping dynamic batching disabled.
  - Offline embedding supports multiple samples per microbatch and preserves each original sample exactly once, including incomplete final batches.
  - Batch sizing, gradient accumulation, and loss handling are aligned with fixed microbatch training.

- **Documentation**
  - Added guidance for DiT microbatching, offline embedding behavior, batching, padding, and model-owned packing.
  - Updated usage navigation and configuration notes to reflect the new behavior.

- **Tests**
  - Added coverage for fixed microbatch training and offline embedding workflows.


NancyFyong:
hi @Coach257 , can you PTAL when you have moment? THX!

Coach257 (changes requested):
## Direction

The goal is right: a DiT microbatch should be allowed to contain `M > 1` samples, and a supporting model should be free to pack valid tokens in its own forward. The shared trainer protocol in this PR is the wrong place to put that.

DiT already collates a `dict[str, list]` of length `M`. The only reason every microbatch is one sample is this override in `DiTTrainer._setup`:

```python
args.train.micro_batch_size = 1
```

Packing, multimodal layout, and SP/CP are modeling problems. They should not grow a second trainer API.

This is also **not** `train.dyn_bsz = true`. Dynamic batching is a dataloader token-budget scheduler. This work is fixed `M`, `dyn_bsz=false`. Keep those independent.

## Do not do this

Please drop the trainer-level packing protocol from this PR:

- `model.use_remove_padding`
- `prepare_samples` vs `process_condition`
- `DiffusionSample` / `DiffusionBatchOutput` / `sample_inputs=`
- capability checks and execution envelopes in `DiTTrainer` / `packing.py` (FSDP2-only, SP/CP/TP/PP = 1, no LoRA, no compile)

Those pull model-owned layout, attention boundaries, and parallelism into the framework.

Unlike LLM packing, DiT must **not** slice SP/CP in the collator. `DiTDataCollator` stays a dict-of-lists. Ulysses already runs after patchify inside modeling (`slice_input_tensor` in Wan / Qwen-Image). Packed `cu_seqlens` after all-to-all belongs in the same modeling path. A model that cannot pack under SP should raise in modeling; the trainer does not need an envelope flag for it.

Wan / Qwen-Image / LTX already iterate the list today. Lifting `M` there without a pack implementation would silently become `M` serial DiT forwards, not remove-padding. That is a modeling choice (loop, concat, or raise). It is not a reason to add `use_remove_padding`.

## Do this instead

Keep:

- `dyn_bsz = False`
- `dataloader_batch_size = global_batch_size / dp_size`  (`= M × K`)

Delete the `micro_batch_size = 1` override. Then the existing formula already holds:

```text
G = M × dp_size × K
```

`MakeMicroBatchCollator` already splits one `next()` into `K` microbatches of `M` samples. The trainer path stays:

```text
get_condition? → process_condition(**batch) → model(**batch) → outputs.loss / K
```

Modeling owns everything after that list:

- serial for-loop (current Wan / Qwen / LTX)
- concat valid rows + segmented / varlen attention (H3, next PR)
- raise if `M > 1` is unsupported (H3 `process_condition` already does)

Keep the existing loss contract: `outputs.loss` is a **sample-mean** scalar dict, then the trainer divides by `K`. Do not add `[B]` loss tensors or a second output type. Averaging packed tokens instead of samples is a modeling bug (long videos would dominate).

## Tasks

No new task forks. Packing is orthogonal to `training_task`.

| Task | After lifting `M` |
| --- | --- |
| `offline_training` | `process_condition` + DiT see `M` samples. This is where H3 packing lands later. |
| `online_training` | `get_condition` already loops the list, then the same `process_condition` / forward path. |
| `offline_embedding` | Never packs. Condition for-loops and returns a dict-of-lists; the saver already `_unpack_dict_of_list` and writes one parquet row per sample. **`M > 1` can be saved.** |

The embedding `M = 1` lock is not a packing constraint. The comment is about `broadcast_object_list` OOMing when **`K > 1`** serializes many video microbatches at once. Embedding has no optimizer, so keep **`K = 1`**:

```text
G = M × dp_size
dataloader_batch_size = M
```

Do not keep forcing `M = 1` for embedding. Do keep `K = 1`. After removing the global override, the embedding block should still set `micro_batch_size` consistently with that `G` (otherwise `num_micro_batch = G / (M × dp)` can become 0 if someone reuses a training yaml with `M > 1`).

## H3 packing

How H3 concatenates `[text | cond | audio | video]`, drops the 64-align pad, isolates samples in attention, remaps RoPE / timesteps, and reduces per-sample video/audio MSE belongs in the consumer PR (#1204), inside H3 modeling. This PR should not invent `sample_inputs=` for it.

## Docs

Please replace the new interface guide with:

- DiT `micro_batch_size` is now honored
- input remains dict-of-lists
- packing / SP / CP stay model-owned
- existing recipes with `M = 1` are unchanged

## Suggested shape of this PR

Drop the new API, the tests that only exist to lock that API, and the trainer branches. Land the `_setup` change plus docs. Keep default yaml at `M = 1` so current jobs do not change. Put the first packed consumer in the H3 PR.

NancyFyong:
@Coach257 Thanks for the direction — the #1201 revision is pushed in f08c63d0e397b7afe3e9eee801acb8c89bbf7422.

- Removed the proposed `use_remove_padding` / `prepare_samples` / `DiffusionSample` / `DiffusionBatchOutput` / `sample_inputs` protocol and Trainer capability envelopes.
- `DiTTrainer` now honors configured `M`, retains `dyn_bsz=False`, dict-of-lists collation, `G = M × dp_size × K`, and the existing condition → model → sample-mean `outputs.loss / K` path.
- Offline embedding preserves `M` while enforcing `K=1`. It now forwards `shuffle=False` and pads an index view without changing `MappingDataset`'s source length; the non-divisible-dataset regression verifies that all mocked DP ranks together save each original sample exactly once.
- Replaced API-only documentation/tests with the microbatch guide and regressions registered in both unit workflows. Updated this PR's title and description accordingly.

Validation: **185 CPU tests passed, 5 accelerator-only tests skipped**; quality, pre-commit and diff checks passed. The tests include real CPU AdamW state/update and scheduler checks, but distributed lifecycle/topology are mocked; no GPU/NPU or distributed-training result is claimed. Follow-up independent static review returned safe.

Packing/SP/CP remain model-owned. **#1204 has not been updated in this push**: its migration to the ordinary condition/forward contract and removal of the sample-layout 64-row alignment are still separate follow-up work, not claimed complete here.

AI assistance was used for this revision and validation. Could you please take another look at the revised #1201 scope?

Coach257 (approved):

PR #1204, opened 2026-09-18
[model, trainer] feat: pack MiniMax H3 offline training

NancyFyong:
### What does this PR do?

Pack **MiniMax H3 FL2VA and visual Ref2VA offline-training microbatches inside modeling**, using the existing `process_condition(**batch) → model(**batch) → outputs.loss / K` contract. Following @Coach257's reviews on #1201 and this PR, there is **no shared remove-padding API or `model.use_remove_padding` switch**.

**Depends on #1201 at `f08c63d0`; related RFC #1198.** This branch merges the revised fixed-microbatch Trainer change. Relative to #1201, the consumer changes are H3 modeling, its configuration/documentation and registered regressions, not another Trainer implementation. If #1201 is squash-merged or advances, reconcile this branch with main before merging.

`M=1` retains the ordinary single-sample attention path; `M>1` concatenates samples for one main-DiT forward. The small text refiners remain sample-local to avoid amplified BF16 projection rounding. **Upstream modulation-gather arithmetic is unchanged.**

FL2VA and Ref2VA layouts now use **`seq_len=used`, `cu_seqlens=[0, used]`**, without a 64-row tail. Cross-sample packing concatenates complete samples rather than cropping a previously added tail. Single-sample Ulysses divisibility padding remains inside `MiniMaxH3DiT.forward`.

### Checklist Before Starting

- Checked open PRs for MiniMax H3 packing and the related RFC. #1201 owns the Trainer prerequisite, #1204 owns this consumer; #1142 is separate kernel-registry work. Related historical resume issue: #1189.
- PR title follows the `model, trainer` / `feat` convention. No performance certification is implied by the title.

#### Current revision

Submitted head: `69d7fffc` (merges #1201's `f08c63d0`).

**New tests explicitly registered by this PR; this revision updates their existing branch versions.** The model test is listed in both GPU/NPU unit workflows; the two-rank FSDP2 test is listed in the GPU workflow. #1201's replacement microbatch test is retained in both workflows, and the deleted shared-protocol test is not restored.

```bash
CUDA_VISIBLE_DEVICES='' PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH="$PWD" \
  python -m pytest -q -ra \
  tests/models/test_minimax_h3_remove_padding.py \
  tests/trainer/test_minimax_h3_remove_padding_fsdp.py \
  tests/trainer/test_dit_microbatch.py \
  tests/trainer/test_step_sync.py \
  tests/trainer/test_channel_loss_callback.py \
  tests/models/test_model_runtime.py \
  tests/models/test_runtime_chat_template.py \
  tests/trainer/test_aux_metrics.py \
  tests/checkpoints/test_checkpoint_callback.py
```

**219 passed, 13 skipped.** Eight skips require two CUDA GPUs; five require an accelerator for FSDP2 meta initialization. `make quality`, changed-file pre-commit, doc/agent path checks, device API checks and diff checks pass.

The 34 native-model CPU cases cover:

- Ordinary condition/model interfaces, registry-based construction, per-sample conditioning RNG, scalar sample-mean losses, outputs and gradients versus serial execution; explicit scalar/list checkpoint-offload rejection for multi-sample inputs and forwarding through both the single-sample wrapper and inference path.
- Native H3 through `DiTTrainer.train_step`, two accumulation microbatches, real CPU SGD updates and scheduler advancement; distributed lifecycle/topology are mocked.
- Variable prompt/reference layouts, target-only loss, sample isolation, valid zero rows, independent DiT/refiner boundaries and backward checkpoint recomputation.
- Single-sample valid-token outputs/losses versus a legacy 64-row-padded input; forward-local SP padding with **mocked collectives**, not distributed SP parity. A poisoned-allocation regression reproduces nonfinite outputs from uncovered SP attention rows; zero-initializing those rows preserves finite parameter gradients.
- FL2VA/Ref2VA training/inference layout consistency, no-audio-reference rejection, stale padded-metadata rejection and upstream modulation-gather dtype/gradient preservation.
- CPU-stub FA2/FA3 dispatch through the foundation loader, preserving the ordinary single-sample path while packing main-DiT attention and keeping refiners sample-local. Stubs are **not real-kernel validation**.

Environment: Python 3.11, Torch 2.11.0+cu130, Transformers 5.16.1, Diffusers 0.37.0; not a complete frozen-lockfile installation. **No GPU/NPU runs were performed for this revision:** permitted GPUs were occupied and existing training was not interrupted. The migrated two-rank FSDP2 regression is collected but remains unexecuted on the revised source.

#### Historical numerical and performance evidence — not validation of this revision

The following results are retained for provenance, not presented as fresh results for the ordinary-interface/no-tail migration.

**Earlier upstream-gather consumer:** official FL2VA/Ref2VA weights (50 blocks, 33,122,992,896 parameters), synthetic precomputed inputs, four H20s, BF16/FA3/FSDP2, batch 2/rank. Eighteen serial/serial/packed triplets per task used fixed inputs, restored RNG and six call orders:

| Task | Serial-repeat gradient relative-L2, mean | Serial-vs-packed gradient relative-L2, mean | Maximum output relative-L2 |
|---|---:|---:|---:|
| FL2VA | 2.286% | 2.333% | 2.943e-5 |
| visual Ref2VA | 2.530% | 2.444% | 3.279e-5 |

All 36 serial-vs-packed comparisons stayed below the original 5% gradient gate. This showed no clear excess over serial repeat variation on those fixtures, **not equivalence or zero-diff**; there was no packed self-repeat baseline. Historical two-rank SDPA/FA2/FA3 tests also passed all eight cases before this migration. Neither result certifies the new layout or interface.

**Earlier FP32-gather experimental variant:** the following timing/Trainer runs additionally applied an FP32 modulation-gather correction to both serial and packed paths. That correction was removed from this PR. Full official weights, eight H20s, synthetic offline inputs, BF16/FA3/FSDP2 and gradient checkpointing:

| Task | Batch/rank | Serial seconds | Packed seconds | Serial / packed | Peak allocated GiB, serial / packed |
|---|---:|---:|---:|---:|---:|
| FL2VA | 1 | 3.741 | 3.679 | 1.017x | 37.842 / 37.847 |
| Ref2VA | 1 | 3.815 | 3.825 | 0.997x | 37.842 / 37.847 |
| FL2VA | 2 | 7.367 | 6.589 | 1.118x | 39.808 / 38.393 |
| Ref2VA | 2 | 7.589 | 6.780 | 1.119x | 39.842 / 38.430 |
| FL2VA | 4 | 14.764 | 12.269 | 1.203x | 39.819 / 40.196 |
| Ref2VA | 4 | 15.260 | 12.568 | 1.214x | 39.853 / 40.273 |
| FL2VA | 8 | 29.734 | 23.564 | 1.262x | 39.836 / 43.815 |
| Ref2VA | 8 | 30.400 | 24.261 | 1.253x | 39.870 / 43.959 |

These are warmed preparation/forward/backward/FSDP measurements, **excluding optimizer, encoders and rollout**, against per-sample serial backward, not tuned dense batching. Gains include fewer FSDP/forward/backward calls. Serial preceded packed; shared-host residency observations do not establish exclusive GPU use. Fixed/variable-length cases measured about 1.12x at batch 2, 1.20–1.23x at batch 4 and 1.25–1.27x at batch 8, with higher packed peak memory at larger batches.

The same historical variant completed standard `tasks/train_dit.py` fresh/resume checks for both tasks, packing off/on, with real full-parameter AdamW state and scheduler audits. Three-step first-step observations were 19.91→18.19 seconds for FL2VA and 20.05→18.21 for Ref2VA, **not warmed stable Trainer throughput**. CUDA conditioning RNG was not restored by the tested base (related #1189), and linear scheduling did not honor the requested `lr_min`; neither issue is fixed here. These runs do not establish current-source Trainer E2E, RNG-correct resume, real-data convergence, RL behavior or optimizer-update parity.

Current-source official-weight correctness and warmed standard-Trainer batch 1/2/4 timing/memory comparisons remain follow-up work. No checkpoints or generated data are committed.

### API and Usage Example

For matching offline FL2VA embeddings and current layout metadata:

```bash
bash train.sh tasks/train_dit.py configs/dit/minimax_h3_fl2va_offline.yaml \
  --train.micro_batch_size 2 \
  --train.global_batch_size 16
```

Keep `dyn_bsz=false`, `data.dataloader.drop_last=true`, and `model.accelerator.fsdp_config.mixed_precision.cast_forward_inputs=false`. The recipe still defaults to `M=1`. Select eager/SDPA or compatible local FA2/FA3 via `model.ops_implementation.attn_implementation`.

**Cached-data compatibility:** regenerate old `packed` metadata with the updated builders before `M>1` training; latent tensors and embeddings need not be re-encoded. Multi-sample packing rejects old tail-padded metadata rather than silently cropping it.

Visual Ref2VA requires matching weights, reference-aware precomputed embeddings, `build_packed_ref2va(...)` metadata and `ref_visual_anchor` rows in block order. The FL2VA embedding recipe is not a Ref2VA encoder. Reference audio is unsupported.

### Design & Code Changes

- Merge #1201's replacement microbatch design without restoring its removed protocol or modifying its Trainer implementation.
- Independently prepare each sample through `process_condition`, preserving RNG order and anchor validation. Multi-sample results are ordinary dict-of-lists fields; existing single-sample dense inputs remain accepted.
- Concatenate complete model inputs, rebase scatter/timestep indices, retain sample-local RoPE and separate main/refiner boundaries, then return ordinary `MiniMaxH3DiTOutput` with scalar sample-mean video/audio losses.
- Remove 64-row sample-layout alignment in FL2VA/Ref2VA training and inference. Keep single-sample forward-local Ulysses padding and zero uncovered attention rows to prevent uninitialized values contaminating gradients; multi-sample Ulysses and block/checkpoint offload fail in modeling. Other unvalidated parallel/LoRA/compile combinations are not claimed supported.
- Preserve upstream modulation gathers, parameter names, FSDP block hierarchy and the single-sample attention path. Cross-sample refiners remain sample-local.
- Update model docs/config and the existing CPU/FSDP tests; preserve both #1201 and H3 CI registration.

AI assistance was used for implementation, validation and drafting. Independent full-branch review identified offload forwarding and stale-documentation issues; these were fixed with regressions, and final static review returned **safe** with no remaining concrete findings. This is not accelerator validation; hardware and historical-source limitations are stated above.

- Read the Contribute Guide.
- Applied `make quality` and pre-commit checks.
- Updated docs, configuration and architecture guidance.
- No task scripts moved; doc task-path validation passed.
- Registered the new CPU/FSDP tests and updated their cases; GPU/NPU execution of this revision remains pending.

## Summary by CodeRabbit

- **New Features**
  - MiniMax H3 now supports multi-sample microbatches with model-managed packing and sample-mean losses.
  - DiT training honors configured microbatch sizes instead of forcing single-sample batches.
  - Packed FL2VA and visual Ref2VA sequences now use exact token lengths without alignment padding.

- **Bug Fixes**
  - Improved handling of sequence-parallel padding by ensuring uncovered attention rows are initialized safely.
  - Preserved sample-local processing for refiner operations and single-sample execution.

- **Documentation**
  - Added guidance on DiT microbatching, H3 packing requirements, supported configurations, and validation behavior.


configs/dit/minimax_h3_fl2va_offline.yaml
```diff
@@ -15,6 +15,8 @@
 # Model
 # -----------------------------------------------------------------------------
 model:
+  # Opt in with micro_batch_size > 1; retain cast_forward_inputs=false below.
```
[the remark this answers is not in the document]
NancyFyong: Fixed in 12795214. The comment now states that micro_batch_size > 1 provides the packing benefit; micro_batch_size=1 remains supported. Targeted CPU coverage and quality checks pass.

tests/trainer/test_minimax_h3_remove_padding_fsdp.py
```diff
@@ -0,0 +1,451 @@
...
+            recorder = _StepRecorder()
+            base._callbacks = [recorder]
+            base.optimizer = torch.optim.SGD(model.parameters(), lr=_LR, foreach=False)
+            base.lr_scheduler = torch.optim.lr_scheduler.LambdaLR(base.optimizer, lambda _: 1.0)
+            scheduler_epoch = base.lr_scheduler.last_epoch
+            trainer = DiTTrainer.__new__(DiTTrainer)
+            trainer.base = base
+            trainer.training_task = "offline_training"
+            trainer.condition_model = condition_model()
+            trainer.condition_model.requires_grad_(False)
```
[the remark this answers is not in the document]
NancyFyong: Fixed in 12795214. The regression now constructs a DiTModelRuntime and stores the condition model, optimizer, and scheduler on that runtime, then passes the runtime to BaseTrainer._build_training_context. This is slightly broader than assigning onto the raw FSDP root: attaching the condition model to the raw nn.Module would register extra parameters and break the later parameter/update oracle. The review-fix CPU subset passes (102 passed; 8 GPU-only cases skipped), quality/hooks pass, and an independent review returned safe.

Coach257 (changes requested):
## 64-align pad should not exist in the sample layout

H3 already does **intra-sample** packing in `build_packed_fl2va`: one sequence is `[text | cond | audio | video]`. That layout belongs in modeling. The trailing pad to a multiple of 64 does not.

It is applied in the layout builder, then inherited by embedding and training:

```python
# veomni/models/diffusers/minimax_h3/minimax_h3_core/packed_sequence.py
seq_len = ((used + 64 - 1) // 64) * 64
cu_seqlens = [0, used, seq_len]
```

`get_condition` only records that table; it has no reason of its own to pad. `process_condition` then allocates `x` / `audio_x` as `[1, seq_len, C]` of zeros and scatters real rows into it. Inference Ref2VA copies the same `_aligned_seq_len`.

This is **not** batch padding (`bs` is already 1). It is not required by H3 modeling: the DiT uses segmented SDPA / varlen FA, and nothing checks `S % 64 == 0`. Pad tokens sit in a **separate** attention segment `[used, seq_len)`, so they do not mix with real tokens, but QKV / FFN / AdaLN still run on them.

If a kernel needs a aligned length, pad/unpad **around that kernel**:

- Ulysses already does this in `MiniMaxH3DiT.forward` (`pad` so `S` splits across `sp_world`, `cu_bounds` stay at the real length, extra rows dropped). Keep that.
- Do not bake 64-align into the saved `packed` metadata or the default input sequence.

## What this PR should not do

Do not add a packing-time unpad on top of that tail:

```python
used = int(inp["packed_seq_params"]["cu_seqlens_q"][1])
chunks[key].append(inp[key][:, :used])
```

That crop only exists because the builder padded first. It is extra code fighting the layout.

## What to do instead

1. In `build_packed_fl2va` (and the Ref2VA copy in `inference.py`): `seq_len = used`, `cu_seqlens = [0, used]`. Drop `_SEQ_ALIGN`.
2. Cross-sample packing then **only concatenates** valid rows:

```text
x = cat(x0, x1, dim=1)
cu_seqlens = [0, L0, L0+L1]
prompt = cat(p0, p1, dim=0)
scatter indices += prefix
RoPE / token_tags concat as-is (do not rebase as one video)
```

3. Leave SP pad where it is (forward, `sp_world > 1` only).

`cu_seqlens` becomes length 2 for a single sample (one segment) and length `M+1` when packing `M` samples. Update anything that assumes the triple `[0, used, padded]`.

Single-sample `M=1` with no 64 tail should match the current valid-token outputs: the pad segment never attended into real rows, and the loss already gathers only real positions. Please keep a regression for that, plus `sp>1` still padding inside the DiT forward.

This also simplifies #1201-style trainer protocol discussion: packing stays concat in H3 modeling, with no layout-level pad and no extra unpad helper.

NancyFyong:
@Coach257 The H3 consumer has now been revised in 69d7fffcd04e2aa8f2a13c558c75acd204006100 to follow your direction on both #1201 and #1204:

- Incorporated #1201's `f08c63d0` fixed-microbatch change. Removed the shared `use_remove_padding` / `prepare_samples` / `sample_inputs` protocol and alternate output type; H3 uses ordinary `process_condition(**batch)` and `model(**batch)` with scalar sample-mean losses.
- Removed `_SEQ_ALIGN` and the 64-row tail from FL2VA and Ref2VA training/inference layouts: `seq_len=used`, `cu_seqlens=[0, used]`. Cross-sample packing concatenates complete samples without cropping a pre-added tail, retaining separate attention boundaries and sample-local positions/timesteps.
- Preserved single-sample Ulysses padding inside the DiT forward. Added legacy-tail valid-output equivalence and forward-local SP-padding regressions. A poisoned-allocation regression also reproduced nonfinite parameter gradients from uncovered attention rows; those rows are now zero-initialized. Upstream modulation-gather arithmetic is unchanged.
- Added explicit checkpoint-offload forwarding/rejection through conditioning, wrapper and inference so kwargs cannot silently bypass the multi-sample guard. Migrated the CPU/FSDP2 tests to the ordinary interface and preserved workflow registration. Updated the configuration, documentation and PR description, including the requirement to regenerate old cached layout metadata before multi-sample training.

Validation: **219 passed, 13 skipped** on CPU; quality, pre-commit and path/device/diff checks pass. This includes native H3 through the Trainer's accumulated CPU optimizer step. SP collectives and distributed lifecycle are mocked in CPU cases. The eight two-GPU FSDP2 cases and five accelerator-only runtime cases were skipped: permitted GPUs were occupied, and no training was stopped.

The description now clearly labels prior official-weight, kernel and performance measurements as **historical results from earlier source variants**, not validation of this interface/layout revision. Fresh GPU/FSDP2 and official-weight validation remain pending.

veomni/models/diffusers/minimax_h3/minimax_h3_transformer/modeling_minimax_h3_transformer.py
```diff
@@ -61,6 +63,52 @@ def __init__(self, config: MiniMaxH3DiTModelConfig, **kwargs):
...
+        if (
+            self.config.latents_dim != 24
+            or self.config.audio_latents_dim != 32
+            or tuple(self.config.patch_size) != (1, 2, 2)
+        ):
+            raise ValueError("H3 remove-padding requires the native 24-video/32-audio patch geometry.")
+        if attn_implementation in (None, "eager", "sdpa"):
+            kernel = None
+        elif attn_implementation in ("veomni_flash_attention_2_with_sp", "veomni_flash_attention_3_with_sp"):
+            kernel = _load_veomni_local_flash_kernel(attn_implementation).flash_attn_varlen_func
```
FoolPlayer: [P2] This unconditional CUDA FlashAttention import breaks Ascend construction with the default FA2 setting, even for a single sample: Transformers previously selected its native NPU implementation, but this loader requires `flash_attn`. Please defer packed-only loading or preserve NPU backend resolution.

veomni/models/diffusers/minimax_h3/minimax_h3_core/minimax_h3_dit.py
```diff
@@ -586,9 +639,9 @@ def forward(
         # every block instead of one tolist() per attention module. Bounds stay
         # at the ORIGINAL seq_len: the per-segment SDPA is non-causal, so an
         # extended bound would leak pad keys into the last real rows. Pad rows
-        # fall outside every segment and are dropped by the index_select below
-        # (their uninitialized attention output never reaches the loss).
-        cu_bounds = tuple(cu_seqlens.tolist())
+        # fall outside every segment, have zero attention output, and are
+        # dropped by the index_select below.
+        cu_bounds = cu_seqlens if packed_batch else tuple(cu_seqlens.tolist())
```
FoolPlayer: [P2] For packed eager/SDPA, these device tensors become Python conditions and slice indices in every layer, forcing repeated device-to-host synchronization, including checkpoint recomputation. CPU profiling found 72 scalar reads for four calls with two segments, versus zero with integer bounds. Please cache host integer bounds once per forward for SDPA.

tests/trainer/test_minimax_h3_remove_padding_fsdp.py
```diff
@@ -0,0 +1,460 @@
...
+import importlib.util
+import os
+from collections.abc import Mapping
+from copy import deepcopy
+
+import pytest
+import torch
+import torch.distributed as dist
+
+from ..tools.launch_utils import torchrun
```
FoolPlayer: [P2] `make quality` fails on this head with CI's Ruff 0.13.2: three `I001` errors in the import blocks at lines 9, 75, and 131. Please place package-relative imports after the `veomni` imports and rerun the check.

veomni/models/diffusers/minimax_h3/minimax_h3_transformer/modeling_minimax_h3_transformer.py
NancyFyong: Fixed in 4e39fb22. `_configure_packed_attention` now only validates/stores the backend at construction; the FA2/FA3 varlen kernel is loaded on the first multi-sample packed forward (`_load_packed_attention_kernel`). Single-sample execution and NPU construction with default FA2 never import CUDA `flash_attn`. A packed FA forward without a loaded kernel raises rather than silently falling back to SDPA. Regression: `test_flash_backend_defers_packed_kernel_until_multisample_forward` (mimics Transformers NPU FA resolution and a failing local loader; fails on the old code).

veomni/models/diffusers/minimax_h3/minimax_h3_core/minimax_h3_dit.py
NancyFyong: Fixed in 4e39fb22. DiT and token-refiner `cu_seqlens` are read to host integers once per forward and passed as a `(device, host)` pair; packed SDPA and the sample-local refiner slice with host ints, FA varlen still gets the device int32 tensor. Measured CUDA syncs for one packed SDPA forward+backward: 69/105 (2 layers, ckpt off/on) and 141/249 (6 layers) before, constant 16 after, independent of depth and recomputation. Regression: `test_packed_sdpa_slices_with_host_bounds`. Two-rank FSDP2 FL2VA/Ref2VA eager, SDPA+ckpt, real FA2/FA3 8/8 pass.

tests/trainer/test_minimax_h3_remove_padding_fsdp.py
NancyFyong: Fixed in 4e39fb22: package-relative imports moved after the `veomni` imports; `make quality` passes with Ruff 0.13.2.

NancyFyong:
@FoolPlayer @Coach257 Hi, ci all green, could you have a look when you have time?

Coach257 (commented):
Thanks for the update — the 64-row tail is gone, packing is now a plain concat, and the trainer stays protocol-free. Two small follow-ups before this is ready:

### 1. Drop the fixed-geometry checks; packing doesn't need them

**`_configure_packed_attention` (modeling_minimax_h3_transformer.py L70-75).** Packing only concatenates encoded rows along the sequence dim and splits them back by `row_counts`. The only requirement is that all samples share the same feature dim, and that is guaranteed because they come from one condition model and one config (`torch.cat` would fail anyway). The 24 / 32 / `(1, 2, 2)` constants are an H3-wide limitation: `_PATCH_T/H/W` and the `channel=24` / `latent_dim=32` defaults in `unpatchify_video` / `unpack_audio` also apply to the single-sample path. They are not a packing constraint. Because the check runs in `__init__`, it also blocks `M=1` construction for non-native configs that build fine on `main`. Please remove it here. If that H3 limitation needs guarding, it belongs in the config or in making unpatchify read the config, which is a separate change.

**`pack_samples` (batch_packing.py L35-36).** Requiring identical `video_latent_shape` / `audio_latent_shape` across a microbatch isn't needed either. Attention is segmented by `cu_seqlens`, RoPE uses each sample's own positions, and loss is computed per sample in `_finish_outputs`. The only thing that depends on equal shapes is merging predictions with `torch.cat` / `torch.stack` in `_forward_batch`. The trainer doesn't consume `predictions`, so returning them as per-sample lists removes the restriction and lets mixed resolutions/durations share a microbatch.

### 2. Rename "remove-padding" to "packing"

There is no padding left to remove, so the name no longer describes what the code does. Please rename:
- error messages in `modeling_minimax_h3_transformer.py` L75 / L83
- `tests/models/test_minimax_h3_remove_padding.py` → `test_minimax_h3_packing.py`
- `tests/trainer/test_minimax_h3_remove_padding_fsdp.py` → `test_minimax_h3_packing_fsdp.py`, including `test_minimax_h3_remove_padding_fsdp2`, the skip messages, and the `from ..models.test_minimax_h3_remove_padding import ...` lines
- workflow entries in `gpu_unit_tests.yml` L147-148 and `npu_unit_tests.yml` L171

Coach257:
On tests, I'd suggest consolidating into a single file, `tests/models/test_minimax_h3_packing.py`:

- **Drop `tests/trainer/test_minimax_h3_remove_padding_fsdp.py`** (plus `tests/trainer/__init__.py` and its `gpu_unit_tests.yml` line). Packing is a single-process forward property, and FSDP only shards params. The trainer-side semantics it adds (1/K accumulation, DP reduce, loss logging) are model-agnostic and already covered by `tests/trainer/test_dit_microbatch.py` and the generic FSDP tests. Its BF16 `5e-2` tolerance is also too loose to catch cross-sample leakage; the FP32 `2e-5` model test catches that. It also costs 8 two-GPU `torchrun` launches.
- **No GPU-only FA case needed.** FA2 varlen kernel correctness is covered by `tests/ops/test_flash_attn_varlen_padding.py`, and FA3 can't run on the SM89 CI runners anyway. The H3 call-site contract is already tested on CPU via the monkeypatched kernel in `test_fused_dispatch_...`. Please just add layout checks there, since SDPA accepts shapes that real FA would reject:
  ```python
  assert q.ndim == 3 and q.shape[1:] == (num_heads, head_dim)
  assert q.is_contiguous() and k.shape == v.shape == q.shape
  ```
- Optionally trim the remaining cases to the core ones: packed-vs-serial equivalence, sample isolation, Ref2VA variable refs, SP padding kept forward-local, and fail-closed inputs.

NancyFyong:
@Coach257 Thanks, both suggestions made sense. Pushed 46421f79.

**Geometry checks.** The 24/32/(1,2,2) check is gone from `_configure_packed_attention`, so non-native configs build again at `M=1`. I also removed the equal-shape check in `pack_samples`. For `M>1`, `predictions` now comes back as per-sample lists (`[videos, audios]`), so mixed resolutions and durations can share a microbatch. I added a case that packs two samples with different latent T/H/W and audio length and compares outputs, losses and grads against serial. It fails on the old code and passes now. The only other consumer of `predictions` is inference, and that stays on the single-sample path, so nothing there changes.

**Rename.** The error messages now say packing, the test is `tests/models/test_minimax_h3_packing.py`, and the GPU/NPU workflow entries are updated.

**Tests.** I followed your consolidation suggestion:
- dropped the FSDP2 test, `tests/trainer/__init__.py` and its workflow line
- added the layout asserts to the stub kernel in the fused-dispatch test
- dropped the trainer-step case and the upstream-gather dtype case

I kept the smaller cases guarding earlier review fixes (lazy FA loading, host-side `cu_seqlens`, SP tail rows, offload rejection), since they're cheap and cover things that actually broke before. Happy to trim further if you'd prefer a leaner file.

I also refreshed the PR description. The FSDP2, SP and official-weight numbers there come from the heads just before this one; this commit only touches the checks, output assembly and tests.

Coach257:
Two more items:

**1. Mixed-task microbatches (FL2VA + Ref2VA) break in the collator.** `DiTDataCollator` only appends keys that a sample actually has (`dit_trainer.py` L118-120). FL2VA samples carry `keyframe_cond_anchor` and Ref2VA samples carry `ref_visual_anchor`, so each of those columns ends up with length 1 and loses track of which sample it belongs to. `process_condition` then raises `H3 input column keyframe_cond_anchor has length 1, expected 2`. Without that check, the Ref2VA anchor would silently be assigned to sample 0. I reproduced this on CPU. After filling missing keys with `None`, the same mixed batch packs and trains fine; H3 already handles `None` anchors. Suggested fix, which leaves `M=1` unchanged:
```python
keys = set().union(*(f.keys() for f in features))
for feature in features:
    for key in keys:
        batch[key].append(feature.get(key))
```
Once that's in, the column-length check in `process_condition` can go. Please also add a mixed FL2VA + Ref2VA (and keyframe / no-keyframe) packed case.

**2. Avoid the GPU->CPU round trip for segment bounds.** Per-layer use is fine: the bounds are read once per forward and shared by all blocks via `_PackedBounds`, and `max_seqlen` is a Python int. But `pack_samples` builds `cu` as a host list from shapes, uploads it to the GPU, and then the DiT reads it back with `.tolist()`. That costs M + 2 syncs per forward:
- `batch_packing.py` L43: `cu_seqlens_q.tolist()` per sample for validation. Checking on the host with `packed_seq_params["max_seqlen_q"] != length` is equivalent for single-segment layouts, including legacy 64-tail caches.
- `minimax_h3_dit.py` L591-592: `cu_seqlens.tolist()` and `refiner_cu.tolist()`. Pass the host tuples from `pack_samples` (e.g. `cu_seqlens_host`) and use them directly.

Consider adding H3 to `tests/models/test_model_forward_no_implicit_sync.py` to keep this guarded.

Coach257:
One more observation. This isn't blocking this PR and could be handled in a separate follow-up PR.

**Video-only samples still contribute audio loss.** When a sample has no audio, `get_condition` substitutes `_make_silent_audio_latent`, an all-zero `[2, 32, T_a]` latent. The layout therefore stays `[text | cond | audio | video]`, and packing mixed audio / no-audio samples already works; I verified this on CPU. However, `_finish_outputs` computes `mse_audio` for every sample, so the model is trained to emit a zero audio latent for silent or missing-audio clips. With packing, these samples also dilute the audio gradient of real-audio samples in the same microbatch. A zero latent is not necessarily what the audio VAE produces for real silence, and clips with a missing audio track shouldn't be supervised on audio at all.

Suggested follow-up:
- Keep the placeholder rows so the layout stays uniform. Don't drop audio rows: `mse_loss` on an empty tensor gives NaN, and it would also diverge from the pretraining and inference layout.
- Carry a per-sample `has_audio` flag, and persist it in offline embeddings.
- Skip or zero-weight `mse_audio` for samples with `has_audio=False`. In `_forward_batch`, average audio loss only over samples that have audio, and return 0 when none do.

NancyFyong:
@Coach257 Thanks for digging into these. I took all three, including the audio one, since it was small enough to fold in. Pushed 96332f6a.

**Mixed-task batches.** `DiTDataCollator` now fills keys a sample doesn't have with `None`, so the anchor columns stay aligned with their samples (`M=1` is unchanged). I removed the column-length check in `process_condition`. There's a new packed-vs-serial case mixing FL2VA, Ref2VA and a no-keyframe FL2VA sample, plus a small collator test in `test_dit_microbatch.py`.

**Segment bounds.** As suggested, the tail check in `pack_samples` now uses `max_seqlen_q`, and `pack_samples` passes `cu_seqlens_host` to the DiT. Going through the sync gate turned up two more read-backs:
- On the offline path, `preforward` has already moved `packed` to the GPU, so `process_condition` was syncing on `pk["cu_seqlens"]`. The bounds now come from the layout metadata instead (`host_cu_seqlens`).
- Inference had the same `int(cu[1])` read.

I also moved the timestep loss-weight lookup in `_finish_outputs` to a host scalar. It was an H2D followed by a D2H every step; the values are the same.

I added H3 to `test_model_forward_no_implicit_sync.py`: single and packed, SDPA and FA2, with conditioning run on device inputs the way the trainer does it. A single-sample step is now sync-free on the model side. A packed step only does the two `cu_seqlens` uploads the varlen kernel needs. Both are allowlisted, along with the existing RNG draw in conditioning. On the old code the gate fails, and it also catches a `tolist()` if I put one back.

**Video-only samples.** `get_condition` now records `has_audio` per sample, and it gets saved into offline embeddings. I kept the placeholder rows as you suggested. Samples with `has_audio=False` get zero-weighted `mse_audio`. In a packed batch, `mse_audio` is averaged only over samples that have audio, and it's 0 if none do. Old caches without the flag keep supervising audio, so existing runs don't change behavior.

CPU: 180 passed. On one H20, the packing, FA varlen and sync-gate tests all pass. I didn't rerun the official-weight comparison for this change; the PR description says so.

Coach257:
Thanks, I re-checked 96332f6a2: geometry checks, rename, test consolidation, collator alignment, host bounds, and mixed-task/geometry tests all look good. The CPU tests pass locally (68 passed, CUDA-only skipped) and ruff is clean.

**One fix needed: the `mse_audio` averaging in `_forward_batch`.** Dividing by `audio_count` (what I suggested last time, sorry) breaks packed/serial equivalence. The trainer divides each microbatch loss by `K` (`dit_trainer.py` L518). So with `M=1`, every audio sample gets weight `1/G`, while with `M>1` it gets `1/(K * audio_count_k)`, which depends on which samples share its microbatch. For example, with `M=2, K=2` and microbatches `[audio, no-audio]` and `[audio, audio]`, the audio sample in the first microbatch is weighted 1/2 instead of 1/4. Please use a plain `.mean()` over all samples. No-audio samples are already zero-weighted, so this gives each audio sample `1/(M*K) = 1/G`, matching `M=1`. Update `test_audio_loss_skips_samples_without_audio` to use `/3` for the reference. Exact normalization by the audio-sample count would need a global count across microbatches and DP ranks, which can be a follow-up.

Minor, non-blocking:
- The attention backend allowlist in `_configure_packed_attention` still runs in `__init__`, so it also gates `M=1` construction. Consider moving it next to the lazy kernel load on the first packed forward.
- The `len(coll) != 1` checks and the "supports micro_batch_size=1 only" message in `_process_single_condition` are now unreachable and misleading, since `process_condition` always passes single-sample lists.
- `predictions` is a tensor pair for `M=1` but a list pair for `M>1`. Nothing consumes it today, but it would be nice to document or unify this.

NancyFyong:
@Coach257 Good catch on the `/K` interaction, thanks. Fixed in 4d62f15f.

- `mse_audio` now uses the plain sample mean like the other losses. `test_audio_loss_skips_samples_without_audio` uses `/3`. I also added a test with your example (`M=2, K=2`, microbatches `[audio, no-audio]` and `[audio, audio]`): summing the two packed losses divided by 2 gives the same loss as the four samples with `M=1`. I left global audio-count normalization for a follow-up, and the docs say so.
- The backend allowlist now runs on the first packed forward, together with the lazy kernel load, so `M=1` construction no longer depends on it.
- I removed the unreachable `len(coll) != 1` checks and their message. The audio-reference rejection is still there.
- I documented `predictions` on `MiniMaxH3DiTOutput`: a `(video, audio)` pair for one sample, per-sample lists when packed. I didn't unify them because inference reads the single-sample pair.

This head also merges current main, which includes #1214's loader rename. H3 packing now uses `_load_veomni_flash_kernel` and accepts the Hub FA2/FA3 backends; the Hub test cases are skipped on NPU. CPU: 205 passed. On an H20, the packing and sync-gate tests pass.

Coach257 (approved):

PR #1214, opened 2026-09-18
[ops, model] feat: add Hugging Face Hub FlashAttention backends

NancyFyong:
### What does this PR do?

Adds explicit opt-in `flash_attention_2_hub` and `flash_attention_3_hub` backends that load version 1 of `kernels-community/flash-attn2` and `kernels-community/flash-attn3` through Hugging Face `kernels==0.16.0`.

The existing `flash_attention_2` and `flash_attention_3` backends remain unchanged and continue to use the locally installed FA2/FA3 packages. Hub kernels are imported and downloaded lazily only when explicitly selected, and dependency or download failures propagate instead of silently falling back.

This follows vLLM-Omni rollout's existing `FLASH_ATTN_HUB`/`FLASH_ATTN_3_HUB` repository choices while adding the capability to VeOmni's training-side attention dispatch. No related open VeOmni issue or PR was found.

### Checklist Before Starting

- Search for relative PRs/issues and link here: no matching open issue or PR found for `kernels-community/flash-attn2`, `kernels-community/flash-attn3`, or VeOmni Hub FlashAttention backends.

### Test

Extended the existing CI-enumerated attention contract tests under `tests/ops/`.

- `pytest -q tests/ops/test_flash_attention_contract.py tests/ops/test_fused_attention_contract.py`: 41 passed
- The same 41 tests with an isolated `pydantic==2.13.5` overlay: passed
- `make quality`: passed
- `pre-commit run --all-files`: passed
- `uv lock --check`: passed
- `git diff --check`: passed
- Both pinned Hub artifacts load under Torch 2.11 + CUDA 13.0 and expose `flash_attn_func` and `flash_attn_varlen_func`.
- NVIDIA H20 BF16 varlen forward/backward comparison between local FA3 and `kernels-community/flash-attn3` version 1:
  - output relative L2: `0.0`
  - maximum gradient relative L2: `1.99e-9`
- Full VeOmni FA3 attention-wrapper comparison on NVIDIA H20:
  - output relative L2: `0.0`
  - maximum gradient relative L2: `1.55e-5`

FA2 Hub GPU execution remains pending because all permitted GPUs were occupied by existing jobs during this update; no running job was interrupted. The PR remains Draft.

The broader `tests/ops` collection was also attempted with CUDA hidden, but an existing fused-MoE test performs an import-time CUDA device query and cannot collect without a visible GPU. The two directly affected, CI-enumerated contract files pass.

### API and Usage Example

```yaml
model:
  ops_implementation:
    # Select one explicitly; local flash_attention_2/3 remain the defaults.
    attn_implementation: flash_attention_2_hub
```

Alternatively:

```yaml
model:
  ops_implementation:
    attn_implementation: flash_attention_3_hub
```

Both options require `MODELING_BACKEND=veomni`. Selecting either with the Hugging Face modeling backend raises a configuration error.

### Design & Code Changes

- Adds `flash_attention_2_hub` and `flash_attention_3_hub` as separate public backends and maps them to their VeOmni SP-aware registry names.
- Lazily loads and caches `get_kernel("kernels-community/flash-attn2", version=1)` or `get_kernel("kernels-community/flash-attn3", version=1)` independently per process.
- Registers both backends with VeOmni fused-attention dispatch, Transformers `ALL_ATTENTION_FUNCTIONS`, model-loading allow-lists, and varlen attention classification.
- Preserves the existing Ulysses gather/scatter path and local `flash_attention_2`/`flash_attention_3` defaults.
- Fails closed when `kernels` is unavailable or a Hub artifact cannot be loaded.
- Adds `kernels==0.16.0` to the Linux GPU extra and updates the lockfile. Its expected Sigstore verification dependencies add crypto/TUF packages and resolve Pydantic 2.13.x for the Linux GPU environment; this dependency impact is documented in `.agents/knowledge/uv.md`, and relevant tests pass with Pydantic 2.13.5.
- Updates the attention adapter and fused-attention documentation.

- Applied pre-commit checks
- Added/updated documentation
- If `tasks/` training scripts were moved or renamed: not applicable
- Tests: extended existing CI-enumerated `tests/ops/` files; no new test file or workflow registration is required.

## Summary by CodeRabbit

- **New Features**
  - Added optional Hub-backed FlashAttention 2 and 3 implementations.
  - Added configuration options with sequence-parallel support.
  - Hub kernels use pinned version 1 artifacts.
  - Local FlashAttention implementations remain the default.
  - Hub options require the supported VeOmni backend and reject unsupported platforms or configurations without fallback.

- **Documentation**
  - Documented Hub-backed options, mappings, requirements, installation details, and error behavior.

- **Tests**
  - Added coverage for kernel loading, caching, missing dependencies, backend validation, and attention classification.


NancyFyong:
@Coach257 , can you review this when you have time? Thx!

veomni/arguments/arguments_types.py
```diff
@@ -1357,7 +1359,9 @@ def __post_init__(self):
         if get_env("MODELING_BACKEND") == "veomni":
             replacements = {
                 "flash_attention_2": "veomni_flash_attention_2_with_sp",
+                "flash_attention_2_hub": "veomni_flash_attention_2_hub_with_sp",
                 "flash_attention_3": "veomni_flash_attention_3_with_sp",
+                "flash_attention_3_hub": "veomni_flash_attention_3_hub_with_sp",
```
FoolPlayer: [P2] Please reject both Hub backends on NPU, including their normalized aliases. Transformers 5.16.1 selects built-in NPU attention before reaching the Hub loader, so these options silently run a different kernel and never check the `kernels` dependency. Validate this before HF preloading and add a regression test.

FoolPlayer:
[P1] Adding `kernels` breaks Llama imports even when Hub attention is not selected.

The [GPU e2e job](https://github.com/ByteDance-Seed/VeOmni/actions/runs/35694728731/job/106696606923) fails in the first Llama test, before training starts:

```text
ValueError: Function `apply_rotary_pos_emb` is not decorated with @use_kernel_forward_from_hub and cannot be used with @use_kernelized_func
```

Adding `kernels==0.16.0` to the GPU extra activates Transformers' previously inert kernel decorators. VeOmni's patchgen replaces `apply_rotary_pos_emb` without its HF decorator, but retains `@use_kernelized_func(apply_rotary_pos_emb)` on `LlamaAttention`. The latter validates the missing `kernel_layer_name` during module import. This also breaks eager/local-attention users.

Reproduced on CPU with Transformers 5.16.1 by importing `veomni.models.transformers.llama.generated.patched_modeling_llama_gpu`: succeeds without `kernels`, fails with `kernels==0.16.0`, and succeeds with `USE_HUB_KERNELS=NO` (diagnostic confirmation).

Please resolve the decorator mismatch in the patchgen source and regenerate the model, audit other models with the same pattern, and add model-import coverage with the real `kernels` package installed. The mocked attention contract tests do not cover this path.

NancyFyong:
@FoolPlayer Thanks for tracking this down, your diagnosis was spot on.

The root cause turned out to be in patchgen itself rather than a single model: `replace_function` only emitted the replacement body and dropped the original function's decorators, while `override_method` keeps them. So every model that replaces `apply_rotary_pos_emb` lost `@use_kernel_forward_from_hub("rotary_pos_emb")` and ended up with exactly this mismatch once `kernels` was installed.

Pushed 25de6e1f:
- patchgen now keeps the source function's decorators on function replacements (same behavior as `override_method`), and I regenerated everything with transformers 5.16.1.
- Audit: 12 generated files had this pattern: llama gpu, qwen2 gpu, qwen3 / qwen3_moe / qwen3_vl / qwen3_vl_moe gpu+npu. All fixed. gemma3, gpt_oss, qwen3_omni_moe, qwen3_5 and qwen4_exp were already fine. One side effect to flag: deepseek_v3_npu and seed_oss gpu/npu also get the HF decorator back now. That just matches upstream HF, but I wanted to mention it.
- Coverage: added an AST check that any function referenced by `@use_kernelized_func` has a hub decorator, plus an import test that only runs when the real `kernels` package is installed and hub kernels are enabled. Both live in `test_generated_modeling_imports.py`, which already runs in the GPU/NPU unit workflows after `kernels` is installed.

Locally with `kernels==0.16.0` on GPU: generated-import and registry tests are 73 passed. The `qwen3vl` registry case fails on the old head and passes now, and an eager Llama built from the regenerated module matches HF logits exactly. CI is re-running now, and I'll keep an eye on the llama e2e.

veomni/arguments/arguments_types.py
NancyFyong: Fixed a while back in a7ac7dce, with the alias normalization follow-up in a7ac3ce1: all four FA2/FA3 Hub names, both the public and normalized forms, are now rejected on NPU before any Transformers preload, in both config init and the `build_foundation_model` override path. I only tested this with a mocked NPU, since I do not have Ascend hardware, so feel free to double-check it on your side.

FoolPlayer (approved):

